PR #2, opened 2022-05-31
Create a release workflow and configuration

zregvart:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/cache@v3
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: ${{ runner.os }}-go-
+
+      - name: Build
+        run: make
+
+      - name: Release
```
joejstuart, line 31: Is there a title or description we can set about the release?

zregvart: Do you mean title/description to the [step name](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsname) in the GitHub workflow, or the resulting GitHub release? For the former only `name` is available; for the latter the name of the release is (currently) set to the git hash and the tag to `tip`. See an example [here](https://github.com/zregvart/ec-cli/releases/tag/tip).

joejstuart: I was thinking the release. Just something to describe it's a dev build or if there's anything we can provide that's informative. It's not a big deal.

.github/release.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
```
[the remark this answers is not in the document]
zregvart: I wanted to demo this better, so I experimented on my fork, seems that the `snapshot` tag wasn't moved so I needed to add deleting the release and re-tagging to the release workflow.

Have a look at the [snapshot release](https://github.com/zregvart/ec-cli/releases/tag/snapshot) in in my fork. Everything below `Unstable development snapshot release.` was auto-generated by the [GitHub auto-release notes feature](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes).

I've artificially added two pull requests (that's where it pulls the data from). I didn't want to go too deep into this so there is much to refine here, like adding auto-labeling to pull requests and then refining this to create more elaborate sections.

zregvart:
Looks good enough so I'll merge.

.github/release.yaml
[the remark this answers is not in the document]
zregvart: The snapshot release in my fork is not representative, I needed to force-push to it so lost the change log. The [one in this repository](https://github.com/hacbs-contract/ec-cli/releases/tag/snapshot) is a much nicer example.

PR #3, opened 2022-06-01
Build for multiple platform/architectures

zregvart:
Adds builds for MacOS (arm and amd64), and Windows in addition to Linux.
Also adds the `help` target.


Makefile
```diff
@@ -1,4 +1,42 @@
-build:
-	@go build -ldflags="-s -w -X github.com/hacbs-contract/ec-cli/cmd.Version=$$(git rev-parse HEAD)" -o dist/ec
+MAKEFLAGS+=-j
+VERSION:=$$(git log -1 --format='%H')
 
 all: build
```
lcarva, line 4: It would be nice to change this to `all: help`.

lcarva (approved):

PR #8, opened 2022-06-03
Add gitignore file

robnester-rh:
This commit adds a default .gitignore file which includes generated go
files and editor / IDE specific files and directories which are not
needed for the application itself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,173 @@
 dist
+### VisualStudioCode template
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+*.code-workspace
+
+# Local History for Visual Studio Code
+.history/
+
+### JetBrains template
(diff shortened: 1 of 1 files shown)


simonbaird (approved):

robnester-rh:
> I think ignoring IDE specific files is anti-pattern leftover when IDEs placed configuration that is specific to machine it was configured on initially. This is largely not the case today and sharing IDE specific settings like the [launch configuration](https://github.com/hacbs-contract/ec-cli/blob/main/.vscode/launch.json) helps folk onboard more easily.

The launch.json file doesn't really contain anything that would be necessary to onboard, it's specific to vscode and I wouldn't want to see jetbrains specific folders included if someone decides to use goland, etc.

zregvart:
> The launch.json file doesn't really contain anything that would be necessary to onboard, it's specific to vscode and I wouldn't want to see jetbrains specific folders included if someone decides to use goland, etc.

Makes it easier for someone to run a debugging session, otherwise they'd need to come up with one on their own. goland-specific files won't bother me I'm using vscode :grin:

I hold my opinions rather loosely so I'm okay with this getting merged, just an opinion/observation I have...

PR #11, opened 2022-06-09
Accept ApplicationSnapshot inputs

caugello:
Validates inputs from string or file agains its schema.


cmd/internal/applicationsnapshot/validation.go
```diff
@@ -0,0 +1,159 @@
...
+
+type Report struct {
+	Success bool    `json:"success"`
+	Images  []Image `json:"images"`
+}
+
+const SCHEMA = "https://raw.githubusercontent.com/redhat-appstudio/release-service/main/config/crd/bases/appstudio.redhat.com_applicationsnapshots.yaml"
+
+// Returns an error on failure.
+func Validate(jsonstring string) error {
```
robnester-rh, line 47: Functions require docstrings in the form of `// Validate ...`

cmd/eval.go
```diff
@@ -22,16 +22,6 @@ import (
...
-var arguments = struct {
-	policyConfiguration string
-	imageRef            string
-	publicKey           string
-	rekorURL            string
-	strict              bool
-}{
-	rekorURL: "https://rekor.sigstore.dev/",
-}
-
```
zregvart, line 34: I don't quite understand how is this compiling, the code below uses `arguments`

cmd/verify.go
```diff
@@ -0,0 +1,90 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		report, err := applicationsnapshot.GenerateReport(snapshot)
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+
+		fmt.Println(report)
+		os.Exit(0)
+	},
```
zregvart, line 83: In eval we do `RunE` instead and return the err, this is not nice and we have [HACBS-633](https://issues.redhat.com/browse/HACBS-633) to clean that up...

zregvart:
Just to note [go generate build](https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source) tag that can be used to fetch the CRD YAML

cmd/eval.go
caugello: I think it's because it's the same package.

cmd/verify.go
```diff
@@ -0,0 +1,90 @@
+/*
```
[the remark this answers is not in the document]
caugello: I am not sure I understand, it's under cmd.

cmd/verify.go
```diff
@@ -0,0 +1,90 @@
...
+// verifyCmd represents the verify command
+var verifyCmd = &cobra.Command{
+	Use:   "verify",
+	Short: "Change me",
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		var jsonstr string
+		// Fail if input and filepath are used.
+		if arguments.input != "" && arguments.filepath != "" {
+			fmt.Println("Use either --input or --filepath")
+			os.Exit(1)
```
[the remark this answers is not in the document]
caugello: Yes, updated it.

cmd/eval.go
zregvart: Right I missed that these got moved to `root.go`, can we keep them here instead, the intent is that these arguments are for the `eval` functionality, trying to share this struct between two sub-comands will lead to unexpected behavior (e.g. nil reference) on trying to access a field that was not bound.

cmd/internal/applicationsnapshot/applicationsnapshot_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+//go:embed test_crd.yaml
+var crdYAML string
+
+func fetchStub() (*os.File, error) {
+	file, err := ioutil.TempFile(os.TempDir(), "ec-cli-stubcrd")
```
zregvart, line 32: Not really important, this is an equivalent:

```suggestion
	file, err := ioutil.TempFile("", "ec-cli-stubcrd")
```

cmd/internal/applicationsnapshot/applicationsnapshot_test.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return nil, err
+	}
```
zregvart, line 35: File should be closed

```suggestion
	}
	defer file.Close()
```

cmd/internal/applicationsnapshot/source.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"io/ioutil"
+	"os"
+
+	"github.com/hashicorp/go-getter"
+)
+
+const SCHEMA = "https://raw.githubusercontent.com/redhat-appstudio/release-service/main/config/crd/bases/appstudio.redhat.com_applicationsnapshots.yaml"
+
+type fetcherFn func() (*os.File, error)
```
zregvart, line 29: Perhaps we should return Reader here instead?
```suggestion
type fetcherFn func() (*io.Reader, error)
```

cmd/internal/applicationsnapshot/report.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Report the states of images from the snapshot
+func Report(snapshot *rs.ApplicationSnapshotSpec) (string, error) {
+	output := report{Success: true}
+
+	for _, image := range snapshot.Images {
+		item := Image{
+			Violations: []string{},
+			Success:    true,
```
zregvart, line 43: Shouldn't we compute values for these?

cmd/internal/applicationsnapshot/report_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"strings"
+	"testing"
+
+	rs "github.com/redhat-appstudio/release-service/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
+)
+
+//go:embed test_snapshot.json
+var jsonstr string
```
zregvart, line 31: Perhaps a more indicative name?

```suggestion
var testSnapshot string
```

cmd/internal/applicationsnapshot/report_test.go
```diff
@@ -0,0 +1,69 @@
...
+          "violations": [],
+          "success": true
+        }
+      ]
+    }
+  `
+
+	expected := strings.ReplaceAll(strings.ReplaceAll(string(output), "\n", ""), " ", "")
+	report, _ := Report(snapshot)
+	assert.Equal(t, expected, report)
```
zregvart, line 68: Could use [`assert.JSONEq`](https://pkg.go.dev/github.com/stretchr/testify/assert#JSONEq) or [jsonassert](https://github.com/kinbiko/jsonassert) here

cmd/internal/applicationsnapshot/source.go
```diff
@@ -0,0 +1,67 @@
...
+	return schemaSource{
+		fetch: fetch,
+	}
+}
+
+// fetch the latest CRD and return the file
+//go:generate build
+func fetch() (*os.File, error) {
+	// Create a tempfile for the CRD
+	file, err := ioutil.TempFile(os.TempDir(), "ec-cli-applicationsnapshot")
```
zregvart, line 46: For consistency of nitpickery

```suggestion
	file, err := ioutil.TempFile("", "ec-cli-applicationsnapshot")
```

cmd/internal/applicationsnapshot/source.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Download the CRD
+	client := getter.Client{
+		Ctx:      context.Background(),
+		Src:      SCHEMA,
+		Dst:      file.Name(),
+		Insecure: false,
```
zregvart, line 56: Should be the default

```suggestion
```

cmd/internal/applicationsnapshot/source.go
```diff
@@ -0,0 +1,67 @@
...
+		Getters: map[string]getter.Getter{
+			"https": &getter.HttpGetter{},
+		},
+	}
```
zregvart, line 60: Simpler version would be:

```go
	getter.Get(file.Name(), SCHEMA, getter.WithContext(ctx))
```

`ctx` can be obtained from the Command instance (`cmd.Context()`)

cmd/internal/applicationsnapshot/source.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+
+	"github.com/hashicorp/go-getter"
+)
+
+const SCHEMA = "https://raw.githubusercontent.com/redhat-appstudio/release-service/main/config/crd/bases/appstudio.redhat.com_applicationsnapshots.yaml"
```
zregvart, line 27: I think the getter client can be forced here, so `https` works out of the box below to make the invocation a bit simpler.

```suggestion
const SCHEMA = "http::https://raw.githubusercontent.com/redhat-appstudio/release-service/main/config/crd/bases/appstudio.redhat.com_applicationsnapshots.yaml"
```

cmd/internal/applicationsnapshot/source_test.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_NewSchemaSource(t *testing.T) {
+	source := NewSchemaSource()
+	assert.IsType(t, schemaSource{}, source)
+}
```
zregvart, line 28: We want to test behavior that is not enforced by the compiler, but this helps with test coverage...

cmd/internal/applicationsnapshot/applicationsnapshot_test.go
```diff
@@ -0,0 +1,48 @@
...
+	_ "embed"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+//go:embed test_crd.yaml
+var crdYAML string
```
zregvart, line 29: Perhaps name this after the file?

```suggestion
var testCRD string
```

cmd/internal/applicationsnapshot/validation.go
```diff
@@ -0,0 +1,82 @@
...
+	// Struct to unmarshal the OpenAPIV3Schema required for validation.
+	var schemaProps struct {
+		Properties struct {
+			Spec *crd.JSONSchemaProps `json:"spec"`
+		} `json:"properties"`
+	}
+
+	//TODO: there must be a better way to avoid those 3 steps
+	// Marshal the openapi schema to JSON
+	openapiV3Schema, err := json.Marshal(s.Spec.Versions[0].Schema.OpenAPIV3Schema)
```
zregvart, line 40: We need to be careful about which version we use here, version at index 0 could be the one that is not served any more, that is I think we should filter on `CustomResourceDefinitionVersion.Served == true`

cmd/internal/applicationsnapshot/validation_test.go
```diff
@@ -0,0 +1,72 @@
...
+			name: "Missing attribute",
+			input: `{
+          "images": [
+            {
+              "component": "spam",
+              "pullSpec": "quay.io/caf/spam@sha256:123…"
+            }
+          ]
+        }
+      `,
```
zregvart, line 59: Nitpick, spaces vs tabs...

cmd/root.go
```diff
@@ -33,6 +33,18 @@ var rootCmd = &cobra.Command{
...
+	policyConfiguration string
+	imageRef            string
+	publicKey           string
+	rekorURL            string
+	strict              bool
+	input               string
+	filepath            string
+}{
+	rekorURL: "https://rekor.sigstore.dev/",
+}
```
zregvart, line 46: Let's have a version of this struct for `eval` and a version for `verify`

zregvart (commented):
It seems to me that we go to much trouble to validate the schema of the `ApplicationSnapshot`, we could have the same benefit of schema validation and more strict validation of individual fields by unmarshalling to an expected `struct` and validating the fields ourselves.

cmd/eval.go
caugello: We can, issue is that eval and verify are under the `cmd` package, any `var` will be accessible for both

caugello: If that the case, we prefix them within the struct to have `arguments.eval.imageRef` wdyt ?

cmd/internal/applicationsnapshot/report.go
caugello: We should, but in a later story. Scope is input validation and display the report from input and with above values.

cmd/internal/applicationsnapshot/source.go
caugello: Weirdly, without the config I get ` no source URL was returned`

cmd/internal/applicationsnapshot/source.go
caugello: This doesn't seem to work. getting ` no source URL was returned`

cmd/internal/applicationsnapshot/validation.go
caugello: Correct, changed it.

cmd/internal/applicationsnapshot/source.go
caugello: With `io.reader` we don't need a tmpfile anymore.

cmd/internal/applicationsnapshot/source_test.go
caugello: Yeah, that's a coverage test.

cmd/internal/applicationsnapshot/source.go
caugello: Yup, changed it.

cmd/root.go
caugello: Agreed, structs accessible trough `arguments`

cmd/eval.go
zregvart: I think we should do something that @robnester-rh suggested and implemented in the validate pipeline PR: https://github.com/hacbs-contract/ec-cli/pull/15/files#diff-6701a60d95525f3d714ff656fdc86ef5dc9753df9e79b60f0a9a7d2be4d884c6R34-R48

But we can refactor this after this PR to have this done first...

caugello: Yup, made the change

caugello: Moved it to eval, and got rid of verify

cmd/eval.go
```diff
@@ -17,80 +17,143 @@ limitations under the License.
...
+
+					jsonstr = string(file)
+				}
+
+				// Input handler
+				if arguments.input != "" {
+					jsonstr = arguments.input
+				}
+				// Unmarshall json into struct, exit on failure
+				if err := json.Unmarshal([]byte(jsonstr), &snapshotSpec); err != nil {
```
zregvart, line 74: Hmm, is this `**appstudioshared.ApplicationSnapshotSpec`? Looks odd given _If v is nil or not a pointer, Unmarshal returns an InvalidUnmarshalError._ from the docs of [json.Unmarshal](https://pkg.go.dev/encoding/json#Unmarshal).
I think we're missing an unit test for this to make sure.

cmd/eval.go
```diff
@@ -17,80 +17,143 @@ limitations under the License.
...
+				if snapshotSpec == nil {
+					return errors.New("ApplicationSnapshot input could not be unmarshalled.")
+				}
+			}
+
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			//TODO: refactor once we handle application snapshots
+			if arguments.input != "" || arguments.filepath != "" {
```
zregvart, line 87: Perhaps we could check `snapshotSpec != nil` here?

PR #12, opened 2022-06-10
Acceptance tests

zregvart:
This adds acceptance tests using Godog -- a Cucumber implementation in go, Testcontainers and WireMock. Initial happy day scenario was added for the `ec eval` case.

Implementation of the tests make them completely self-contained -- meaning: any service needed for running the system under test (the ec command) is stubbed either by an appropriate Linux container launched using Testcontainers or via WireMock (which is also launched via Testcontainers). This not only makes the tests deterministic but also allows them to be quick to execute (currently about 5 sec).

More details in ./internal/acceptance/README.md.

Ref. https://issues.redhat.com/browse/HACBS-636

Also included in a separate commit to ease review:

Acceptance test data [c41218d](https://github.com/hacbs-contract/ec-cli/pull/12/commits/c41218db699a93ed959b686fef3253b856028e82)

Contains WireMock stubs recorded when the ec command line interacted
with Kubernetes apiserver.


simonbaird:
It's quite a lot of moving parts, but I have no objections.

Is there a chance that maintaining all those wiremock stubs will become a pain?

simonbaird (approved):

PR #13, opened 2022-06-10
Adding machine readible output

bcaton85:
Outputting in JSON structure success of policy check, image signature verification, and attestation signature verification.


zregvart (approved):
LGTM, we might want to expand on this by having a parameter that determines the kind of output we want (like `--output JSON` or `--output plain`). Let's start here and figure out that in a different PR/issue...

PR #14, opened 2022-06-13
Fix fetching git repository via go-getter

zregvart:
If we point to an existing directory go-getter will try to update and
expect a git repository at that path. Given that we create a temporary
directory without any git information within it fetching will fail with
an error:

```
/usr/bin/git exited with 128: fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

Since go-getter expects an non-existing directory we extend the path of
the temporary directory to provide one.

### DIFF
--- a/cmd/internal/policy/policy_test.go
+++ b/cmd/internal/policy/policy_test.go
@@ -103,6 +103,10 @@ func Test_NewPolicyEvaluator(t *testing.T) {
 }
 
 func policyFetchStub(dst string, src string, opts ...getter.ClientOption) error {
+	err := os.MkdirAll(dst, 0755)
+	if err != nil {
+		return err
+	}
 	f, err := os.Create(path.Join(dst, "main.rego"))
 	if err != nil {
 		return err
--- a/cmd/internal/policy/source.go
+++ b/cmd/internal/policy/source.go
@@ -19,6 +19,7 @@ package policy
 import (
 	"context"
 	"os"
+	"path"
 
 	ecp "github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
 	"github.com/hashicorp/go-getter"
@@ -65,6 +66,8 @@ func (s *policySource) fetchPolicySourceFromGit(ctx context.Context, repository 
 	if err != nil {
 		return nil, err
 	}
+
+	dest = path.Join(dest, "source")
 
 	if err := s.fetch(dest, source, getter.WithContext(ctx)); err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


zregvart:
@robnester-rh is this the issue you encountered?

robnester-rh:
@zregvart while this is the issue I ran into, I feel like this is a rather hacky way to deal with this. Ultimately, I'd argue that using a package that doesn't exhibit this behavior would be preferable. I'm currently working with the `go-git` library for git clones, etc.

simonbaird:
My 2 cents on the "use another library": It seems like bike shedding to me, unless we think go-getter is seriously broken for some reason.

robnester-rh:
> My 2 cents on the "use another library" suggestion: It seems like bike shedding to me, unless we think go-getter is seriously broken for some reason.

The [proposed workaround](https://github.com/hacbs-contract/ec-cli/pull/14) to the go-getter issue is to:
* create a directory
* delete the directory
* then pass that directory name on to go-getter to use for the clone destination

The reason for this is that if the directory were to exist, go-getter fails to clone into because the directory isn't a git repository already. The [issue for this](https://github.com/hashicorp/go-getter/issues/114) was opened in 2018 but the referenced PR doesn't actually resolve the issue. I've tried this with v2 of go-getter which still remains broken in this regard.

I completely understand the appeal of a library that handles "fetching" from multiple "targets" (although, currently, we're using no other "target" than git), and I didn't choose a different library to be obstinate. It just seems to me that go-getter is broken when it comes to git cloning imho. With that said, if others agree that they're ok with this arguably broken implementation in go-getter, and the proposed workaround, then I'll update my code accordingly.

robnester-rh:
> An additional point was that this was the exact behavior of the bash scripts -- they failed in cases when the destination directory for fetching existed.

While that may have been the case, other libraries do no experience this issue.

> My 2 cents on the "use another library" suggestion: It seems like bike shedding to me, unless we think go-getter is seriously broken for some reason.

I feel like the fact that we're having to code around an issue with go-getter that was identified in 2018 but still persists in the current v2 library, says it's broken, at least in regard to fetching with git. Other libraries make git fetching a no-brainer, whether the directory exists or not.

I completely understand the appeal of go-getter, I agree that it's nice to have the ability to fetch from multiple "targets" with one library. I just feel like we're coding around an issue (inability to clone into an existing directory) that is kind of a core functionality. With all of that said, if the consensus is that folks want to do that, so be it and LGTM.

PR #15, opened 2022-06-14
Add validate pipeline functionality

robnester-rh:
This commit adds the validate subcommand which has a subsequent
subcommand for pipeline. See the help command output for information on
how to validate multiple pipeline definition files against the given
policy.


cmd/internal/pipeline/validate.go
```diff
@@ -0,0 +1,104 @@
...
+	// Checkout policy repo into work directory.
+	_, err = git.PlainClone(workDir, false, &git.CloneOptions{
+		URL:           policyRepo,
+		Progress:      nil,
+		ReferenceName: plumbing.NewBranchReferenceName(ref),
+		SingleBranch:  true,
+	})
+	if err != nil {
+		fmt.Printf("Error checking out polices: %s\n", err)
+		os.Exit(1)
```
zregvart, line 50: Don't write to stdout and terminate the process, return an error instead

cmd/internal/pipeline/validate.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	policyPath := fmt.Sprintf("%s/%s", workDir, policyDir)
+
+	// Ensure that the specified policy path exists in our cloned repo
+	exists, err = afero.Exists(AFS, policyPath)
+	if !exists {
+		fmt.Println("Specified policy directory does not exist in policy repo:", policyDir)
+		os.Exit(1)
```
zregvart, line 59: Here as well

cmd/internal/pipeline/validate.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	if err != nil {
+		fmt.Println("Error determining if specified policy directory exists.")
+		os.Exit(1)
```
zregvart, line 63: Here as well

cmd/internal/pipeline/validate.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Create our policy path arrays
+	var policyPaths = []string{policyPath}
+
+	// Create data dir
+	dataDir := fmt.Sprintf("%s/data/", workDir)
+	_ = AFS.MkdirAll(dataDir, 0755)
+	err = afero.WriteFile(AFS, fmt.Sprintf("%s/data/data.json", workDir), []byte("{\"config\":{}}\n"), 0777)
+	if err != nil {
+		os.Exit(1)
```
zregvart, line 74: Don't terminate the process return an error instead

cmd/internal/pipeline/validate.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Create our conftest runner
+	var runner = runner2.TestRunner{
+		Policy:    policyPaths,
+		Data:      dataPaths,
+		Namespace: []string{namespace},
+		NoFail:    true,
+		Output:    outputFormat,
+	}
+	results, err := runner.Run(context.Background(), []string{fpath})
```
zregvart, line 88: Use the context from the `Command`, pass it in as an argument to the function

cmd/internal/pipeline/validate.go
```diff
@@ -0,0 +1,104 @@
...
+	var dataPaths = []string{dataDir}
+
+	// Create our conftest runner
+	var runner = runner2.TestRunner{
+		Policy:    policyPaths,
+		Data:      dataPaths,
+		Namespace: []string{namespace},
+		NoFail:    true,
+		Output:    outputFormat,
+	}
```
zregvart, line 87: `eval` command uses [`engine.Check`](https://pkg.go.dev/github.com/open-policy-agent/conftest@v0.32.1/policy#Engine.Check) we should find a way to share this common logic between the two commands

cmd/internal/pipeline/validate.go
```diff
@@ -0,0 +1,104 @@
...
+		SuppressExceptions: runner.SuppressExceptions,
+	})
+	fmt.Println("# Input file:", fpath)
+	fmt.Println("# Pipeline name:", pipelineName)
+	fmt.Println("# Policy repo:", policyRepo)
+	fmt.Println("# Git ref:", ref)
+	err = outputter.Output(results)
+	if err != nil {
+		os.Exit(1)
+	}
```
zregvart, line 103: We should return the output here and have the `validatePipelineCmd` present it, as we do it in the `eval` command [here](https://github.com/hacbs-contract/ec-cli/blob/6fc1fb44e4ebd468d838471454d1845c9a8066da/cmd/eval.go#L90)

zregvart (commented):
We should strive to be consistent here. There should be common logic that is reused (fetching and evaluation of policies), error handling should be done in the same manner to help facilitate changes we'll introduce with [HACBS-633](https://issues.redhat.com/browse/HACBS-633).

cmd/internal/pipeline/validate.go
```diff
@@ -0,0 +1,104 @@
+package pipeline
+
+import (
+	"context"
+	"fmt"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/open-policy-agent/conftest/output"
+	runner2 "github.com/open-policy-agent/conftest/runner"
```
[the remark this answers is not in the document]
robnester-rh: renamed because runner was used for a variable name, this has been updated and the import no longer has an alias.

cmd/internal/pipeline/utils_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func setupTest(pipelineFilePath string, content string) {
+	AFS.Fs = afero.NewMemMapFs()
+	_ = AFS.WriteFile(pipelineFilePath, []byte(content), 0755)
```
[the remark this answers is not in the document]
zregvart: My understanding is that error can be produced in this case as well (say out of memory), tracking it down will be difficult if we don't know what the error is.

cmd/internal/pipeline/validate.go
robnester-rh: Executing the `conftest test` cli command [invokes](https://github.com/open-policy-agent/conftest/blob/88e6b0ee78c461bb66cdbe93e42488b5c91c1201/internal/commands/test.go#L119) a `runner.TestRunner` as above, it seemed to me like this was the most direct approximation of what we had been doing previously?

cmd/internal/pipeline/evaluation_target_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+	"github.com/hacbs-contract/ec-cli/cmd/internal/utils"
+	"github.com/spf13/afero"
+	"os"
+	"testing"
+)
+
+var pipelineDefinitionDir = "./test_data/pipeline_definitions"
+
+func deployPipelineFile() error {
```
zregvart, line 13: I _think_ we could replace this with `//go:embed` something like

```go
//go:embed test_data/pipeline_definitions/
var pipelineDefinitions embed.FS
```

And then in tests we could do something like:

```go
defaultAppFS := util.AppFs
util.AppFs = afero.FromIOFS{fs: pipelineDefinitions}
defer util.AppFs = defaultAppFS
```

cmd/internal/pipeline/evaluator.go
```diff
@@ -0,0 +1,59 @@
...
+	PolicySources []PolicySource
+	Paths         ConfigurationPaths
+	TestRunner    runner.TestRunner
+	Namespace     []string
+	OutputFormat  string
+	workDir       string
+}
+
+func (e *Evaluator) addDataPath() error {
+	dataDir := fmt.Sprintf("%s/data/", e.workDir)
```
zregvart, line 25: Let's use `filepath.Join`
```suggestion
	dataDir := filepath.Join(e.workDir, "data")
```

cmd/internal/pipeline/evaluator.go
```diff
@@ -0,0 +1,59 @@
...
+	exists, err := afero.DirExists(utils.AppFS, dataDir)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		_ = utils.AppFS.MkdirAll(dataDir, 0755)
+		err = afero.WriteFile(utils.AppFS, fmt.Sprintf("%s/data/data.json", e.workDir), []byte("{\"config\":{}}\n"), 0777)
```
zregvart, line 32: Should the config be in `data/config/policy/data.json`? We could use `filepath.Join` here as well

cmd/internal/pipeline/evaluator.go
```diff
@@ -0,0 +1,59 @@
...
+	return nil
+}
+func (e *Evaluator) addPolicyPaths() error {
+	for i := range e.PolicySources {
+		err := e.PolicySources[i].getPolicies(e.workDir)
+		if err != nil {
+			return err
+		}
+		policyDir := e.PolicySources[i].getPolicyDir()
+		policyPath := fmt.Sprintf("%s/%s", e.workDir, policyDir)
```
zregvart, line 47: Let's use `filepath.Join` here

go.mod
```diff
@@ -3,12 +3,15 @@ module github.com/hacbs-contract/ec-cli
 go 1.17
 
 require (
+	github.com/ghodss/yaml v1.0.0
+	github.com/go-git/go-git/v5 v5.4.2
 	github.com/google/go-containerregistry v0.8.1-0.20220216220642-00c59d91847c
 	github.com/hacbs-contract/enterprise-contract-controller v0.0.0-20220525095719-1a34e1604a54
-	github.com/hashicorp/go-getter v1.6.1
```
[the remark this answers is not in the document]
robnester-rh: ack

cmd/internal/pipeline/evaluator.go
robnester-rh: That's a good question, do we have any documentation on where we expect it to exist? In the [original bash script](https://github.com/redhat-appstudio/build-definitions/blob/0b05a990917f20f82394d02f28841500df23e612/hack/verify-pipeline.sh#L26), `DATA_DIR` was defined as `$EC_WORK_DIR/data`, which I've replicated here.

zregvart: I think what we have here is correct at least according to the [current Makefile](https://github.com/hacbs-contract/ec-policies/blob/f17f2e54d59560b5b84ce5b9a1f08b59573f53de/Makefile#L186-L187), we _might_ have had it in `data/config/policy/data.json` before or I imagined it to be so.

zregvart (approved):
:rocket:

PR #18, opened 2022-06-17
Acceptance test coverage report

zregvart:
This modifies the source files to add code coverage instructions much the same as `go tool cover` would do using github.com/mendersoftware/gobinarycoverage. The `acceptance` target now
copies the source code to a temporary directory makes the source code instrumentation and applies the patch to make sure the `coverReport` is invoked and that the missing `os` import is added. Next the build is performed so that the instrumented version of the `ec` command is built for use by the acceptance tests. Which in turn configure environment variables to generate the `coverage-acceptance-*.out` file in the root of the repository.

In addition to this from the `cmd/internal` package the `internal` is removed as the variable holding the coverage needs to be imported from `main.go` via `.../cmd/internal` which is not allowed.

Obviously this includes the un-merged #12 as well.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -51,9 +51,15 @@ jobs:
 
       - name: Upload coverage report
         uses: codecov/codecov-action@v3
-
-      - name: Build
-        run: make dist/ec_linux_amd64
+        with:
+          files: ./coverage-unit.out
+          flags: unit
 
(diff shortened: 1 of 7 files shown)


caugello (approved):

zregvart:
I've split out #23 from this as it is requirement of this PR (for `main.go` to be able to access internal packages).

PR #20, opened 2022-06-17
Configure linting using golangci-lint

zregvart:
Configures linting to be run on GitHub actions and via `lint` make
target using golangci-lint[1]

[1] https://golangci-lint.run/

Ref. https://issues.redhat.com/browse/HACBS-635


zregvart:
@hacbs-contract/devs can someone have a look and see if I missed anything?

PR #26, opened 2022-06-22
Adopt golang 1.18

zregvart:
I wanted to use `testing.F`[1] which was added in 1.18 to add a fuzzing test without adding a new dependency like gofuzz[2]. This should make the transition in all relevant places.

[1] https://pkg.go.dev/testing#F
[2] github.com/google/gofuzz


robnester-rh (approved):
LGTM

PR #30, opened 2022-06-22
Add NewPipelineEvaluator function

robnester-rh:
This commit adds a NewPipelineEvaluator function which returns an
Evaluator that is ready to execute its TestRunner.Run method to evaluate
its associated Pipeline Definition file against its associated policy
namespace.

### DIFF
--- a/internal/pipeline/evaluator_test.go
+++ b/internal/pipeline/evaluator_test.go
@@ -18,6 +18,7 @@ package pipeline
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


internal/pipeline/evaluator_test.go
```diff
@@ -27,10 +28,23 @@ import (
...
 	return &git.Repository{}, nil
 }
 
+func checkoutRepoFailureMock(_ string, _ bool, _ *git.CloneOptions) (*git.Repository, error) {
+	return nil, fmt.Errorf("unable to checkout repo")
+}
+
+func createWorkDirMock(_ afero.Fs, _ string, _ string) (string, error) {
+	_ = utils.AppFS.MkdirAll("/tmp/ec-work-1234", 0755)
+	return "/tmp/ec-work-1234", nil
```
zregvart, line 41: Would this be an issue if we hit this concurrently? Perhaps we could use [TempDir](https://pkg.go.dev/github.com/spf13/afero#TempDir) instead?

internal/pipeline/evaluator_test.go
```diff
@@ -193,3 +207,142 @@ func TestEvaluator_createWorkDir(t *testing.T) {
...
+				TestRunner: runner.TestRunner{
+					Policy:    []string{"/tmp/ec-work-1234/policy"},
+					Data:      []string{"/tmp/ec-work-1234/data"},
+					Namespace: []string{"pipeline.main"},
+					NoFail:    true,
+					Output:    "json",
+				},
+				Namespace:    []string{"pipeline.main"},
+				OutputFormat: "json",
+				workDir:      "/tmp/ec-work-1234",
```
zregvart, line 262: Aah, I guess it would complicate the test here...

zregvart (approved):
:heart: the refactor

internal/pipeline/evaluator_test.go
robnester-rh: This shouldn't be an issue as when this is called, we've set `utils.AppFS` to a new `afero.NewMemMapFs()` each time, see [this](https://github.com/hacbs-contract/ec-cli/pull/30/files/a3ed3a2775183376d0af696df63f65e941b034d6#diff-d16fab1c28968f90e4d9a19026cc0068615f5a92d49dc093d6d53096b78d67dcR316) as an example.

internal/pipeline/evaluator_test.go
robnester-rh: Yes.

PR #31, opened 2022-06-22
Update docstring for EvaluationTarget interface

robnester-rh:
Updated docstring for EvaluationTarget interface to include additional
required function to implement.


robnester-rh:
This is a docstring comment only, merging.

PR #32, opened 2022-06-23
Align with refactoring in ec-policies

zregvart:
We changed the namespace in ec-policies so we need to update it here.
The `hack/demo.sh` is now pointing to images with HACBS PipelineRun
attestations, which includes the test stream data.


robnester-rh (approved):
LGTM

PR #33, opened 2022-06-23
Update README.md to include validate command.

robnester-rh:
This commit adds the `validate` command to the `ec` output in the
README.md file.


robnester-rh:
Just a quick `README.md` update to include `validate` command. No need for checks.

PR #34, opened 2022-06-23
Fix usage of outfile-file with AppSnapshot

lcarva:
Prior to this change, if `--output-file` was used in combination with
either `--input` or `--filepath` , ec-cli didn't terminate after writing
the report to a file. It only did that on error. This caused an issue
later on about ec-cli not being able to parse an empty image reference,
because... imageRef was indeed an empty string.

The fix is to always return early when either `--input` or `--filepath`
are used.


lcarva:
@hacbs-contract/devs ptal

zregvart (approved):

PR #36, opened 2022-06-24
Fix deadlock and concurrency in eval command

lcarva:
```
Fix concurrency of eval command

By reading from the channels immediately after invoking a goroutine, we
are effectively waiting for that goroutine to complete before
proceeding, taking away the desired concurrent behavior.

The fix is to only read from the channels once all the goroutines have
completed.
```

```
Fix deadlock in eval command

Prior to this commit, if there were any errors found during the
goroutine to validate an image, nothing was written to the component
channel. Since this channel is always read immediatelly after calling
the goroutine, it blocked forever waiting for something to be added to
the component channel.

This commit fixes the issue by ensuring something is always written to
the component channel.

https://issues.redhat.com/browse/HACBS-759
```

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -121,11 +121,8 @@ func evalCmd() *cobra.Command {
 			if snapshotSpec != nil {
 
 				var wg sync.WaitGroup
-				component := make(chan applicationsnapshot.Component, len(snapshotSpec.Components))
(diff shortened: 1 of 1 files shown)


lcarva:
@hacbs-contract/devs ptal

robnester-rh (approved):
LGTM

PR #37, opened 2022-06-24
Make namespace optional for the --policy param

lcarva:
https://issues.redhat.com/browse/HACBS-754

### DIFF
--- a/internal/policy/kubernetes_client.go
+++ b/internal/policy/kubernetes_client.go
@@ -18,10 +18,12 @@ package policy
 
 import (
 	"context"
+	"fmt"
 
 	ecp "github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/clientcmd"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


internal/policy/kubernetes_client_test.go
```diff
@@ -41,33 +43,83 @@ func init() {
...
+			ecp:            nil,
+			kubeconfig:     "",
+			err:            "Unable to determine current namespace: missing current context",
+		},
+		{
+			name:           "fetch-policy-not-found",
+			namespacedName: types.NamespacedName{Name: "ec-policy", Namespace: "missing"},
+			ecp:            nil,
+			kubeconfig:     "",
+			err:            `enterprisecontractpolicies.appstudio.redhat.com "ec-policy" not found`,
```
lcarva, line 90: The expected message changed because the previous message was not correct, which highlighted a previous bug in the tests that is now fixed.

lcarva:
@hacbs-contract/devs ptal

internal/policy/kubernetes_client.go
```diff
@@ -64,10 +66,47 @@ func createControllerRuntimeClient() (client.Client, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
 	return policy, nil
 }
+
+// getCurrentNamespace returns the namespace of the current context if one is set.
+func getCurrentNamespace() (namespace string, err error) {
+	baseErr := "Unable to determine current namespace"
```
zregvart, line 86: Nit: go errors start with [lowercase letters](https://github.com/golang/go/wiki/CodeReviewComments#error-strings), it is weird but a go thing

```suggestion
	baseErr := "unable to determine current namespace"
```

zregvart (approved):
Looks good

internal/policy/kubernetes_client.go
lcarva: Thanks for the tip!

PR #38, opened 2022-06-27
Refactor eval command

zregvart:
This refactors the eval command to process both a single image and an
ApplicationSnapshot parameters in the same manner. When image is
specified, an artificial ApplicationSnapshot is created containing one
Unnamed component with the provided image.
The bulk of the processing is pulled out into two functions to ease
maintenance and testing.

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


cmd/eval.go
```diff
@@ -26,170 +27,121 @@ import (
...
+					e := fmt.Errorf("error validating image %s of component %s: %w", r.component.ContainerImage, r.component.Name, r.err)
+					err = multierror.Append(err, e)
+				} else {
+					components = append(components, r.component)
 				}
-				out.SetPolicyCheck(results)
-
-				return out, nil
+			}
+			if err != nil {
```
lcarva, line 121: FYI, we should change this. We still want to produce a report if there was a validation error that is not due to a policy violation, e.g. missing signature, missing entry in rekor. We have a tracker for this and tackle it later. Just mentioning it here because that will likely do away with this multierror variable.

cmd/eval.go
```diff
@@ -223,3 +175,80 @@ func evalCmd() *cobra.Command {
...
+			return nil, err
+		}
+
+		return &appSnapshot, nil
+	}
+
+	// create ApplicationSnapshot with a single image
+	if arguments.imageRef != "" {
+		return &appstudioshared.ApplicationSnapshotSpec{
+			Components: []appstudioshared.ApplicationSnapshotComponent{
```
lcarva, line 210: :heart:

lcarva (approved):

cmd/eval.go
zregvart: Yeah, we should file a separate story for that, not sure it'll be covered by [HACBS-633](https://issues.redhat.com/browse/HACBS-633). For now I think it is beneficial to report all errors.

lcarva: I believe it will be covered by [HACBS-760](https://issues.redhat.com/browse/HACBS-760).

PR #40, opened 2022-06-28
Refactor Evaluate function in policy.go

robnester-rh:
This commit refators the Evaluate function in policy.go to use the
runner.TestRunner struct conftest/runner. As part of this refactoring it
was found that the attestation acceptance test needed to be updated as
well.


zregvart (approved):
:sunglasses:

PR #41, opened 2022-06-28
Restore stubbed environment in acceptance tests

zregvart:
When tests are run with `-persist` parameter a `.persisted` file is created with data needed to re-run the acceptance tests with the exactly the same data as it was run. To re-run the acceptance tests in such a way `-restore` parameter needs to be specified.

With this a marker interface is introduced `testenv.WithState` needed only to access the key under which the state is stored in `context.Context` and in the persisted file.

Given that the same structure is used, for ease of development, for storing data in `context.Context` and in the persisted file, storing complex values (such as `v1.Image` from `google/go-containerregistry`) is not possible, the data needs to be fetched. For example in the case of signature/attestation data it is fetched from the stubbed registry and not stored in memory within `context.Context`.

Using a single key to store the whole persisted data in `context.Context` reduces the computational complexity from O(N) to O(1); though there are still some values that are stored in `context.Context` and not managed by `testenv`, such as the ones in the `cli` package.

The `registry.go` was moved to the `registry` package in order for the `key` type to differ between it and `image` package.

Also includes:

[Visual Studio Code configuration](https://github.com/hacbs-contract/ec-cli/commit/10f06e63017e6f186f504ed44675c0276ea17f62)

Adds launch configurations to ease running acceptance tests in various combinations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -171,3 +171,5 @@ tags
 # Persistent undo
 [._]*.un~
 
+# Last persisted acceptance test run information
+internal/acceptance/.persisted
--- a/internal/acceptance/git/git.go
+++ b/internal/acceptance/git/git.go
@@ -42,13 +42,33 @@ import (
 
 type key int
 
-const (
-	gitHostKey         key = iota // we store the host:port under this in the Context
-	gitRepositoriesKey            // key to the path to the TEMP directory mounted in the git server container in the Context
-)
+const gitStateKey = key(0) // we store the gitState struct under this key in Context and when persisted
+
+type gitState struct {
+	HostAndPort     string
+	RepositoriesDir string
+}
+
+func (g gitState) Key() any {
+	return gitStateKey
+}
+
+func (g gitState) Up() bool {
+	return g.HostAndPort != "" && g.RepositoriesDir != ""
+}
 
 // startStubGitServer launches a stub git server and exposes the port via NAT from the container
 func startStubGitServer(ctx context.Context) (context.Context, error) {
+	var state *gitState
+	ctx, err := testenv.SetupState(ctx, &state)
+	if err != nil {
+		return ctx, err
+	}
+
+	if state.Up() {
+		return ctx, nil
+	}
+
 	// the directory on the host where we'll store the git repositories
 	repositories, err := os.MkdirTemp("", "git.*")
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -171,3 +171,5 @@ tags
 # Persistent undo
 [._]*.un~
 
+# Last persisted acceptance test run information
+internal/acceptance/.persisted
```
lcarva, line 175: Missing new line at end of file

internal/acceptance/git/git.go
```diff
@@ -42,13 +42,33 @@ import (
 
 type key int
 
-const (
-	gitHostKey         key = iota // we store the host:port under this in the Context
-	gitRepositoriesKey            // key to the path to the TEMP directory mounted in the git server container in the Context
-)
+const gitStateKey = key(0) // we store the gitState struct under this key in Context and when persisted
```
lcarva, line 45: For my own understanding, what's the point of using `key(0)` instead of `int(0)`, or simply `0`?

lcarva (approved):
No concerns on my end, but I have to admit that I don't fully understand the `acceptance` portion of the code base.

internal/acceptance/git/git.go
zregvart: When placed in `context.Context` we can end up with multiple keys having control over the same value, for example:

```go
const (
	keyA = 0
	keyB = 0
)

ctx := context.WithValue(context.TODO(), keyA, "A")
ctx = context.WithValue(context.TODO(), keyB, "B")

ctx.Value(keyA) == "B"
ctx.Value(keyB) == "B"
```

Giving each key a type, even though they have the same value makes them differ in the context; so these two do not control the same entry in Context:

```go

type keyA int
type keyB int

const (
	key1 = keyA(0)
	key2 = keyB(0)
)

ctx := context.WithValue(context.TODO(), keyA, "A")
ctx = context.WithValue(context.TODO(), keyA, "B")

ctx.Value(keyA) == "A"
ctx.Value(keyB) == "B"
```

(all examples written from the top of my head, so don't mind the syntax issues)

lcarva: This is great, thanks for taking the time to explain.

PR #44, opened 2022-06-29
Refactor policy output

zregvart:
The policy output (`internal/policy.Output`) now prints to `io.Writer` instead of printing to `stdout` directly. This allows us to control where the output is printed to. Printing is now possible for a slice of `Output`s as well, which also generates the output in JSON format.

In pipeline validations a refactor was done so that `ValidatePipeline` now returns `Output` so that printing of the output is done in the validate command rather than in `ValidatePipeline`.

To ease testing, by mocking the `ValidatePipeline`, the `validatePipelineCmd` now takes a function parameter for the implementation of `ValidatePipeline`-like function.

`validatePipelineCmd` was also moved to `validate_pipeline.go`.

In `validatePipelineCmd` all errors returned from `ValidatePipeline` are accumulated and combined into a single `error`.

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -30,69 +30,8 @@ func validateCmd() *cobra.Command {
 	return cmd
 }
 
-func validatePipelineCmd() *cobra.Command {
-	var data = struct {
-		FilePaths         []string
-		PolicyRepo        string
-		PolicyDir         string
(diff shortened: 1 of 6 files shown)


robnester-rh (approved):
LGTM

PR #51, opened 2022-07-04
Refine `eval` output

zregvart:
This sets the `violations` to contain string messages instead of output of Conftest and makes sure that image and attestation signature errors are present in it. The output to standard error was left unchanged.

Ref. https://issues.redhat.com/browse/HACBS-760

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -98,8 +98,10 @@ func evalCmd(validate imageValidationFn) *cobra.Command {
 
 					// Skip on err to not panic. Error is return on routine completion.
 					if err == nil {
-						res.component.Violations = out.PolicyCheck
+						res.component.Violations = out.Violations()
 					}
+
+					res.component.Success = err == nil && len(res.component.Violations) == 0
 
 					ch <- res
 				}(c)
@@ -109,33 +111,30 @@ func evalCmd(validate imageValidationFn) *cobra.Command {
 			close(ch)
 
 			components := []applicationsnapshot.Component{}
-			var err error = nil
+			var allErrors error = nil
 			for r := range ch {
 				if r.err != nil {
 					e := fmt.Errorf("error validating image %s of component %s: %w", r.component.ContainerImage, r.component.Name, r.err)
-					err = multierror.Append(err, e)
+					allErrors = multierror.Append(allErrors, e)
 				} else {
 					components = append(components, r.component)
 				}
 			}
-			if err != nil {
-				return err
-			}
 
 			report, err, success := applicationsnapshot.Report(components)
-			if err != nil {
-				return err
+			if allErrors != nil {
+				return multierror.Append(allErrors, err)
 			}
 
 			if arguments.output != "" {
 				if err := ioutil.WriteFile(arguments.output, []byte(report), 0644); err != nil {
-					return err
+					return multierror.Append(allErrors, err)
 				}
 				fmt.Printf("Report written to %s\n", arguments.output)
 			} else {
-				_, err = cmd.OutOrStdout().Write([]byte(report))
+				_, err := cmd.OutOrStdout().Write([]byte(report))
 				if err != nil {
-					return err
+					return multierror.Append(allErrors, err)
 				}
 			}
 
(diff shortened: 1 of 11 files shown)


cmd/eval.go
```diff
@@ -220,31 +236,38 @@ func determineInputSpec(arguments args) (*appstudioshared.ApplicationSnapshotSpe
...
-		return nil, err
+	} else {
+		out.SetAttestationSignatureCheck(true, "success")
 	}
-	out.SetAttestationSignatureCheck(true, "success")
 
-	p, err := policy.NewPolicyEvaluator(policyConfiguration)
-	if err != nil {
-		return nil, err
+	if !out.ImageSignatureCheck.Passed || !out.AttestationSignatureCheck.Passed {
```
lcarva, line 266: Technically, we don't need the `ImageSignatureCheck` to have passed to perform the policy evaluation. `AttestationSignatureCheck` is the only requirement.

cmd/eval.go
```diff
@@ -220,31 +236,38 @@ func determineInputSpec(arguments args) (*appstudioshared.ApplicationSnapshotSpe
...
+		return out, nil
 	}
 
-	results, err := p.Evaluate(ctx, i.Attestations())
+	results, err := policyEvaluator.Evaluate(ctx, imageValidator.Attestations())
 	if err != nil {
 		return nil, err
```
lcarva, line 272: Maybe this should call something like `out.SetPolicyCheck(false, err)` like the other checks.

cmd/eval.go
```diff
@@ -98,9 +98,28 @@ func evalCmd(validate imageValidationFn) *cobra.Command {
 
 					// Skip on err to not panic. Error is return on routine completion.
 					if err == nil {
-						res.component.Violations = out.PolicyCheck
+						violations := make([]string, 0, 10)
```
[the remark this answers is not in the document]
zregvart: I think it makes more sense to move it to `policy.Output`

cmd/eval.go
```diff
@@ -220,31 +236,38 @@ func determineInputSpec(arguments args) (*appstudioshared.ApplicationSnapshotSpe
...
+	}
 
-	i, err := image.NewImageValidator(ctx, imageRef, publicKey, rekorURL)
+	policyEvaluator, err := policy.NewPolicyEvaluator(policyConfiguration)
 	if err != nil {
 		return nil, err
 	}
 
-	if err := i.ValidateImageSignature(ctx); err != nil {
+	return validateImageWith(ctx, imageRef, policyConfiguration, publicKey, rekorURL, imageValidator, policyEvaluator)
```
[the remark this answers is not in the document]
zregvart: Can't mock `validateImage` in the test then

cmd/eval.go
zregvart: I think we could check if there are any attestations present

cmd/eval.go
zregvart: `policy.Output` holds [results](https://github.com/hacbs-contract/ec-cli/blob/70dbabae239e5f1f3ff2b3ef6582c0a604814c80/internal/policy/output.go#L35) from Conftest unlike image and attestation signature validations which hold [the flag and message](https://github.com/hacbs-contract/ec-cli/blob/70dbabae239e5f1f3ff2b3ef6582c0a604814c80/internal/policy/output.go#L33-L34) in `VerificationStatus`. Something we'll surely revisit as we're loosing quite a bit of fidelity from the policy output with this.

zregvart:
@lcarva please have another look

internal/policy/output.go
```diff
@@ -58,6 +58,28 @@ func (o *Output) SetPolicyCheck(results []output.CheckResult) {
 	o.ExitCode = output.ExitCode(results)
 }
 
+func (o Output) Violations() []string {
+	violations := make([]string, 0, 10)
```
lcarva, line 62: This seems like an arbitrary capacity for the slice. Maybe just use `violations := []string{}` and let the compiler figure out what's a good default value?

lcarva (approved):

zregvart:
Snuck in small refactor for readability and added more tests. Going to merge this once checks pass.

internal/policy/output.go
lcarva: Thank you for the reference link. Very helpful.

As you explained, `10` is not an arbitrary number. It's a good enough guess based on your understanding of how the slice will be used. That's fine. For posterity, it would be nice if there was a comment in the code explaining the reasoning for choosing that value, and also the use of a named variable to better express intent.

PR #66, opened 2022-07-07
WIP Logging

simonbaird:
The tests are failing, but I want to share the WIP since I'm new to go and not sure what I'm doing.

That said I think the logging is working fairly nicely so far.

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -180,43 +180,60 @@ func determineInputSpec(arguments args) (*appstudioshared.ApplicationSnapshotSpe
 	var appSnapshot appstudioshared.ApplicationSnapshotSpec
 
 	// read ApplicationSnapshot provided as a file
-	if arguments.filepath != "" {
+	filepath := arguments.filepath
+	if filepath != "" {
+		utils.Trace("Reading application snapshot from %s", filepath)
 		j, err := ioutil.ReadFile(arguments.filepath)
 		if err != nil {
+			utils.Error("Failed to read application snapshot from %s")
 			return nil, err
 		}
 
+		utils.Trace("Parsing application snapshot data from %s", filepath)
 		err = json.Unmarshal(j, &appSnapshot)
 		if err != nil {
+			utils.Error("Unable to parse json snapshot data in %s", filepath)
 			return nil, err
 		}
 
+		utils.Info("Application snapshot loaded from %s", filepath)
+		utils.Trace("Application snapshot:\n  %s", appSnapshot)
 		return &appSnapshot, nil
 	}
 
 	// read ApplicationSnapshot provided as a string
-	if arguments.input != "" {
+	input := arguments.input
+	if input != "" {
+		utils.Info("Reading application snapshot from input param")
 		// Unmarshall json into struct, exit on failure
-		if err := json.Unmarshal([]byte(arguments.input), &appSnapshot); err != nil {
+		if err := json.Unmarshal([]byte(input), &appSnapshot); err != nil {
+			utils.Error("Unable to parse json snapshot data from input param")
 			return nil, err
 		}
 
+		utils.Info("Application snapshot loaded from input param")
+		utils.Trace("Application snapshot:\n  %s", appSnapshot)
 		return &appSnapshot, nil
 	}
 
 	// create ApplicationSnapshot with a single image
-	if arguments.imageRef != "" {
-		return &appstudioshared.ApplicationSnapshotSpec{
+	imageRef := arguments.imageRef
+	if imageRef != "" {
+		utils.Info("Generating application snapshot from image ref")
+		result := &appstudioshared.ApplicationSnapshotSpec{
 			Components: []appstudioshared.ApplicationSnapshotComponent{
 				{
 					Name:           "Unnamed",
 					ContainerImage: arguments.imageRef,
 				},
 			},
-		}, nil
+		}
+		utils.Trace("Application snapshot:\n  %s", *result)
+		return result, nil
 	}
 
-	return nil, errors.New("neither ApplicationSnapshot nor image reference provided to validate")
+	utils.Error("Not able to find or create an application snapshot!")
+	return nil, errors.New("Unable to load an application snapshot to validate")
 }
 
 func validateImage(ctx context.Context, imageRef, policyConfiguration, publicKey, rekorURL string) (*policy.Output, error) {
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -27,6 +27,7 @@ import (
 	"github.com/hacbs-contract/ec-cli/internal/applicationsnapshot"
 	"github.com/hacbs-contract/ec-cli/internal/image"
 	"github.com/hacbs-contract/ec-cli/internal/policy"
+	"github.com/hacbs-contract/ec-cli/internal/utils"
 	"github.com/hashicorp/go-multierror"
 	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
 	"github.com/spf13/cobra"
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -22,7 +22,12 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+
+	"github.com/hacbs-contract/ec-cli/internal/utils"
 )
+
+// The default log level
+var logLevel = 1
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
@@ -32,6 +37,17 @@ var rootCmd = &cobra.Command{
 
 	// Prevent usage instructions showing on every error
 	SilenceUsage: true,
+
+	// Initialize logging
+	PersistentPreRun: func(cmd *cobra.Command, _ []string) {
+		utils.InitLogging(logLevel)
+	},
+}
+
+func init() {
+	rootCmd.PersistentFlags().IntVarP(&logLevel, "log-level", "l", logLevel,
+		"controls how much log output (to stderr) is produced.\n"+
+			"3 is most verbose. -1 suppresses all log output.")
 }
 
 const globalTimeout = 5 * time.Minute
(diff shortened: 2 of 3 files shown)


internal/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+// A logLevel of -1 means silent mode with all logging
+// suppressed, even error messages.
+//
+func InitLogging(logLevel int) {
+	errorLogger = initLogger("ERROR", 0, logLevel)
+	warnLogger = initLogger("WARNING", 1, logLevel)
+	infoLogger = initLogger("INFO", 2, logLevel)
+	traceLogger = initLogger("TRACE", 3, logLevel)
+
+	Trace("Log level set to %d", logLevel)
```
simonbaird, line 45: :shrug:

internal/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+		msg = fmt.Sprintf(msg, params...)
+	}
+	// The first arg here is the number of call stack frames to skip when
+	// deciding what file and line number should appear. 3 seems to be right.
+	// (Seems like there's no way to set it globally so that's why we must
+	// use logger.Output instead of the nicer functions like logger.Printf etc.)
+	//
+	err := logger.Output(3, msg)
+	if err != nil {
+		panic(err)
```
simonbaird, line 84: Added because `make lint` complained the error was being ignored. Is panic good enough here?

internal/utils/logger.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Not being used currently, but how do we feel about it..?
+// Do any finalizers or teardowns get skipped if we just exit like this?
+// Is it really useful to return err through 16 levels of call stack???
+//
+func Fatal(msg string, params ...any) {
+	logHelper(errorLogger, msg, params...)
+	os.Exit(1)
+}
```
simonbaird, line 66: Could also have a Panic function here maybe.

internal/utils/logger.go
robnester-rh: lint shouldn't, from what I understand, care if you explicitly ignore the error like this:

`_ = logger.Output(3, msg)`

cmd/eval.go
```diff
@@ -179,43 +180,60 @@ func determineInputSpec(arguments args) (*appstudioshared.ApplicationSnapshotSpe
 	var appSnapshot appstudioshared.ApplicationSnapshotSpec
 
 	// read ApplicationSnapshot provided as a file
-	if arguments.filepath != "" {
+	filepath := arguments.filepath
+	if filepath != "" {
```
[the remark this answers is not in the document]
simonbaird: I think maybe that's a cleanup/refactor for a future PR.

internal/utils/logger.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+)
+
+var (
```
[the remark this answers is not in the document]
simonbaird: Nice.

zregvart:
Since logging is static, in the sense one defines a level for logging and that level doesn't change in runtime, I would go with an approach that I think is a bit simpler and has less of a impact in runtime. Something like (writing from the top of my head):

```go
type Logger func(format string, v ...any)

var Trace = nop
var Warn = nop
//...

func nop(format string, v ...any) {
}

func logWithPrefix(prefix string) Logger {
	return func(format string, v ...any) {
		log.Printf(prefix + ": " + format, v)
	}
}

func InitLogging(logLevel int) {
	if logLevel >= 3 {
		Trace = logWithPrefix("TRACE")
	}
	if logLevel >= 2 {
		Trace = logWithPrefix("WARN ")
	}
	// ...
}
```

We should also use constants for `logLevel` above and initialize the logger with flags/where to write to and such, this is just to illustrate the idea.

zregvart:
The tests will panic with because `*Logger` variables in `logger.go` are initialized to `nil` and without a call to `InitLogging` they will stay `nil`. We don't want to entangle log initialization with tests (or even cobra). One approach might be to use the [flag package](https://pkg.go.dev/flag@go1.18.3) and determine the level in a `init` method implicitly from that. This would also allow us to configure logging via other means (say an `EC_LOG_LEVEL` environment variable), and would make sure that logging is initialized without necessitating the call to `InitLogging`.

internal/utils/logger.go
```diff
@@ -0,0 +1,94 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package utils
```
zregvart, line 17: I think it would make sense to put this in `internal/logging`, would make the calls more explicit e.g. `logging.Info`.

cmd/eval.go
```diff
@@ -179,43 +180,60 @@ func determineInputSpec(arguments args) (*appstudioshared.ApplicationSnapshotSpe
...
+		utils.Trace("Reading application snapshot from %s", filepath)
 		j, err := ioutil.ReadFile(arguments.filepath)
 		if err != nil {
+			utils.Error("Failed to read application snapshot from %s")
```
zregvart, line 188: I think we want to return error here and log/report errors centrally, this is the intent of [HACBS-633](https://issues.redhat.com/browse/HACBS-633), so the line below could be:

```go
		return nil, ecerrors.FE001.CausedBy(err).Params(filepath)
```

Where `ecerrors.FE001` would already be initialized with the message `Failed to read application snapshot from %s"`.

cmd/root.go
```diff
@@ -22,13 +22,32 @@ import (
...
+	SilenceUsage: true,
+
+	// Initialize logging
+	PersistentPreRun: func(cmd *cobra.Command, _ []string) {
+		utils.InitLogging(logLevel)
+	},
+}
+
+func init() {
+	rootCmd.PersistentFlags().IntVarP(&logLevel, "log-level", "l", logLevel,
```
zregvart, line 48: `logLevel` here is the default value when the flag is not provided

cmd/root.go
```diff
@@ -22,13 +22,32 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+
+	"github.com/hacbs-contract/ec-cli/internal/utils"
 )
 
+// The default log level
+var logLevel = 1
```
zregvart, line 30: I'd prefer if we use constants instead of magic numbers; see also the comment about doing this in `logger.go`.

cmd/root.go
```diff
@@ -22,13 +22,32 @@ import (
...
+
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
 	Use:   "ec",
 	Short: "Tool to enforce enterprise contracts",
 	Long:  `TODO: description`,
+
+	// Prevent usage instructions showing on every error
+	SilenceUsage: true,
```
zregvart, line 39: :+1: I thought it was a good incentive (seeing the usage being printed for every error) to do error handling ([HACBS-633](https://issues.redhat.com/browse/HACBS-633)), but we can do this as well even though error handling might not result in errors reaching cobra

internal/utils/logger.go
zregvart: Well, there should only be errors from ec that surface to the end users, at each point we do `if err != nil` we need to introduce a specific enough error [code, message], and/or handle the `err` in that spot with specificity it requires. Errors then end up being one level deep. That is why [HACBS-633](https://issues.redhat.com/browse/HACBS-633) is important.
We might not want to exit with the same status for every case, we might not want any `os.Exit` calls other than the ones in the error handling logic.

internal/utils/logger.go
zregvart: I would not panic from logging or emit new errors. We could test in initialization if logging works and error from there, and at this point we might want to either silently fail or try outputting directly to `stderr` that the logging failed and further log messages might not appear.

cmd/eval.go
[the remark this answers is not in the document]
zregvart: I prefer not to use `len` for strings, given that types are commonly inferred the `len` has different meaning for strings, slices or maps whereas `str != ""` has only one explicit meaning. To me reading `len(x)` implies that `x` is either _slice or map_ with some elements in it, and reading `x != ""` implies that `x` needs to be a non empty _string_. I find that being explicit with the intent is important for readability of the code.

internal/utils/logger.go
zregvart: I also need to note here that the shed needs to be blue.

cmd/root.go
simonbaird: Yep. I think that's a common cobra pattern.

cmd/root.go
simonbaird: +1

cmd/root.go
simonbaird: It's done in #67 I noticed.

simonbaird:
Thanks for all the suggestions. I'll keep working on it.

cmd/eval.go
```diff
@@ -179,43 +180,60 @@ func determineInputSpec(arguments args) (*appstudioshared.ApplicationSnapshotSpe
 	var appSnapshot appstudioshared.ApplicationSnapshotSpec
 
 	// read ApplicationSnapshot provided as a file
-	if arguments.filepath != "" {
+	filepath := arguments.filepath
+	if filepath != "" {
+		utils.Trace("Reading application snapshot from %s", filepath)
```
lcarva, line 185: Consider using `%q` instead of `%s` when the value is a string. This will quote the value making it easier to read. 
```
Reading application snapshot from yolo
```
vs
```
Reading application snapshot from "yolo"
```
I find it's helpful to diagnose issues caused by extra white spaces.

internal/utils/logger.go
lcarva: FYI, in most cases, `panic` is not the answer: https://go.dev/doc/effective_go#panic

internal/utils/logger.go
```diff
@@ -0,0 +1,94 @@
+// Copyright 2022 Red Hat, Inc.
```
lcarva, line 1: I'm surprised we need to have this file at all. Isn't there a logging library that we can use instead?

simonbaird:
Replaced by #72 which is still WIP.

PR #69, opened 2022-07-08
Create signoff methods for builds

joejstuart:
validate the image signature and attestation

from the validated attestation determine the signoff source

get the signoff from the source

### DIFF
--- a/internal/image/build.go
+++ b/internal/image/build.go
@@ -90,12 +90,25 @@ func (a *attestation) NewSignOffSource() (signOffSource, error) {
 
 // get the last commit used for the component build
 func (a *attestation) getBuildCommitSha() string {
-	return a.getMaterialsString("CHAINS-GIT_COMMIT")
+	digest := a.getMaterialsMap("digest")
+	if digest != nil {
+		return digest["sha"]
+	}
+	return ""
 }
 
 // the git url used for the component build
 func (a *attestation) getBuildSCM() string {
-	return a.getMaterialsString("CHAINS-GIT_URL")
+	return a.getMaterialsString("uri")
+}
+
+func (a *attestation) getMaterialsMap(key string) map[string]string {
+	materials := a.Predicate.Materials
+	// materials is an array. If it's empty or greater than 1, return nothing
+	if len(materials) != 1 {
+		return map[string]string{}
+	}
+	return materials[0][key].(map[string]string)
 }
 
 func (a *attestation) getMaterialsString(key string) string {
@@ -104,7 +117,7 @@ func (a *attestation) getMaterialsString(key string) string {
 	if len(materials) != 1 {
 		return ""
 	}
-	return a.Predicate.Materials[0][key].(string)
+	return materials[0][key].(string)
 }
 
 // returns the signOff signature and body of the source
--- a/internal/image/build.go
+++ b/internal/image/build.go
@@ -45,60 +45,33 @@ type attestation struct {
 	Type          string                   `json:"_type"`
 }
 
-type buildSigner interface {
-	GetBuildSignOff() (*signOffSignature, error)
-}
-
-type commitSignoffSource struct {
+type commitSignOff struct {
 	source    string
 	commitSha string
 }
 
-type jiraSignoffSource struct {
-	source string
-	jiraid string
-}
-
-type tagSignoffSource struct {
-	source    string
-	tag       string
-	commitSha string
+type signOffSource interface {
+	GetSignOff() (*signOffSignature, error)
+	getSource() (interface{}, error)
 }
 
 type signOffSignature struct {
-	Payload interface{} `json:"payload"`
-	Source  string      `json:"source"`
-}
-
-// mocking
-type sourceRepository interface {
-	getRepository(string) (*git.Repository, error)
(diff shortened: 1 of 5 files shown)


joejstuart:
Putting this up now for feedback, but still writing tests.

zregvart (changes requested):
Looks like there are bits missing here so I might be inferring this incorrectly. This seems to confuse the ownership of a source code change with signing off on a built artifact. Those are two different things. The ownership of the source code is asserted by adding the `Signed-off-by` commit message trailer, you can read about this [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst#n367). Folk use the `Signed-off-by` on regular basis when authoring changes and sharing those as contributions to an upstream project, we cannot base the built artifact sign off on these. See [Developer certificate of origin](https://wiki.linuxfoundation.org/dco) (DCO),

zregvart (commented):
For the sake of demoing this let's keep at it to collect feedback

zregvart:
For whatever reason the GitHub Workflow cache keeps getting corrupted, I've cleared the cache and rerun the lint workflow which now points to a codestyle issue with linting imports. @joejstuart can you rebase to pickup #80 and then run `make lint-fix` to sort those imports?

internal/image/build.go
```diff
@@ -0,0 +1,157 @@
...
+	signatureHeader := "Signed-off-by"
+	// loop over each line of the commit message looking for "Signed-off-by"
+	for _, line := range strings.Split(message, "\n") {
+		regex := fmt.Sprintf("^%s", signatureHeader)
+		match, _ := regexp.MatchString(regex, line)
+		// if there's a match, split on "Signed-off-by", then capture each signature after
+		if match {
+			results := strings.Split(line, signatureHeader)
+			for _, signature := range strings.Split(results[len(results)-1], ",") {
+				sigRegex := regexp.MustCompile("([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+.[a-zA-Z0-9_-]+)")
```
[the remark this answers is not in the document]
zregvart: This allows only for domains that have a single dot in the domain part; well, the `.` here is not escaped so `"a@aaa"` matches and so does `a@a..a`.
There is [a lot more nuance](https://www.regular-expressions.info/email.html) to creating a [regexp for email](https://www.emailregex.com/).
My personal grievance is that this doesn't allow the `+` in the local-part with which I struggle on 90% of websites today.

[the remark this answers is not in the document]
zregvart: Noting that there is [mail.ParseAddress](https://pkg.go.dev/net/mail#ParseAddress) which doesn't use regular expressions and needs to be fed the rest of the `Signed-off-by: ` line. Which [could be parsed out](https://pkg.go.dev/github.com/libgit2/git2go/v33@v33.0.9#MessageTrailers) using the libgit2 wrapper.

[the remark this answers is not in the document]
joejstuart: Thanks for the suggestions. I switched to using mail.ParseAddressList, but had some issues with libgit2.

internal/image/build.go
```diff
@@ -0,0 +1,175 @@
...
+			commitSha: commitSha,
+		}, nil
+	}
+	return nil, nil
+}
+
+// get the last commit used for the component build
+func (a *attestation) getBuildCommitSha() string {
+	if len(a.Predicate.Materials) == 1 {
+		return a.Predicate.Materials[0].Digest["sha"]
```
lcarva, line 100: I think this might be `sha1`.

internal/image/build.go
```diff
@@ -0,0 +1,175 @@
...
+	return &commit{
+		Sha:     gitCommit.Hash.String(),
+		Author:  fmt.Sprintf("%s <%s>", gitCommit.Author.Name, gitCommit.Author.Email),
+		Date:    gitCommit.Author.When.String(),
+		Message: gitCommit.Message,
+	}, nil
+
+}
+
+// parse a commit and capture signatures
```
lcarva, line 150: Let's update this comment to say "sign offs" instead of "signatures". Same for the comments within this function.

internal/image/build.go
```diff
@@ -0,0 +1,175 @@
...
+	Parameters   map[string]string      `json:"parameters"`
+	Environment  map[string]interface{} `json:"environment"`
+}
+
+type materials struct {
+	Uri    string            `json:"sha"`
+	Digest map[string]string `json:"digest"`
+}
+
+type predicate struct {
```
lcarva, line 43: Wouldn't this be sufficient?
```go
type predicate struct {
	Materials   []materials            `json:"materials"`
}
```

Same for `attestation`. I think we can remove the fields we don't care about.

internal/image/build.go
```diff
@@ -0,0 +1,175 @@
...
+			if err != nil {
+				continue
+			}
+			for _, signature := range signatures {
+				capturedSignatures = append(capturedSignatures, signature.Address)
+			}
+		}
+	}
+
+	if len(capturedSignatures) > 0 {
```
lcarva, line 171: Wouldn't `return capturedSignatures` work in either case?

PR #72, opened 2022-07-09
Add some debug logging

simonbaird:
Use logrus for logging.

I'm just using ~~warns and~~ debugs here since I'm not sure what the
plan is for error handling, so rather than mix in error messages
with the existing error handling, I'll let that be a thing to figure
out later.

There are a few extra minor fixes and tweaks in this PR now,
see commit messages for explanations.


zregvart (approved):
Looks good

simonbaird:
Code coverage report is informational only, tests and acceptance tests are passing, so maybe that's good enough for a fast merge. :sweat_smile:

PR #74, opened 2022-07-11
Refactor `kubernetes_client` to `kubernetes`

zregvart:
Also renames `Kubernetes` to `Client`, so now we can use `kubernetes.Client` instead of `kubernetes_client.Kubernetes`.

### DIFF
--- a/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
+++ b/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
@@ -38,7 +38,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/hacbs-contract/ec-cli/internal/evaluator"
-	"github.com/hacbs-contract/ec-cli/internal/kubernetes_client"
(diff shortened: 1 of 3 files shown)


internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -86,9 +86,9 @@ func NewApplicationSnapshotImage(ctx context.Context, image string, publicKey st
 		checkOpts: checkOpts,
 	}
 
-	k8s, err := kubernetesCreator()
+	k8s, err := kubernetesClientCreator()
```
lcarva, line 89: Why don't we just call `kubernetes.NewClient` directly? Easier to mock in tests?

lcarva (approved):

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
zregvart: We need this in tests, though that test was removed in a recent refactor. I'm working on bringing it back in.

PR #79, opened 2022-07-12
Hide Conftest implementation details

zregvart:
This introduces the `Evaluator` interface so we don't depend (so much) on the implementation of `conftestEvaluator`, which is now private with private fields. As `context.Context` can vary between calls, e.g. if we want to run image and attestation signature validation concurrently it needs to be passed as function parameter rather than stored in the evaluator implementation.

### DIFF
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"path/filepath"
 
+	"github.com/open-policy-agent/conftest/output"
 	"github.com/open-policy-agent/conftest/runner"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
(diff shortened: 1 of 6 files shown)


internal/evaluator/conftest_evaluator.go
```diff
@@ -49,21 +49,20 @@ type ConfigurationPaths struct {
 }
 
 // NewConftestEvaluator returns a properly initialized ConftestEvaluator for usage
```
lcarva, line 51: Let's update the comment with the correct capitalization of `conftestEvaluator`.

internal/evaluator/evaluator.go
```diff
@@ -0,0 +1,28 @@
...
+package evaluator
+
+import (
+	"context"
+
+	"github.com/open-policy-agent/conftest/output"
+)
+
+type Evaluator interface {
+	// TODO refactor not to expose Conftest type here
```
lcarva, line 26: I think this `TODO` is already addressed.

robnester-rh:
[internal/evaluation_target/application_snapshot_image/application_snapshot_image.go](https://github.com/hacbs-contract/ec-cli/pull/79/files#diff-c89fbe58faac75c0a251f1b8af58b7789adc10532e0814ad59ef0adc47d91445) needs updated.

internal/evaluator/evaluator.go
zregvart: We're returning `github.com/open-policy-agent/conftest/output.CheckResult`, that ties any calling code to conftest.

zregvart:
> [internal/evaluation_target/application_snapshot_image/application_snapshot_image.go](https://github.com/hacbs-contract/ec-cli/pull/79/files#diff-c89fbe58faac75c0a251f1b8af58b7789adc10532e0814ad59ef0adc47d91445) needs updated.

Missed committing that change

robnester-rh (approved):
LGTM

PR #80, opened 2022-07-13
Sort imports using gci

zregvart:
This configures golangci linter to use gci to enforce sorting of imports and adds invocation of gci to the `lint-fix` target to reformat imports to the new convention.


robnester-rh (approved):
LGTM

PR #81, opened 2022-07-13
Perform git clone with context

zregvart:
Cloning a repository is potentially an time consuming/expensive operation, we should adhere to the timeout limitations of the configured context when performing it. This propagates the context of the command to the git cloning code.

### DIFF
--- a/internal/policy/source/repo.go
+++ b/internal/policy/source/repo.go
@@ -17,6 +17,7 @@
 package source
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"strings"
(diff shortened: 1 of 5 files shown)


internal/policy/source/repo.go
```diff
@@ -30,12 +31,12 @@ import (
 )
 
 //CheckoutRepo is used as an alias for git.PlainClone in order to facilitate testing
```
robnester-rh, line 33: Update docstring with new call

robnester-rh (approved):
LGTM

lcarva (approved):

PR #82, opened 2022-07-13
Support single namespace

zregvart:
We only support a single Rego namespace, there is no need to pass in multiple namespaces.


robnester-rh (approved):
LGTM

PR #83, opened 2022-07-13
Fix the digest of cosign module

lcarva:


lcarva:
Use `GOPROXY="https://proxy.golang.org,direct"` to get the expected version instead.

PR #84, opened 2022-07-13
Add track bundle subcommand

lcarva:

### DIFF
--- a/internal/image/process.go
+++ b/internal/image/process.go
@@ -1,0 +1,113 @@
+// Copyright 2022 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package image
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/hashicorp/go-multierror"
+)
+
+// ParseAndResolveAll parses each of the urls into an ImageReference object. The digest
+// is resolved for each reference if needed.
+func ParseAndResolveAll(urls []string) ([]ImageReference, error) {
+	var errs error
+
+	refs := make([]ImageReference, 0, len(urls))
+	for _, url := range urls {
+		ref, err := newImageReference(url)
+		if err != nil {
+			errs = multierror.Append(errs, err)
+			continue
+		}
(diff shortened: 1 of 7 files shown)


internal/image/process.go
```diff
@@ -0,0 +1,113 @@
...
+	digest := descriptor.Digest.String()
+	if digest == "" {
+		return nil, fmt.Errorf("digest for image %q is empty", i.ref.String())
+	}
+
+	return newimageReference(fmt.Sprintf("%s:%s@%s", i.Repository, i.Tag, digest))
+}
+
+// newimageReference returns an ImageReference instance based on the given url.
+func newimageReference(url string) (*ImageReference, error) {
```
zregvart, line 89: nit
```suggestion
func newImageReference(url string) (*ImageReference, error) {
```

internal/image/process.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return nil, err
+	}
+	imageRef := &ImageReference{ref: ref}
+
+	// An image reference may contain both a tag and a digest. However, the parsing library
+	// will drop the tag in favor of the digest. Since we care about the value of the tag,
+	// parse the url without the digest reference to force the library to retain the tag
+	// value in all cases where it is present.
+	tagRef, err := name.NewTag(strings.Split(url, "@")[0])
```
zregvart, line 100: Are we certain that the `url` contains `@`?

zregvart (approved):
Looks good

internal/image/process.go
lcarva: I'm certain that it may not contain `@` and that's perfectly [valid](https://pkg.go.dev/strings#Split) here:
> If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s. 

Because sep is not empty `@`, the returned slice will always have at least 1 item so the index access should be safe here.

internal/image/process.go
lcarva: Nice catch! Fixed.

internal/image/process.go
zregvart: Perhaps like this?

```suggestion
	_, tag, hasAt := strings.Cut(url, "@")
	if !hasAt {
		// error?
	}
	tagRef := name.NewTag(tag)
```

zregvart: > Because sep is not empty @, the returned slice will always have at least 1 item so the index access should be safe here.

Are we then creating a `tagRef` equal to `url`, is that an issue?

lcarva: I don't think we should allow an implicit tag. We don't know the tag for `registry.com/repo@sha256:123`. Assuming it's `latest` seems incorrect based on the examples I'm aware of. That's the bug you caught :smile:  In such cases, we have two options, a) do not allow such image refs, or b) allow tag to be an empty string.  For now (b) seems like a good option. I changed it as such.

The fix was to make sure the calls to `NewTag` and `NewDigest` are also using `StrictValidation`. By doing so, the call to `NewTag` with `registry.com/repo` returns an error, thus the default value of `imageRef.Tag` is never overwritten.

lcarva: This unit test proves that it works as described: https://github.com/hacbs-contract/ec-cli/pull/84/files#diff-6b4f378d4649b351387b7ebb463c1ede12a55bb309846e48043435e01a058b41R69-R81

zregvart (approved):
Merge!

PR #93, opened 2022-07-15
Adds acceptance test for `track bundle` subcommand

zregvart:
Currently only the trivial case is explored, whereby an image hash is tracked from the registry. The feature mentions Tasks and Pipelines but that doesn't affect the created image in the stub registry -- it's only a random image that gets placed there.


zregvart:
Needs to be rebased on #94

features/track_bundle.feature
```diff
@@ -0,0 +1,24 @@
+Feature: track bundles
+  The ec command line can track bundles and generate tracking data files
+
+  Background:
+    Given stub registry running
+
+  Scenario:
+    Given an tekton bundle image named "acceptance/bundle:tag" containing
```
lcarva, line 8: s/an/a/

features/track_bundle.feature
```diff
@@ -0,0 +1,24 @@
...
+      | type     | name      |
+      | Task     | task1     |
```
lcarva, line 10: These are currently not used, is that right?

zregvart: Yeah, I wanted to remove the `--type` parameter, but thought to it would be best to create a scenario for the current state, so I stopped myself before that...

lcarva (approved):

PR #95, opened 2022-07-15
Support reading non-blocking data from ecp.Spec

robnester-rh:
This commit adds support for reading the non-blocking data from the
ecp.Spec.Exceptions in order to write this to the data json file used in
the conftestEvaluator creation.


internal/evaluator/conftest_evaluator.go
```diff
@@ -97,16 +99,37 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) ([]out
...
 // addDataPath adds the appropriate data path to the ConfigurationPaths DataPaths field array.
-func (c *conftestEvaluator) addDataPath() error {
-	// Todo: Read from epc.Spec and right the nonblocking to this file.
+func (c *conftestEvaluator) addDataPath(spec *v1alpha1.EnterpriseContractPolicySpec) error {
 	dataDir := filepath.Join(c.workDir, "data")
 	exists, err := afero.DirExists(utils.AppFS, dataDir)
 	if err != nil {
 		return err
 	}
 	if !exists {
```
lcarva, line 108: The data dir will always exist since https://github.com/hacbs-contract/ec-policies/pull/63. I think we need to change this logic around a bit. Maybe something like:

1. Create data dir if it doesn't exist (in case a different policy repo is used)
2. Always write out the `data.json` file. I don't think it makes much sense to ever skip this because skipping means an inconsistent result is reported.

robnester-rh: Ok. I'll tweak this to write the file but perhaps add a random suffix to the "data" portion of the name to avoid overwriting an existing data.json file perhaps?

robnester-rh: So we wouldn't have a scenario where we'd want to add to non-blocking items already present in a different policy repo?

lcarva: I don't think so but I'm not very creative. We could log a warning if a file already exists right before we overwrite it?

We can change this behavior in the future if we come up with a use case for it. wdyt?

robnester-rh: Agreed. I skipped logging as that can be added when we find a use case that needs it.

lcarva (approved):

PR #98, opened 2022-07-18
Always show report when validating image

lcarva:
https://issues.redhat.com/browse/HACBS-881

### DIFF
--- a/cmd/validate_image.go
+++ b/cmd/validate_image.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"sync"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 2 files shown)


zregvart:
I think this goes against the caller's intent. If the caller choose to output the results to a file that indicates that they do not want to see the output on the standard out. If the output needs to be both in the file and on the standard out, they can use `tee` and omit `--output-file`.

lcarva:
> I think this goes against the caller's intent. If the caller choose to output the results to a file that indicates that they do not want to see the output on the standard out. If the output needs to be both in the file and on the standard out, they can use `tee` and omit `--output-file`.

An older version of me agrees with you.

The problem outlined in [HACBS-881](https://issues.redhat.com/browse/HACBS-881) has to do when using the ec-cli from a Tekton Task. I think we can adjust the task to address the problem without any changes to the ec-cli.

PR #104, opened 2022-07-19
Image commit validator

joejstuart:


cmd/sign_off.go
```diff
@@ -0,0 +1,119 @@
...
+
+func signOffCmd() *cobra.Command {
+	var data = struct {
+		imageRef  string
+		publicKey string
+		filePath  string
+		input     string
+		spec      *appstudioshared.ApplicationSnapshotSpec
+	}{
+		imageRef:  "",
```
robnester-rh, line 41: You don't have to initialize struct fields to their "zero" value in Go.

cmd/sign_off.go
```diff
@@ -0,0 +1,119 @@
...
+		publicKey: "",
+	}
+	cmd := &cobra.Command{
+		Use:   "sign-off",
+		Short: "Capture signed off signatures from a source (github repo, Jira)",
+		Long: `Supported sign off sources are commits captured from a git repo and jira issues.
+               The git sources return a signed off value and the git commit. The jira issue is
```
robnester-rh, line 48: nit: line up with left edge of line above.

internal/applicationsnapshot/input.go
```diff
@@ -0,0 +1,61 @@
+package applicationsnapshot
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/hacbs-contract/ec-cli/internal/utils"
```
robnester-rh, line 7: Imports are not sorted. Ensure that imports from hacbs-contract are placed last. Run `make lint-fix` or manually correct.

internal/image/build.go
```diff
@@ -0,0 +1,175 @@
...
+	"net/mail"
+	"regexp"
+	"strings"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/storage/memory"
+)
+
+var gitClone = git.Clone
```
robnester-rh, line 30: You may want to consider `git.PlainCloneContext` here.

PR #109, opened 2022-07-21
Determine bundle content from the bundle image

zregvart:
This removes the `--type` parameter from the `track bundle` command, and
instead determines the type of content within the bundle by examining
the bundle image.

Also includes #108 so acceptance tests are not flaky.


Makefile
```diff
@@ -47,7 +47,7 @@ build: dist/ec_$(shell go env GOOS)_$(shell go env GOARCH) ## Build the ec binar
 
 .PHONY: test
 test: ## Run unit tests
-	@go test -race -covermode=atomic -coverprofile=coverage-unit.out -short -timeout 500ms ./...
+	@go test -race -covermode=atomic -coverprofile=coverage-unit.out -short -timeout 30s ./...
```
lcarva, line 50: Can we add a comment here explaining the 30s value? The PR description does a good job at explaining and also mentions that this should be reverted back to 500ms at some point. That will not be obvious to someone looking at this file.

cmd/track_bundle.go
```diff
@@ -91,17 +89,41 @@ func trackBundleCmd(track trackBundleFn) *cobra.Command {
...
+	if err != nil {
+		return nil, err
+	}
+
+	manifest, err := img.Manifest()
+	if err != nil {
+		return nil, err
+	}
+
+	collectionsMap := map[string]any{}
```
lcarva, line 116: `map[string]bool{}` ?

cmd/track_bundle_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/tektoncd/pipeline/pkg/remote/oci"
+
+	"github.com/hacbs-contract/ec-cli/internal/image"
+)
+
+func Test_tektonBundleCollector(t *testing.T) {
+	defaultFetchImage := fetchImage
+	t.Cleanup(func() {
```
lcarva, line 39: Will this execute after each call to `t.Run` below? I can't tell if that's what the [docs](https://pkg.go.dev/testing#T.Cleanup) mean:
> Cleanup registers a function to be called when the test (or subtest) and all its subtests complete. Cleanup functions will be called in last added, first called order.

cmd/track_bundle_test.go
```diff
@@ -0,0 +1,99 @@
...
+			name: "failed to fetch image",
+			err:  errors.New("expected"),
+		},
+		{
+			name:     "a task and a pipeline",
+			kinds:    []string{"task", "pipeline"},
+			expected: []string{"task", "pipeline"},
+		},
+		{
+			name:     "multiple tasks and a pipelines",
```
lcarva, line 59: `multiple tasks and multiple pipelines`

cmd/track_bundle_test.go
```diff
@@ -0,0 +1,99 @@
...
+			kinds:    []string{"task", "pipeline", "pipeline", "task", "task"},
+			expected: []string{"task", "pipeline"},
+		},
+	}
```
lcarva, line 63: Do we need to add coverage for just `task` and just `pipeline`? Feels odd to not have the main use case.

cmd/track_bundle_test.go
```diff
@@ -0,0 +1,99 @@
...
+			fetchImage = func(name name.Reference, _ ...remote.Option) (v1.Image, error) {
+				if c.err != nil {
+					return nil, c.err
+				}
+
+				img := empty.Image
+
+				var err error
+				for _, k := range c.kinds {
+					if img, err = mutate.Append(img, mutate.Addendum{
```
lcarva, line 76: Cool!

internal/tracker/tracker_test.go
```diff
@@ -158,10 +167,107 @@ task-bundles:
...
 			assert.Equal(t, tt.output, string(output))
 		})
 	}
 
 }
+
+func TestTrackerAddKeepsOrder(t *testing.T) {
+	parameters := gopter.DefaultTestParameters()
+
+	// uncomment when test fails and set the seed to the printed value
```
lcarva, line 183: Is this leftover?

internal/tracker/tracker_test.go
```diff
@@ -158,10 +167,107 @@ task-bundles:
...
+	properties.Property("collections are sorted", arbitraries.ForAll(
+		func(records []Record) bool {
+			tracker := Tracker{}
+			for _, r := range records {
+				tracker.add(r)
+			}
+
+			raw, err := tracker.Output()
+			if err != nil {
+				panic(err)
```
lcarva, line 209: Maybe something like `t.Fatal(...)` is better here?

lcarva: Oh maybe it can't because this function is not being called by golang's testing.

lcarva (commented):
Looks great. Just some really minor comments.

internal/tracker/tracker_test.go
zregvart: When generative tests fail they print the random seed, with this random seed one can reproduce that exact case, this is code that does that, it needs to be uncommented and the seed value needs to be placed, I thought the comment explained that

internal/tracker/tracker_test.go
zregvart: Yeah, we want to [panic out of gopter tests](https://pkg.go.dev/github.com/leanovate/gopter#example-package-Panic)

Makefile
zregvart: Updated in #108

cmd/track_bundle_test.go
zregvart: There is [one already](https://github.com/hacbs-contract/ec-cli/pull/109/files#diff-d3aa3c8fc83384b5f3aee62abaafbda79e331eb4e648bdcf66f242eb1c29de7bR54)

internal/tracker/tracker_test.go
lcarva: It makes sense now. Thanks!

cmd/track_bundle_test.go
lcarva: I meant:
```go
		{
			name:     "tasks only",
			kinds:    []string{"task", "task", "task"},
			expected: []string{"task"},
		},
		{
			name:     "pipelines only",
			kinds:    []string{"pipeline", "pipeline", "pipeline"},
			expected: []string{"pipeline"},
		},
```

Probably overkill...

lcarva (approved):

PR #110, opened 2022-07-21
Count partial coverage as a hit

lcarva:
Partial indicates that the source code was not fully executed by the
test suite; there are remaining branches that were not executed.

Given that in most cases these are corner case error handling, let's
count any coverage as full coverage.


lcarva:
Waiting for the codecov comment to see if this actually works.

PR #111, opened 2022-07-21
Add replace command for bundle refs

lcarva:
https://issues.redhat.com/browse/HACBS-897

### DIFF
--- a/cmd/replace.go
+++ b/cmd/replace.go
@@ -1,0 +1,174 @@
+// Copyright 2022 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"os"
+	"regexp"
+
+	"github.com/spf13/cobra"
+
+	"github.com/hacbs-contract/ec-cli/internal/image"
+)
+
+type replaceFn func([]string, string) ([]byte, error)
+
+func replaceCmd(replace replaceFn) *cobra.Command {
+	var data = struct {
+		Source     string
+		Overwrite  bool
+		OutputFile string
+	}{}
+
+	cmd := &cobra.Command{
+		Use:   "replace",
+		Short: "Replace image references in the given input",
+		Example: `ec replace --source <source path> <image uri> [<image uri> ...]
+
+# replace all occurences of an image reference in source file
+ec replace --source resource.yaml <IMAGE>
+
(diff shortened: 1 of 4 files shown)


cmd/replace.go
```diff
@@ -0,0 +1,171 @@
...
+	"regexp"
+
+	"github.com/spf13/cobra"
+
+	"github.com/hacbs-contract/ec-cli/internal/image"
+)
+
+type replaceFn func([]string, string) ([]byte, error)
+
+func replaceCmd(r replaceFn) *cobra.Command {
```
zregvart, line 33: Found the `r(...)` confusing to read

```suggestion
func replaceCmd(replace replaceFn) *cobra.Command {
```

cmd/replace.go
```diff
@@ -0,0 +1,171 @@
...
+	var data = struct {
+		Images     []string
+		Source     string
```
zregvart, line 36: Perhaps?

```suggestion
		InputFile     string
```

cmd/replace.go
```diff
@@ -0,0 +1,171 @@
...
+		Replace    bool
```
zregvart, line 37: Perhaps?
```suggestion
		Overwrite    bool
```

cmd/replace.go
```diff
@@ -0,0 +1,171 @@
...
+	r := replaceCmd(replace)
+	rootCmd.AddCommand(r)
+}
+
+type ImageReplacer struct {
+	*image.ImageReference
+	regex *regexp.Regexp
+}
+
+func (i *ImageReplacer) Match(b []byte) bool {
```
zregvart, line 115: Doesn't need to be public

```suggestion
func (i *ImageReplacer) match(b []byte) bool {
```

cmd/replace.go
```diff
@@ -0,0 +1,171 @@
...
+	return i.regex.Match(b)
+}
+
+func (i *ImageReplacer) Replace(b []byte) []byte {
```
zregvart, line 119: Probably doesn't need to be public

```suggestion
func (i *ImageReplacer) replace(b []byte) []byte {
```

cmd/replace.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return cmd
+}
+
+func init() {
+	r := replaceCmd(replace)
+	rootCmd.AddCommand(r)
+}
+
+type ImageReplacer struct {
```
zregvart, line 110: Can be made private

```suggestion
type imageReplacer struct {
```

zregvart (approved):
:sunglasses: An acceptance test would be nice :)

cmd/replace.go
lcarva: I think I want to keep `Source` because we'll need to support different hinds of sources, e.g. a GitHub repo, a directory, etc. I think it would make the CLI a little bit simpler by having a single parameter, much like the `--key` parameter of `cosign verify`. Wdyt?

zregvart: Yeah, that should fit nicely

robnester-rh (commented):
No additional comments. I agree with @zregvart regarding making methods / structs private unless/until needed to be public.

cmd/replace.go
lcarva: Yes, less confusing.

lcarva:
All comments addressed (either resolved or replied)

> sunglasses An acceptance test would be nice :)
I agree! But I don't think just yet. IMO this needs to take a bit more shape before doing so.

robnester-rh (approved):
LGTM

PR #116, opened 2022-07-26
Replace bundle refs with Tekton Hub data

lcarva:


cmd/replace.go
```diff
@@ -17,40 +17,55 @@
...
+		CatalogHubAPIURL: "https://api.hub.tekton.dev",
+	}
 
 	cmd := &cobra.Command{
 		Use:   "replace",
 		Short: "Replace image references in the given input",
-		Example: `ec replace --source <source path> <image uri> [<image uri> ...]
+		Example: `ec replace --source <source path> [<image uri> ...]
+
+# display a modified version of the source file where
```
zregvart, line 49: We don't start with the comment for examples in other commands

zregvart (approved):
LGTM. Adding acceptance tests would be nice

cmd/replace.go
lcarva: Ack. I wasn't aware. Updated.

PR #122, opened 2022-07-29
Trust Rekor public key from Rekor

zregvart:
In order for Cosign to verify the log entries from Rekor it needs the
Rekor public key. The public key, when using Cosign, can be provided
either via TUF, via `SIGSTORE_REKOR_PUBLIC_KEY` environment variable
that points to a file containing PEM encoded public key, or via
`SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY` environment variable which allows
fetching of the public key via Rekor API.

This opts to use `SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY` environment
variable, because implementing TUF is part of another story (HACBS-724),
setting `SIGSTORE_REKOR_PUBLIC_KEY` (which we have for the acceptance
tests to pass) unconditionally prints to standard error a warning like:

```
**Warning ('SIGSTORE_REKOR_PUBLIC_KEY' is only for testing)
** Using a non-standard public key for Rekor: <key>\n
```

Using `SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY` also is a smaller change.

This updates the acceptance tests not to specify
`SIGSTORE_REKOR_PUBLIC_KEY` and for the stub Rekor API to return the
public key via API.

We also need this for end to end testing the v2 version of the
`verify-enterprise-contract` Tekton Task.

Ref. https://issues.redhat.com/browse/HACBS-762

### DIFF
--- a/internal/acceptance/rekor/rekor.go
+++ b/internal/acceptance/rekor/rekor.go
@@ -66,7 +66,31 @@ func stubRekordRunning(ctx context.Context) (context.Context, error) {
 		return ctx, err
 	}
 
-	return wiremock.StartWiremock(ctx)
+	if state.KeyPair == nil {
+		// not used for any signing, we just need the public key in PEM for the
+		// in-toto schema below
+		keyPair, err := crypto.GenerateKeyPair()
+		if err != nil {
+			return ctx, err
+		}
+
+		state.KeyPair = keyPair
+	}
+
+	ctx, err = wiremock.StartWiremock(ctx)
+	if err != nil {
+		return ctx, err
+	}
+
+	if err = wiremock.StubFor(ctx, wiremock.Get(wiremock.URLPathEqualTo("/api/v1/log/publicKey")).
+		WillReturn(string(state.KeyPair.PublicBytes),
+			map[string]string{"Content-Type": "application/x-pem-file"},
+			200,
+		)); err != nil {
+		return ctx, err
+	}
+
+	return ctx, nil
 }
 
 // randomHex generates a random hex string of the given length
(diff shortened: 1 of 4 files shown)


internal/acceptance/rekor/rekor.go
```diff
@@ -66,7 +66,31 @@ func stubRekordRunning(ctx context.Context) (context.Context, error) {
...
+
+		state.KeyPair = keyPair
+	}
+
+	ctx, err = wiremock.StartWiremock(ctx)
+	if err != nil {
+		return ctx, err
+	}
+
+	if err = wiremock.StubFor(ctx, wiremock.Get(wiremock.URLPathEqualTo("/api/v1/log/publicKey")).
```
lcarva, line 85: Something for the future... In a rekor installation, is the public key auto-generated?

lcarva (approved):

internal/acceptance/rekor/rekor.go
zregvart: Few lines up in that method.

lcarva: I see how it's done here for the acceptance tests. I meant on a real installation.

zregvart: That would be something that Rekor does on it's on. From what I understand it can either [generate the key and hold it in-memory](https://github.com/sigstore/rekor/blob/07ec8fc064309472d62c939c0e04b7b56fac15b7/pkg/signer/memory.go#L36), or use a [GCP KMS reference](https://github.com/sigstore/rekor/blob/07ec8fc064309472d62c939c0e04b7b56fac15b7/pkg/signer/signer.go#L30-L31) (configured via `--rekor_server.signer`).

PR #123, opened 2022-07-29
Add error handling in ec-cli

robnester-rh:
This commit implements error handling in ec-cli. An error var should be
constructed for a given "type" and "indicator" of error and the calling
function should return an `ec_errors.Error` pointer instead of an error
object itself.

### DIFF
--- a/internal/ec_errors/ec_error.go
+++ b/internal/ec_errors/ec_error.go
@@ -1,0 +1,54 @@
+// Copyright 2022 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package ec_errors
+
+type Error struct {
+	code       string
+	message    string
+	cause      string
+	exitStatus int
+}
+
+const (
+	SuccessExitStatus = iota
+	ErrorExitStatus
+	PolicyExitStatus
+)
+
+var (
+	/*
+		Additional errors can be added in a similar way as below
+	*/
+	GE001 = Error{
+		code:       "GE001",
+		message:    "General error",
+		exitStatus: ErrorExitStatus,
+	}
+)
+
+func (e Error) CausedBy(err error) *Error {
+	return &Error{
+		code:       e.code,
+		message:    e.message,
+		cause:      err.Error(),
+		exitStatus: e.exitStatus,
+	}
+}
+
+func (e Error) Error() string {
+	return e.cause
+}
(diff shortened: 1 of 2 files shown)


robnester-rh:
I've gone back and forth and worked through about four different solutions to this, but this solution suggested by @zregvart is probably the most flexible / lightweight so I'm going to just implement it.

internal/ec_errors/ec_error.go
```diff
@@ -0,0 +1,50 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package ec_errors
+
+type Error struct {
```
lcarva, line 19: It would help me understand this more if this could be incorporated in a couple of places in the code base.

internal/ec_errors/ec_error.go
```diff
@@ -0,0 +1,50 @@
...
+	code       string
+	message    string
+	cause      string
+	exitStatus int
```
lcarva, line 23: My main concern with this is scope. A helper function nested in the code base is probably not the right place to decide if a certain error means the process should exit, let alone which status code.

I can see this being used directly in the code under `cmd/*.go`. But using this approach in every function that can return an error seems unnecessarily cumbersome.

internal/ec_errors/ec_error.go
lcarva, line 19: Consider attaching an `Error() string` function to this struct so it implements the [error interface](https://pkg.go.dev/builtin#error).

zregvart: Something like:

```go
if err != nil {
    return errors.GE001.CausedBy(err)
}
```

Though the first followup PR should remove `GE001`, and go through all those `if err != nil` and assign them an Error. I don't think we want any general errors, we want specific errors we can trace back to code.

zregvart (approved):
For the followup PR:

- use this in places where we have `if err != nil` kinds of code, assigning unique error codes as needed
- use the `exitStatus` to set the exit code
- error logging

For some cases, where this is a business logic, for example when validating image signature, I think we want to handle errors differently or we need to incorporate/format those errors for the output JSON.

internal/ec_errors/ec_error.go
lcarva: Thanks! I'm more interested in where in the code base this will done. See my [comment](https://github.com/hacbs-contract/ec-cli/pull/123#discussion_r933531712) about scope.

internal/ec_errors/ec_error.go
zregvart: Most cases we propagate the same error up the stack. I don't see a way around this. Also it is essential to have many different error codes to know where the error originated. Otherwise we'll have very difficult time troubleshooting. When we "handle" an error, we need to generate a new error constant here and at that point we define the exit status as well.

internal/ec_errors/ec_error.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const (
+	SuccessExitStatus = iota
+	ErrorExitStatus
+	PolicyExitStatus
+)
+
+var (
```
zregvart, line 32: Can this be a `const`?

robnester-rh: > Can this be a `const`?

Unfortunately, no.

https://go.dev/ref/spec#Constants

internal/ec_errors/ec_error.go
lcarva: It's an approach worth trying out :+1:

lcarva (approved):

PR #130, opened 2022-08-02
Simple Hugo static site for the documentation

zregvart:
This configures the Hugo static site generator for publishing of the
documentation. The simple `hugo-book` theme suits our purposes for now.


zregvart:
Preview can be seen [here](https://zregvart.github.io/ec-cli/)

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "docs/themes/hugo-book"]
+	path = docs/themes/hugo-book
+	url = ../../alex-shpak/hugo-book
```
lcarva, line 3: Any reason not to use `github.com/alex-shpak/hugo-book`?

internal/documentation/documentation.go
```diff
@@ -17,59 +17,72 @@
...
-	cmd.RootCmd.DisableAutoGenTag = true
+	}()
 
-	// Create the target paths
-	for _, p := range paths {
-		if err := os.MkdirAll(p, os.ModePerm); err != nil {
-			errs = multierror.Append(errs, err)
-			continue
+	// Markdown
+	if *markdown != "" {
```
lcarva, line 50: Could this be a null pointer?

lcarva (approved):

.gitmodules
zregvart: Probably something my git client optimized, switched to full URL...

internal/documentation/documentation.go
zregvart: [Nope](https://pkg.go.dev/flag#String)

PR #136, opened 2022-08-04
Antora-based documentation build

zregvart:
This adds Antora-based documentation build. This relies on YAML files
generated by `internal/documentation` and an Antora extension in
`docs/extension` to generate content from those YAML files at runtime
using Handlebars template in `docs/reference.hbs`.

PR #137, opened 2022-08-05
Set a 5% drop threshold for code coverage

zregvart:
This should help with getting the Codecov checks report as success a bit
more.


zregvart:
Helped on #136, merging

PR #142, opened 2022-08-08
Replace golangci GitHub Action with Make target

zregvart:
This keeps us on the same version of golangci as used when running via
make (i.e. via go run).

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -35,6 +35,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - name: Setup Go environment
         uses: actions/setup-go@v3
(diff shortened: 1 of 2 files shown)


internal/evaluator/conftest_evaluator.go
```diff
@@ -61,10 +61,7 @@ func NewConftestEvaluator(ctx context.Context, policySources []source.PolicySour
 	}
 
 	dir, err := c.createWorkDir()
```
robnester-rh, line 63: is this `err` no longer being handled?

zregvart: It's part of [Test remove me](https://github.com/hacbs-contract/ec-cli/pull/142/commits/578bdba55d7d46ac3eff378a59102e921dd8f231) commit to test the linter on GitHub Actions

robnester-rh: > It's part of [Test remove me](https://github.com/hacbs-contract/ec-cli/pull/142/commits/578bdba55d7d46ac3eff378a59102e921dd8f231) commit to test the linter on GitHub Actions

ack. Gotcha!

lcarva (approved):

PR #154, opened 2022-08-11
Fix ec-replace diff generation

lcarva:
Use os.exec instead of go-git to generate the diff. This is not ideal,
but there are two issues with using go-git for this use case.

First, it is currently not possible to produce a diff for uncommitted
changes. Discussion: https://github.com/src-d/go-git/issues/604

Second, the output generated by go-git is just plain wrong and in some
cases does not work with "git apply".


robnester-rh (approved):
LGTM

PR #155, opened 2022-08-11
Read RekorUrl and PublicKey from ECP

lcarva:
https://issues.redhat.com/browse/HACBS-767


internal/policy/policy.go
```diff
@@ -0,0 +1,108 @@
...
+
+	ecc "github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
+	"github.com/sigstore/cosign/cmd/cosign/cli/rekor"
+	"github.com/sigstore/cosign/pkg/cosign"
+	cosignSig "github.com/sigstore/cosign/pkg/signature"
+	sigstoreSig "github.com/sigstore/sigstore/pkg/signature"
+	log "github.com/sirupsen/logrus"
+)
+
+// WithPublicKey returns a new instance of EnterpriseContractPolicy if an update
```
joejstuart, line 33: typo. `WithRekorUrl`

lcarva: Guilty of copy and paste! Fixed :)

internal/kubernetes/client.go
```diff
@@ -66,20 +65,37 @@ func createControllerRuntimeClient() (client.Client, error) {
...
-	if err != nil {
-		log.Debug("Failed to get policy from cluster!")
+	if err := k.client.Get(ctx, *name, policy); err != nil {
+		log.Debugf("Failed to fetch the policy from cluster: %s", err)
 		return nil, err
 	}
-	policyJson, _ := json.Marshal(policy.Spec)
-	log.Debugf("Policy fetched:\n%s", policyJson)
+	if policyJson, err := json.Marshal(policy.Spec); err != nil {
+		log.Debugf("Unable to marshal policy to json: %s", policyJson)
```
joejstuart, line 94: Do you need to return the err after logging?

internal/policy/policy.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/sigstore/cosign/cmd/cosign/cli/rekor"
+	"github.com/sigstore/cosign/pkg/cosign"
+	cosignSig "github.com/sigstore/cosign/pkg/signature"
+	sigstoreSig "github.com/sigstore/sigstore/pkg/signature"
+	log "github.com/sirupsen/logrus"
+)
+
+// WithRekorUrl returns a new instance of EnterpriseContractPolicy if an update
+// to its rekorUrl is needed. Otherwise, the same instance is returned.
+func WithRekorUrl(policy *ecc.EnterpriseContractPolicy, rekorUrl string) *ecc.EnterpriseContractPolicy {
```
joejstuart, line 35: What if the rekorUrl is passed on the cli and it's in the policy? Is the user notified the cli param is ignored? Or should the policy be overridden?

internal/kubernetes/client.go
lcarva: Well I don't know... The call to `json.Marshall` is only meant to assist in logging (which seems a bit odd to me). In the previous code, the error code was ignored completely. My assumption is that it was intended to be an optional thing. I can't picture a scenario where the `json.Marshall` call actually fails though. I can switch back to ignoring the error as it was before?

internal/policy/policy.go
lcarva: I had a [discussion|https://issues.redhat.com/browse/HACBS-767?focusedCommentId=20778298&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20778298] with myself about the precedence of these values. I reasoned that parameters from the CLI should overwrite whatever is in the ECP.

In `cmd/validate_image.go`, I changed the description of the rekorUrl and publicKey parameters to indicate this behavior. Does that address your concern?

internal/kubernetes/client.go
joejstuart: you're right, I didn't look close enough. Maybe log `err` though instead of `policyJson`?

lcarva: I took it a step further. I removed the `json.Marshall` call altogether. Instead, I just log the policy object directly. I'm using the `%#v` "verb" now so it actually shows up in a more readable way. Wdyt?

internal/policy/policy.go
joejstuart: perfect!

joejstuart:
LGTM

PR #179, opened 2022-09-07
HACBS-1048 Check if image exists before proceeding

robnester-rh:
This commit adds a ValidateImageAccess method on the ApplicationSnapshotImage type which verifies that the given image ref is accessible. Failures are added to the output.

### DIFF
--- a/cmd/validate_image_test.go
+++ b/cmd/validate_image_test.go
@@ -152,6 +152,9 @@ func Test_ValidateImageCommand(t *testing.T) {
 			ImageSignatureCheck: output.VerificationStatus{
 				Passed: true,
 			},
+			ImageAccessibleCheck: output.VerificationStatus{
+				Passed: true,
+			},
 			AttestationSignatureCheck: output.VerificationStatus{
 				Passed: true,
 			},
@@ -221,12 +224,65 @@ func Test_ValidateErrorCommand(t *testing.T) {
 	assert.Empty(t, out.String())
 }
 
+func Test_FailureImageAccessibility(t *testing.T) {
+	validate := func(ctx context.Context, imageRef, policyConfiguration, publicKey, rekorURL string) (*output.Output, error) {
+		return &output.Output{
+			ImageSignatureCheck: output.VerificationStatus{
+				Passed: false,
+				Result: &conftestOutput.Result{Message: "skipped due to inaccessible image ref"},
+			},
+			ImageAccessibleCheck: output.VerificationStatus{
+				Passed: false,
+				Result: &conftestOutput.Result{Message: "image ref not accessible. HEAD registry/image:tag: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details)"},
+			},
+			AttestationSignatureCheck: output.VerificationStatus{
+				Passed: false,
+				Result: &conftestOutput.Result{Message: "skipped due to inaccessible image ref"},
+			},
+		}, nil
+	}
+
+	cmd := validateImageCmd(validate)
+
+	cmd.SetArgs([]string{
+		"--image",
+		"registry/image:tag",
+		"--public-key",
+		"test-public-key",
+	})
+
+	var out bytes.Buffer
+	cmd.SetOut(&out)
+
+	err := cmd.Execute()
+	assert.NoError(t, err)
+	assert.JSONEq(t, `{
+		"success": false,
+		"components": [
+		  {
+			"name": "Unnamed",
+			"containerImage": "registry/image:tag",
+			"violations": [
+			  {"msg": "skipped due to inaccessible image ref"},
+			  {"msg": "image ref not accessible. HEAD registry/image:tag: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details)"},
+			  {"msg": "skipped due to inaccessible image ref"}
+			],
+			"warnings": [],
+			"success": false
+		  }
+		]
+	  }`, out.String())
+}
+
 func Test_FailureOutput(t *testing.T) {
 	validate := func(ctx context.Context, imageRef, policyConfiguration, publicKey, rekorURL string) (*output.Output, error) {
 		return &output.Output{
 			ImageSignatureCheck: output.VerificationStatus{
 				Passed: false,
 				Result: &conftestOutput.Result{Message: "failed image signature check"},
+			},
+			ImageAccessibleCheck: output.VerificationStatus{
+				Passed: true,
 			},
 			AttestationSignatureCheck: output.VerificationStatus{
 				Passed: false,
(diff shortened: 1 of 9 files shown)


internal/utils/helpers.go
```diff
@@ -57,3 +60,24 @@ func hasPrefix(buf []byte, prefix []byte) bool {
...
+
+// CheckImageRefAccess verifies if the provided imageRef is accessible
+func CheckImageRefAccess(imageRef string) (ok bool, err error) {
+	ref, _ := name.ParseReference(imageRef)
+	logrus.Debugf("Parsed ref: %s", ref)
+	resp, err := remote.Head(ref, ImageRefTransport)
+	if err != nil {
+		return
+	}
+	if resp == nil {
```
lcarva, line 77: I wonder if creating, and returning, a new error here would make the API a little bit simpler (e.g. callers just have to worry about checking err, not both err and ok).

internal/image/validate.go
```diff
@@ -36,14 +37,33 @@ func ValidateImage(ctx context.Context, imageRef, policyConfiguration, publicKey
 		return nil, err
 	}
 
+	if err = a.ValidateImageAccess(ctx); err != nil {
+		log.Debugf("Image access check failed. Error: %s", err.Error())
+		out.SetImageAccessibleCheck(false, fmt.Sprintf("image ref not accessible. %s", err))
+		out.SetImageSignatureCheck(false, "skipped due to inaccessible image ref")
+		out.SetAttestationSignatureCheck(false, "skipped due to inaccessible image ref")
+		//attRes := []conftestOutput.CheckResult{
```
lcarva, line 45: Remove?

internal/mocks/mocks.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package mocks
+
+import "net/http"
+
+type HttpTransportMockSuccess struct {
+}
+type HttpTransportMockFailure struct {
```
lcarva, line 23: This one seems to not be used.

internal/utils/helpers.go
```diff
@@ -57,3 +58,8 @@ func hasPrefix(buf []byte, prefix []byte) bool {
 func CreateWorkDir() (string, error) {
 	return CreateTmpDir(AppFS, afero.GetTempDir(AppFS, ""), "ec-work-")
 }
+
+// ImageRefTransport is used to inject the type of transport to use with the
+// remote.WithTransport function. By default, remote.DefaultTransport is
+// equivalent to http.DefaultTransport, with a reduced timeout and keep-alive
+var ImageRefTransport = remote.WithTransport(remote.DefaultTransport)
```
lcarva, line 65: I think it's a bit dangerous to use a public global variable like this. Can we make it private within the `application_snapshot_image` module?

internal/image/validate.go
```diff
@@ -36,14 +37,33 @@ func ValidateImage(ctx context.Context, imageRef, policyConfiguration, publicKey
 		return nil, err
 	}
 
+	if err = a.ValidateImageAccess(ctx); err != nil {
+		log.Debugf("Image access check failed. Error: %s", err.Error())
+		out.SetImageAccessibleCheck(false, fmt.Sprintf("image ref not accessible. %s", err))
+		out.SetImageSignatureCheck(false, "skipped due to inaccessible image ref")
```
lcarva, line 43: Could it be that adding the "skipped" messages under violations may cause confusion?

cmd/validate_image_test.go
```diff
@@ -221,13 +228,66 @@ func Test_ValidateErrorCommand(t *testing.T) {
...
+func Test_FailureImageAccessibility(t *testing.T) {
+	validate := func(ctx context.Context, imageRef, policyConfiguration, publicKey, rekorURL string) (*output.Output, error) {
+		return &output.Output{
+			ImageSignatureCheck: output.VerificationStatus{
+				Passed: false,
+				Result: &conftestOutput.Result{Message: "skipped due to inaccessible image ref"},
+			},
+			ImageAccessibleCheck: output.VerificationStatus{
+				Passed: false,
+				Result: &conftestOutput.Result{Message: "image ref not accessible. HEAD registry/image:tag: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details)"},
```
lcarva, line 240: Wait, is this actually making an HTTP request to the host `registry`?

robnester-rh: No, it's just just validating output if it were to have those results.

internal/mocks/mocks.go
robnester-rh: Added some tests, it's used.

internal/utils/helpers.go
robnester-rh: ack

internal/image/validate.go
robnester-rh: Would you rather them be left empty?

internal/image/validate.go
robnester-rh: ack

internal/image/validate.go
lcarva: Not sure. I think arguably, "skipped" is not really a violation, but I see the motivation for adding it here. Maybe a debug message is a better approach? Either way is fine. We can always tweak later after some experience over time.

cmd/validate_image_test.go
lcarva: Doh. Of course, the real validate function is never called, that's what this mock is. Thanks for clarifying.

internal/image/validate.go
```diff
@@ -36,14 +37,24 @@ func ValidateImage(ctx context.Context, imageRef, policyConfiguration, publicKey
 		return nil, err
 	}
 
+	if err = a.ValidateImageAccess(ctx); err != nil {
+		log.Debugf("Image access check failed. Error: %s", err.Error())
+		out.SetImageAccessibleCheck(false, fmt.Sprintf("image ref not accessible. %s", err))
+		out.SetImageSignatureCheck(false, "")
```
lcarva, line 43: I don't think this is needed, right? It should default to `false`.

lcarva (approved):

internal/image/validate.go
robnester-rh: I've left them empty for now.

internal/image/validate.go
robnester-rh: If you're referring to the `false` argument provided to `out.SetImageSignatureCheck()`, it is required. While the default for a `boolean` is `false`, you can't have default parameter values in method signatures, so both parameters "passed" and "message" must be provided.

lcarva: Oh, that's not what I meant. I meant omitting the call to `out.SetImageSignatureCheck` altogether. The underlying default state for the `ImageSignatureCheck` should be false by default.

robnester-rh: Turns out that skipping any of the following results in a nil pointer dereference:
```
out.SetImageSignatureCheck(false, "")
out.SetAttestationSignatureCheck(false, "")
out.SetPolicyCheck(nil)
```

PR #184, opened 2022-09-19
HACBS-1089 Remove git+ from git URL

robnester-rh:
Chains seems to add the `git+` scheme to a git URL to conform with the SPDX format [1]. This causes issues with accessesing the URL later in the process.

[1] https://github.com/tektoncd/chains/commit/243e90b3736f2fa4daa013c38c1cefba254cf106#diff-43d1069a9ea72a3c7aaf57b35324bd9650f7f5f3359cf88fa05efaf6f5e8c2e0R297-R306

### DIFF
--- a/internal/image/attestation.go
+++ b/internal/image/attestation.go
@@ -18,6 +18,7 @@ package image
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/go-git/go-git/v5/plumbing/object"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


internal/image/attestation.go
```diff
@@ -81,6 +82,10 @@ func (a *attestation) getBuildSCM() string {
 	uri := "" //https://github.com/joejstuart/ec-cli.git"
 	if len(a.Predicate.Materials) == 1 {
 		uri = a.Predicate.Materials[0].Uri
+		// If our URI has 'git+https' as the scheme, we will strip the "git+" portion.
+		if strings.HasPrefix(uri, "git+") {
+			uri = strings.TrimLeft(uri, "git+")
+		}
```
zregvart, line 88: nit: this could be:
```suggestion
		// If our URI has 'git+https' as the scheme, we will strip the "git+" portion.
		uri = strings.TrimPrefix(a.Predicate.Materials[0].Uri, "git+")
```
(if we want one conditional less)

zregvart (approved):

lcarva (approved):

PR #188, opened 2022-09-21
Add support for inline ECP and use context for mocking

lcarva:
```
Remove unnecessary default values
```

```
Support inline policy in "validate image"
```

```
Experiment with context to mock external call

Based on the pattern defined in:
https://incident.io/blog/golang-client-mocks

The pattern is only currently used in a couple of places as a mechanism
to illustrate the pattern.

Mocks are not auto generated at this time, but it's something to
consider if this pattern is adopted throughout the code base.
```

There are other placed in ec-cli that read the ECP from kubernetes that can also be updated. I purposely kept it out of this PR as to keep it to a reasonable size.


cmd/validate_image.go
```diff
@@ -23,15 +23,17 @@ import (
...
 	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
 	"github.com/spf13/cobra"
 
 	"github.com/hacbs-contract/ec-cli/internal/applicationsnapshot"
 	"github.com/hacbs-contract/ec-cli/internal/output"
+	"github.com/hacbs-contract/ec-cli/internal/policy"
 )
 
-type imageValidationFunc func(ctx context.Context, imageRef, policyConfiguration, publicKey, rekorURL string) (*output.Output, error)
+type imageValidationFunc func(ctx context.Context, imageRef string, policy *ecc.EnterpriseContractPolicySpec) (*output.Output, error)
```
zregvart, line 36: For further refactor (another PR), we could instead of `ecc.EnterpriseContractPolicySpec` have a ec-cli or ec-policies configuration object here and map the `ecc.EnterpriseContractPolicySpec` to it instead, to have a intermediate common layer so to say...

cmd/validate_image.go
```diff
@@ -46,8 +46,6 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		spec                *appstudioshared.ApplicationSnapshotSpec
 	}{
 		policyConfiguration: "ec-policy",
-		rekorURL:            "",
-		strict:              false,
```
zregvart, line 50: :+1:

cmd/validate_image.go
```diff
@@ -101,15 +104,28 @@ Use an EnterpriseContractPolicy resource from the currently active kubernetes co
...
-  ec validate image --image registry/name:tag --policy my-namespace/my-policy`,
-		PreRunE: func(cmd *cobra.Command, args []string) error {
-			s, err := applicationsnapshot.DetermineInputSpec(data.filePath, data.input, data.imageRef)
-			if err != nil {
-				return err
+  ec validate image --image registry/name:tag --policy my-namespace/my-policy
+
+Use an inline EnterpriseContractPolicy spec
+  ec validate image --image registry/name:tag --policy '{"publicKey": "<path/to/public/key>"}'`,
+		PreRunE: func(cmd *cobra.Command, args []string) (allErrors error) {
```
zregvart, line 111: I _think_ I'm wrong on this, but just to check if `applicationsnapshot.DetermineInputSpec()` fails and `policy.NewPolicy` fails, do we still want to proceed?
Also if one of these fails but the other succeeds are we okay with returning the error?

internal/policy/policy.go
```diff
@@ -28,36 +30,72 @@ import (
...
+// EnterpriseContractPolicySpec.
+func NewPolicy(ctx context.Context, policyRef, rekorUrl, publicKey string) (*ecc.EnterpriseContractPolicySpec, error) {
+	var policy *ecc.EnterpriseContractPolicySpec
+
+	if policyRef == "" {
+		log.Debug("Using an empty EnterpriseContractPolicy")
+		// Default to an empty policy instead of returning an error because the required
+		// values, e.g. PublicKey, may be provided via other means, e.g. publicKey param.
+		policy = &ecc.EnterpriseContractPolicySpec{}
+	} else if strings.Contains(policyRef, "{") {
```
zregvart, line 55: Nitpick: trim leading whitespace in `policyRef`?

internal/image/authorization.go
```diff
@@ -121,14 +121,14 @@ func (k *K8sSource) GetSource() (authorizationGetter, error) {
...
-	k8s, err := kubernetesClientCreator()
+func fetchECSource(ctx context.Context, namedResource string) (*ecp.EnterpriseContractPolicy, error) {
+	k8s, err := kubernetesClientCreator(ctx)
 	if err != nil {
 		log.Debug("Failed to initialize Kubernetes client")
 		return nil, err
 	}
 
-	ecp, err := k8s.FetchEnterpriseContractPolicy(context.TODO(), namedResource)
+	ecp, err := k8s.FetchEnterpriseContractPolicy(ctx, namedResource)
```
zregvart, line 131: :+1:

zregvart (approved):
Very cool!

zregvart:
One more thing™️, do we want to version `acceptance/.sigstore`, perhaps we should put it in .gitignore...

robnester-rh (approved):
Nitpick: Use consistent import alias for `github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1`. I think it makes sense to replace `ecp` with `ecc` as you did in a few locations in this change.

internal/policy/policy.go
lcarva: This is using `Contains`, not `HasPrefix`, so it should work regardless. I picked the `{` character because it's likely the first one to be found (and it's also not a valid character for a resource name).

zregvart: Yeah you're right, I read that as `strings.HasPrefix`. `strings.Contains` should be good enough

cmd/validate_image.go
lcarva: The behavior I was aiming for was: if any, or both, checks fail then it's a fatal error. Added a unit test to confirm this behavior.

cmd/validate_image.go
lcarva: :+1:  I like this idea. Let's do that on a follow up PR.

lcarva:
> One more thingtm, do we want to version `acceptance/.sigstore`, perhaps we should put it in .gitignore...

@zregvart. I have not idea :smile: Maybe we can remove it on a follow up PR?

lcarva:
> Nitpick: Use consistent import alias for `github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1`. I think it makes sense to replace `ecp` with `ecc` as you did in a few locations in this change.

@robnester-rh, added a commit for this. Thanks.

PR #192, opened 2022-09-22
Rework how rego is fetched, add 'fetch policy' command

simonbaird:
Instead of using the existing git clone code in
internal/policy/source/repo try using conftests Download function instead.

### DIFF
--- a/cmd/fetch_policy.go
+++ b/cmd/fetch_policy.go
@@ -1,0 +1,114 @@
+// Copyright 2022 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Define the `ec fetch policy` command
+package cmd
+
+import (
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/hacbs-contract/ec-cli/internal/policy/source"
+	"github.com/hacbs-contract/ec-cli/internal/utils"
+)
+
+func fetchPolicyCmd() *cobra.Command {
+	var (
+		sourceUrls []string
+		destDir    string
+		useWorkDir bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "policy --source <source-url>",
+		Short: "Fetch policy rules from a git repository or other source",
+		Long: `Fetch policy rules (rego files) from a git repository or other source.
+
+Each policy source will be downloaded into a separate unique directory inside
+the "policy" directory under the destination directory specified. The
+destination directory is either an automatically generated temporary work dir
+if --work-dir is set, the directory specified with the --dest flag, or the
+current directory if neither flag is specified.
+
+This command is based on 'conftest pull' so you can refer to the conftest pull
+documentation for more usage examples and for details on the different types of
+supported source URLs.
+
+Note that this command is not typically required to verify the Enterprise
+Contract. It has been made available for troubleshooting and debugging
+purposes.`,
+
+		Example: `Fetching policies from multiple sources to a specific directory:
+
+  ec fetch policy --dest fetched-policies \
+    --source github.com/hacbs-contract/ec-policies//policy/lib \
+    --source github.com/hacbs-contract/ec-policies//policy/release
+
+Fetching policies from multiple sources to an automatically generated temporary
+work directory:
+
+  ec fetch policy --work-dir \
+    --source github.com/hacbs-contract/ec-policies//policy/lib \
+    --source github.com/hacbs-contract/ec-policies//policy/release
+
+Different style url formats are supported. In this example "policy" is treated as
+a subdirectory even without the go-getter style // delimiter:
+
+  ec fetch policy --source https://github.com/hacbs-contract/ec-policies/policy
+
+Notes:
+
+- The --dest flag will be ignored if --work-dir is set
+- Adding a protocol prefix such as 'git::' to the source url forces it to be treated
+  as a go-getter style url.`,
+
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if useWorkDir {
+				workDir, err := utils.CreateWorkDir()
+				if err != nil {
+					log.Debug("Failed to create work dir!")
+					return err
+				}
+				destDir = workDir
+			}
+
+			for _, s := range sourceUrls {
+				// Do everything the same way that it would be done when an image validation happens
+				policyRepo, err := source.CreatePolicyRepoFromRepoAndRevision(s, nil)
+				if err != nil {
+					return err
+				}
+				err = policyRepo.GetPolicies(cmd.Context(), destDir, true)
+				if err != nil {
+					return err
+				}
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringArrayVarP(&sourceUrls, "source", "s", []string{}, "source url. multiple values are allowed")
+	cmd.Flags().StringVarP(&destDir, "dest", "d", ".", "use the specified download destination directory. ignored if --work-dir is set")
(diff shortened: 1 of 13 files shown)


internal/downloader/downloader.go
```diff
@@ -0,0 +1,46 @@
...
+// sources that we might need in future, so let's use it rather than do our
+// own fetching.
+//
+//
+package downloader
+
+import (
+	"context"
+
+	conftestDownloader "github.com/open-policy-agent/conftest/downloader"
```
zregvart, line 32: Nitpick, we usually don't see package names this long, so perhaps
```suggestion
	d "github.com/open-policy-agent/conftest/downloader"
```

or

```suggestion
	conftest "github.com/open-policy-agent/conftest/downloader"
```

simonbaird: Yes, we must avoid readable code at all costs when using golang. :stuck_out_tongue_winking_eye:

internal/policy/source/repo.go
```diff
@@ -56,6 +56,7 @@ func (p *PolicyRepo) GetPolicyDir() string {
 func (p *PolicyRepo) GetPolicies(ctx context.Context, dest string) error {
 	// Checkout policy repo into work directory.
 	log.Debugf("Checking out repo %s at %s to work dir", p.RepoURL, p.RepoRef)
+	// Todo: Consider if we can use downloader.Download here instead
```
simonbaird, line 59: I'm interested in the idea of doing all our downloads with conftest downloader, but not sure I'm ready to take on the patch.

internal/downloader/downloader.go
zregvart: image

zregvart:
You also need to run `make docs`

internal/policy/source/repo.go
lcarva: +1 I think it would be a nice idea to explore.

cmd/fetch_policy.go
```diff
@@ -0,0 +1,80 @@
...
+		Short: "Fetch policy rules from a git repository or other source",
+		Long: `Fetch policy rules (rego files) from a git repository or other source.
+
+This command is actually a thin wrapper for 'conftest pull' so you can refer to the
+conftest pull documentation for more details on the different types of supported source
+URLs.
+
+This is not typically required to verify the Enterprise Contract, but it's exposed for
+troubleshooting and debugging purposes.
+`,
```
lcarva, line 55: nit: cobra will add a new line, so this will end up looking like an extra line. Consider removing it.

cmd/fetch_policy.go
```diff
@@ -0,0 +1,80 @@
...
+			if useWorkDir {
+				workDir, err := utils.CreateWorkDir()
+				if err != nil {
+					log.Debug("Failed to create work dir!")
+					return err
+				}
+				fullDestDir = filepath.Join(workDir, destDir)
+			}
+
+			return downloader.Download(cmd.Context(), fullDestDir, sourceUrl)
```
lcarva, line 73: Maybe we need to write to stdout to tell the user where exactly it was downloaded, specially if `useWorkDir` is `true`.

lcarva (approved):

cmd/fetch_policy.go
simonbaird: I was thinking about that, and I considered doing a log.Info(...) here, but I decided against it since you can do `--verbose` and get a similar message from the internal/downloader.

simonbaird: On the other hand, you'd probably always want to see the random tmp work dir...

lcarva: Yeah, I think stdout is probably a good place for that info.

simonbaird: As a lazy solution, I mentioned in the docs that you should use --verbose to see the work dir.

simonbaird:
Recent fixes:
* Git add docs/reference/ec_fetch_policy.yaml which was missing
* Docs improvements, e.g. make the one example actually work
* Suggest in the help to use --verbose with --use-work-dir

zregvart:
I think we need @robnester-rh's thoughts on this. For consistency we'd like to use this instead of the git clone we do now.

zregvart:
_Whispers in the dark: tests would be nice_

internal/downloader/downloader.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"context"
+
+	ctdl "github.com/open-policy-agent/conftest/downloader"
+	log "github.com/sirupsen/logrus"
+)
+
+// Let's do just one url at a time even though ctdl.Download can take a list of urls
+//
```
robnester-rh, line 37: Nit: Exported functions should have a proper doc string
```
// Function Name does blah blah blah
```

cmd/fetch_policy.go
```diff
@@ -0,0 +1,83 @@
...
+//
+// Define the `ec fetch policy <source-url>` command.
+//
+// As it says in the docs it's not generally needed for doing EC checks, but
+// it can be useful generally for hacking and troubleshooting.
+//
+// If/when we designate some commands as "experimental" or "plumbing" then this
+// could be one of them.
+//
+//
```
robnester-rh, line 27: Nit: remove comment markers from blank lines

simonbaird:
> _Whispers in the dark: tests would be nice_

The force is not yet strong enough.

internal/downloader/downloader.go
```diff
@@ -0,0 +1,45 @@
...
+//////////////////////////////////////////////////////////////////////////////
+//
+// Provide a thin-as-possible wrapper for conftest's `Download` function,
+// which is itself mostly a wrapper for hashicorp/go-getter.
+//
+// Conftest's download function already has support for a lot of different
+// types of sources that we might want to use in future, so let's aim to use
+// it rather than writing our own code for fetching rego rules.
+//
+//
```
robnester-rh, line 26: Nit: remove comment markers from empty lines.

cmd/fetch_policy.go
robnester-rh: I think putting it to stdout would be preferable as well.

robnester-rh (approved):
a few notes provided. I'm fine if this goes in as is, however.

simonbaird:
The existing directory thing might be a problem since we likely would want to be able to download rego files from multiple sources into the default location, `./policy`. Let me see what happens in that scenario.

Re tests, actually I could take a look at it.

simonbaird:
Back to draft...

simonbaird:
So actually it *does* trash anything in the destination dir.

Observe:

```
localhost:~/code/ec-cli> [fetch-policy-cmd] go run main.go fetch policy github.com/hacbs-contract/ec-policies//policy/release --dest-dir policy && ls policy | head -3
Downloading from source url 'github.com/hacbs-contract/ec-policies//policy/release' to local dir 'policy'
attestation_task_bundle.rego
attestation_task_bundle_test.rego
attestation_type.rego
localhost:~/code/ec-cli> [fetch-policy-cmd] go run main.go fetch policy github.com/hacbs-contract/ec-policies//policy/pipeline --dest-dir policy && ls policy | head -3
Downloading from source url 'github.com/hacbs-contract/ec-policies//policy/pipeline' to local dir 'policy'
basic.rego
basic_test.rego
main.rego
```
Not sure if it's a feature or a bug, but we need to consider it I guess.

simonbaird:
Anyway, I added a test.

zregvart:
> Not sure if it's a feature or a bug, but we need to consider it I guess.

TBH, I much preferred the the go-getter behavior where it would error out when an attempt to overwrite files was detected. Considering multiple policy sources, we should place them in different directories to prevent overwriting.

cmd/fetch_policy.go
simonbaird: Done in the current revision.

simonbaird:
I'm planning to do it like this:

Policy source 1 - downloaded to $workdir/policy/$randomstring/
Policy source N - downloaded to $workdir/policy/$randomstring/

The random string is based on the source and a timestamp so there's no chance of a name clash. Since conftest will traverse the tree, it should not matter what the directory names are.

cmd/fetch_policy.go
```diff
@@ -0,0 +1,114 @@
...
+supported source URLs.
+
+Note that this command is not typically required to verify the Enterprise
+Contract. It has been made available for troubleshooting and debugging
+purposes.`,
+
+		Example: `Fetching policies from multiple sources to a specific directory:
+
+  ec fetch policy --dest fetched-policies \
+    --source github.com/hacbs-contract/ec-policies//policy/lib \
```
lcarva, line 57: Out of curiosity, how do you specify a branch?

lcarva (approved):

cmd/fetch_policy.go
simonbaird: You can add params, e.g. `?ref=main`

zregvart (approved):

PR #195, opened 2022-09-23
Mega error linter

zregvart:
[Move error to pkg/error](https://github.com/hacbs-contract/ec-cli/commit/ca3b683245ba896d2f644b0c3bd4730555f054ec)

We need to use this from `cmd` and `internal` packages.

[Error conventions linter & fixer](https://github.com/hacbs-contract/ec-cli/commit/332045d8874dd91f8551c776e508c4bade640d9c)

Adds a linter/fixer for the convention of not returning raw errors. Raw errors are errors that haven't been wrapped by our `pkg/error`, i.e. they don't have an error code and our custom message.


zregvart:
Let's first see how folk feel about this before I go ahead and introduce errors to the existing code (that's why the lint fails here).

lcarva:
@zregvart , lots of linter errors reported. Expected?

zregvart:
> @zregvart , lots of linter errors reported. Expected?

Yup, its the new linter working, `make lint-fix` might fix them, but it'll be a bit to get all `TODO`s resolved left by the fixes...

cmd/validate_image.go
```diff
@@ -199,7 +202,7 @@ Use an inline EnterpriseContractPolicy spec
 
 			if data.strict && !success {
 				// TODO: replace this with proper message and exit code 1.
-				return errors.New("success criteria not met")
+				return CM024.CausedBy(errors.New("success criteria not met"))
```
lcarva, line 205: Maybe as a future feature, we can do something like: `CM024.New("success criteria not met")`

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -38,6 +38,26 @@ import (
 	"github.com/hacbs-contract/ec-cli/internal/evaluator"
 	"github.com/hacbs-contract/ec-cli/internal/policy"
 	"github.com/hacbs-contract/ec-cli/internal/policy/source"
+	e "github.com/hacbs-contract/ec-cli/pkg/error"
+)
+
+var (
+	AP001 = e.NewError("AP001", "", e.ErrorExitStatus) // TODO: add message and set the exit status
```
lcarva, line 45: It looks like error codes are not unique across the application. How would we know if `AP001` came from an error in this file or in `internal/applicationsnapshot/input.go`? Well.. I should say: If a user reports seeing `AP001`, how do we map it to the right place in the code?

internal/kubernetes/client.go
```diff
@@ -20,13 +20,26 @@ import (
 	"context"
 	"errors"
 
+	e "github.com/hacbs-contract/ec-cli/pkg/error"
```
lcarva, line 23: Wrong import order?

internal/lint/error.go
```diff
@@ -0,0 +1,558 @@
...
+		if spec, ok = s.(*ast.ValueSpec); !ok {
+			continue
+		}
+
+		if len(spec.Values) != 1 {
+			continue
+		}
+
+		val := spec.Values[0]
+		if len(spec.Values) != 1 {
```
lcarva, line 196: We do this check above. Remove it?

internal/lint/error.go
```diff
@@ -0,0 +1,558 @@
...
+
+		if spec, ok = s.(*ast.ValueSpec); !ok {
+			continue
+		}
+
+		if len(spec.Values) != 1 {
+			continue
+		}
+
+		val := spec.Values[0]
```
lcarva, line 195: nit: I think not having an empty line before this statement would improve the readability of the code, as it would group it with the relevant if-block.

internal/lint/error.go
```diff
@@ -0,0 +1,558 @@
...
+
+	names := make([]string, 0, 5)
+	for _, s := range g.Specs {
+		var spec *ast.ValueSpec
+
+		if spec, ok = s.(*ast.ValueSpec); !ok {
+			continue
+		}
+
+		if len(spec.Values) != 1 {
```
lcarva, line 191: Could it be more than 1? What happens in that case?

internal/ecgit/ecgit.go
```diff
@@ -35,6 +35,39 @@ import (
...
+	EC020 = e.NewError("EC020", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC021 = e.NewError("EC021", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC022 = e.NewError("EC022", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC023 = e.NewError("EC023", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC024 = e.NewError("EC024", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC025 = e.NewError("EC025", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC026 = e.NewError("EC026", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC027 = e.NewError("EC027", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC028 = e.NewError("EC028", "", e.ErrorExitStatus) // TODO: add message and set the exit status
+	EC029 = e.NewError("EC029", "", e.ErrorExitStatus) // TODO: add message and set the exit status
```
simonbaird, line 70: So ugly though. Could we maybe generate a code from file and line number?

simonbaird: Is it going to be tedious to main a message that is far away from the place where it's used?

internal/lint/ec.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/multichecker"
+)
+
+type ecPlugin struct{}
+
+// GetAnalyzers returns the private linters implemented in this package. This is
+// the interfice required for custom golangci-linters, but due to an issue with
```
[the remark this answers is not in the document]
zregvart: I guess changing "This" to "That" would help

internal/lint/ec.go
```diff
@@ -0,0 +1,52 @@
...
+// applying the suggestions (fixes) we can't use it in that capacity. It remains
+// here as we still can use it from the main function below.
+func (*ecPlugin) GetAnalyzers() []*analysis.Analyzer {
+	return []*analysis.Analyzer{
+		&current,
```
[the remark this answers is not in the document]
zregvart: They're vars in [internal/lint/error.go](https://github.com/hacbs-contract/ec-cli/pull/195/files#diff-b14f039fc7fe7371ba40b200701fb23dfa73a14dbdbc1f2ddf13ef7a16f8f653)

cmd/validate_image.go
zregvart: Totally thought of that but I did not want to have this bigger than it already is, also for the `fmt.Errorf` variant.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
zregvart: Hard for the linter to know about that, linters work per-package and cross-package with much more effort. Though existing error prefixes would be picked up, so if we take care to have them unique on a per-package basis (something review would need to catch) we should be okay. Also generating documentation from this, listing all error codes should not be that difficult, and that could ease the review...

internal/kubernetes/client.go
zregvart: Yeah, I didn't _gci_ this, I was thinking of adding gci call to the linter, perhaps I'll do that on this PR. The workaround would be running `make lint-fix` twice.

internal/lint/error.go
zregvart: Decided not to support cases when initializing two variables in the same declaration, e.g.:

```
var a, b = 1, 2
```

internal/ecgit/ecgit.go
zregvart: > So ugly though. Could we maybe generate a code from file and line number?

The linter has no notion of aesthetics, but a person could take this and refactor as they see it. I do see errors declared at the top of the file in other folk's golang code, so I guess it could be a convention folk use. Folk also tend to optimize the allocation by pre-allocating the errors this way (not sure if we'd care at all).

I don't think file/line numbers for error codes are not a good idea as they would be difficult with different versions and code changes to link up. Those could be in the message or `pkg/error.Error` could be extended to include those separately.

Linter finds all occurrences where we are returning a raw error and generates a separate `pkg/error.Error` code for each of those, we don't need to have that many errors, might not be an issue with 3-5 errors per file. Though I would argue for having many errors, e.g. say we have the same error code for "access denied" error and reused it in three places in a package for accessing three different resources, it might be difficult to reason about accessing what resource causes the error.

There is a pattern in ODBC/SQL errors where each error code has a reason code associated with it as well, though I think that would make it a bit more difficult to maintain.

simonbaird: Ack, thanks. I have no objections, just thinking it through.

simonbaird (approved):

zregvart:
The linter now doesn't fail on GitHub actions, but reports via GitHub [warning annotations](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions). The `make lint-fix` doesn't apply suggestions from the error linter.

PR #197, opened 2022-09-26
HACBS-1089 Report errors if no predicate.materials

robnester-rh:
This commit adds functionality which keeps track of how many components return an error from the validateGitSource function, which indicates components that have no, or blank, .predicate.materials sections. If the length of these components which return an error are equal to the length of the components, we return an error indicating that no valid .predicate.materials were found.

### DIFF
--- a/cmd/fetch_commit.go
+++ b/cmd/fetch_commit.go
@@ -18,9 +18,12 @@ package cmd
 
 import (
 	"context"
-	"log"
+	"errors"
+	"fmt"
 
+	"github.com/hashicorp/go-multierror"
 	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/hacbs-contract/ec-cli/internal/applicationsnapshot"
@@ -79,14 +82,15 @@ Use public key from a kubernetes secret:
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			var errs error
 			for _, comp := range data.spec.Components {
 				err := validateGitSource(cmd.Context(), comp.ContainerImage, data.publicKey)
 				if err != nil {
-					log.Println(err)
+					errs = multierror.Append(errs, fmt.Errorf("component: %+v - %w", comp, err))
 					continue
 				}
 			}
-			return nil
+			return errs
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/fetch_commit.go
```diff
@@ -79,13 +80,18 @@ Use public key from a kubernetes secret:
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			invalidPredicateMaterialCount := 0
 			for _, comp := range data.spec.Components {
 				err := validateGitSource(cmd.Context(), comp.ContainerImage, data.publicKey)
 				if err != nil {
 					log.Println(err)
+					invalidPredicateMaterialCount += 1
```
lcarva, line 88: This is counting if we got a git source for each component, which seem like a reasonable thing to do. However, this is not the problem outlined in [HACBS-1089](https://issues.redhat.com/browse/HACBS-1089). 

There could be, and there will always be in our use cases, multiple attestations per component/image. So for each image, we want to make sure that at least one of the attestations contains the git source information.

The check, I believe, should be added somewhere [here](https://github.com/hacbs-contract/ec-cli/blob/1162bdbd0e01426f84275faed47a0a11342e2cbb/cmd/fetch_commit.go#L107-L135).

Here's my reproducer:

```
🐚 cat cosign.pub 
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEe/UANElkGi2vnQsVr1rGOKfzmxkR
34RSOTNutukkeDmhFsHvI7BYEaii9wlWvsjDqc1Nt2IHxp9CKHQafRH4Vw==
-----END PUBLIC KEY-----

🐚 go run main.go fetch commit --image-ref quay.io/lucarval/tekton-test:latest --public-key cosign.pub
fetch_commit.go:87: there is no authorization source in attestation. sha: , url: 
Error: no valid .predicate.materials found in attestations
exit status 1
```

The image has two attestations, only one of which contains the git source information:
```
🐚 cosign verify-attestation --key cosign.pub quay.io/lucarval/tekton-test:latest 2> /dev/null | jq '.payload | @base64d | fromjson | .predicate.materials'
null
[
  {
    "uri": "git+https://github.com/lcarva/minimal-container.git",
    "digest": {
      "sha1": "10ab45329f9f29a2c84379b86f5b7e98a54f250e"
    }
  }
]
```

cmd/fetch_commit.go
```diff
@@ -109,7 +113,18 @@ func validateGitSource(ctx context.Context, imageRef, publicKey string) error {
 		return err
 	}
 
+	invalidPredicateMaterialCount := 0
 	for _, att := range validatedImage.Attestations {
+		if att.Predicate.Materials == nil {
```
lcarva, line 118: Although this achieves the job, I'm not sure if teaching `fetch_commit.go` about `.Predicate.Materials` is the best approach. What if we decide to look for git info differently in `NewGitSource` ? It won't be obvious that we also have to update here.

Can we just check the error from `att.NewGitSource()` a few lines below?


```go
gitSourceFound := false

for _, att := range validatedImage.Attestations {
  ...
  gitSource, err := att.NewGitSource()
  if err != nil {
    log.Debugf(...)
    continue
  }
  gitSourceFound = true
  ...
}

if !gitSourceFound {
  return errors.New(...)
}
return nil
```

zregvart (approved):

PR #198, opened 2022-09-26
Handle time-based rules

lcarva:
Introduces the concept of time-based rules to the ec-cli. See the corresponding ec-policies PR for context and nuances: https://github.com/hacbs-contract/ec-policies/pull/137


internal/image/validate.go
```diff
@@ -92,8 +93,42 @@ func ValidateImage(ctx context.Context, imageRef string, policy *ecc.EnterpriseC
 		return nil, err
 	}
 
+	now := time.Now()
+	// TODO: This should probably go somewhere else. Maybe in Evaluator.Evaluate?
```
zregvart, line 97: Yeah

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -41,7 +41,9 @@ import (
 )
 
 // ConftestNamespace is the default rego namespace using for policy evaluation
-const ConftestNamespace = "release.main"
+// TODO: This is actually not configurable.
+// const ConftestNamespace = "release.main"
```
zregvart, line 45: Let's not have dead code
```suggestion
```

internal/image/validate.go
```diff
@@ -92,8 +93,42 @@ func ValidateImage(ctx context.Context, imageRef string, policy *ecc.EnterpriseC
...
+	}
+
 	log.Debug("Conftest policy check complete")
 	out.SetPolicyCheck(results)
 
 	return out, nil
 }
+
+func isEffective(failure conftestOutput.Result, now time.Time) bool {
+	raw, ok := failure.Metadata["effective_on"]
```
zregvart, line 119: Perhaps also allow specifying metadata via command line parameter

zregvart (commented):
Simple enough...

internal/image/validate.go
lcarva: Good idea. I'll file a follow up story for this.

lcarva: https://issues.redhat.com/browse/HACBS-1219

internal/evaluator/conftest_evaluator.go
```diff
@@ -31,11 +32,38 @@ import (
 	"github.com/hacbs-contract/ec-cli/internal/utils"
 )
 
+type contextKey string
+
+const clientContextKey contextKey = "ec.evaluator.client"
+
+type client interface {
```
zregvart, line 39: Perhaps something like this?

```suggestion
type testRunner interface {
```

Perhaps for interfaces with a single function we don't need the interface but could do something like:

```suggestion
type testRunner func(context.Context, []string) ([]output.CheckResult, error)
```

PR #199, opened 2022-09-26
support list of authorizations

joejstuart:


zregvart:
Run `make docs` to update those yaml files, there is also a handy `ci` target for day-to-day development.

cmd/fetch_k8s_resource.go
```diff
@@ -46,90 +38,57 @@ func k8sResourceAuthorizationCmd() *cobra.Command {
...
-func validateK8sSource(ctx context.Context, imageRef, publicKey, namespace, server, resource string) error {
-	k8sSource, err := image.NewK8sSource(namespace, server, resource)
+func GetK8sAuthorization(ecp *ecc.EnterpriseContractPolicySpec) error {
+	resource, err := image.GetK8sResource(ecp)
 	if err != nil {
 		return err
 	}
 
-	authorization, err := image.GetAuthorization(ctx, k8sSource)
+	auth, err := resource.GetSignOff()
```
zregvart, line 75: We should probably handle the error here
```suggestion
	auth, err := resource.GetSignOff()
	if err != nil {
		return err
	}
```

cmd/fetch_k8s_resource.go
```diff
@@ -46,90 +38,57 @@ func k8sResourceAuthorizationCmd() *cobra.Command {
...
-	cmd.Flags().StringVarP(&data.resource, "resource", "r", data.resource, "name of the EnterpriseContractPolicy resource")
+	cmd.Flags().StringVarP(&data.policyConfiguration, "policy", "p", data.policyConfiguration,
+		"EntepriseContractPolicy reference [<namespace>/]<name>")
 
 	return cmd
 }
 
-func validateK8sSource(ctx context.Context, imageRef, publicKey, namespace, server, resource string) error {
-	k8sSource, err := image.NewK8sSource(namespace, server, resource)
+func GetK8sAuthorization(ecp *ecc.EnterpriseContractPolicySpec) error {
```
zregvart, line 69: Perhaps
```suggestion
func PrintK8sAuthorization(ecp *ecc.EnterpriseContractPolicySpec) error {
```
or
```suggestion
func LogK8sAuthorization(ecp *ecc.EnterpriseContractPolicySpec) error {
```

cmd/fetch_k8s_resource.go
```diff
@@ -20,23 +20,15 @@ import (
...
 func k8sResourceAuthorizationCmd() *cobra.Command {
 	var data = struct {
-		imageRef  string
-		publicKey string
-		filePath  string
-		input     string
-		namespace string
-		server    string
-		resource  string
-		spec      *appstudioshared.ApplicationSnapshotSpec
```
zregvart, line 39: Perhaps there is also value in having this provided individually? Either way is fine really...

cmd/fetch_k8s_resource.go
```diff
@@ -46,90 +38,57 @@ func k8sResourceAuthorizationCmd() *cobra.Command {
...
+	err = image.PrintAuthorization(auth)
 	if err != nil {
 		return err
 	}
 
-	imageValidator, err := image.NewImageValidator(ctx, imageRef, publicKey, "")
+	return nil
+}
+
+func validateK8sSource(ctx context.Context, policyConfiguration string) error {
```
zregvart, line 85: Perhaps?

```suggestion
func printK8sAuthorization(ctx context.Context, policyConfiguration string) error {
```

cmd/fetch_k8s_resource.go
joejstuart: Possibly, but since I don't have a clear picture of it, maybe we leave it out for now.

cmd/fetch_k8s_resource.go
joejstuart: The function is actually not needed so I removed it :)

zregvart (approved):

internal/image/authorization_test.go
```diff
@@ -18,50 +18,62 @@ package image
...
+			RepoUrl:     "my-git-repo",
+			Commit:      "1234",
+			Authorizers: []string{"ec@redhat.com"},
+		},
 	}
 
 	signOff, err := input.GetSignOff()
 	assert.NoError(t, err)
-	assert.Equal(t, expected, signOff)
+	assert.ObjectsAreEqualValues(expected, signOff)
```
lcarva, line 76: Oh! This actually doesn't cause a test failure. It just returns a boolean. [ObjectsAreEqualValues](https://pkg.go.dev/github.com/stretchr/testify/assert#ObjectsAreEqualValues). The tell-tale is that it doesn't take a `TestingT` parameter like `Equal` does. So it does not have a way to mark the test as failed. We should fix that here, and in the other places in this file.

joejstuart: Good catch! I simplified the test and removed that function call.

lcarva (approved):

PR #203, opened 2022-09-28
Regenerate demo images, script to rebuild images

zregvart:
Rebuilds the demo images at `quay.io/hacbs-contract-demo` repository, and adds a script (`hack/rebuild.sh`) for rebuilding those images.


zregvart:
Using non-HACBS pipeline bundle, quick fix after a test locally...

simonbaird (approved):
Lgtm, but I didn't try it.

PR #205, opened 2022-10-04
Restore unit test timeout limit of 500ms

zregvart:
By using golang build tags we can separate the generative, acceptance and unit tests and allow individual settings, like timeout.

This restores the 500ms unit test timeout limit, moves the 30s limit to the generative test and sets the 2s timeout limit for integration tests.


.github/workflows/checks.yaml
```diff
@@ -45,12 +45,24 @@ jobs:
...
         run: make test
 
-      - name: Upload coverage report
+      - name: Upload unit test coverage report
         uses: codecov/codecov-action@v3
         with:
           files: ./coverage-unit.out
           flags: unit
 
+      - name: Upload integration test coverage report
```
lcarva, line 54: What does it mean when multiple coverage reports are uploaded? Do they all get merged?

internal/replacer/replacer_test.go
```diff
@@ -14,6 +14,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+//go:build integration
```
lcarva, line 17: Why are these not `unit` tests? It certainly was the intention.

cmd/replace_test.go
```diff
@@ -14,6 +14,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+//go:build unit
```
lcarva, line 17: TIL: If a file does not define a tag, then it'll always be picked up. This is probably why you added the `unit` tag here. So it doesn't get executed when all we want to run are the acceptance tests.

Makefile
```diff
@@ -60,11 +60,11 @@ website: docs ## Render the website
 
 .PHONY: test
 test: ## Run unit tests
-# Given the nature of generative tests the test timeout was increased from 500ms
+	@go test -race -covermode=atomic -coverprofile=coverage-unit.out -timeout 500ms -tags=unit ./...
```
lcarva, line 63: What's the benefit of splitting up the coverage reports?

lcarva (approved):

.github/workflows/checks.yaml
zregvart: Yeah, Codecov already merges unit and acceptance code coverage reports. The [report](https://app.codecov.io/gh/hacbs-contract/ec-cli/pull/205) has overall the same percent covered, with the expected drop in unit test coverage

internal/replacer/replacer_test.go
zregvart: There is a dependency between `image_test.go` that uses and `replacer_test.go` that defines a constant `testDigest`. `image_test.go` seemed like an integration test (took more than 500ms), now that I revisited it this is not the case, so I'll remove the `integration` tag from this

Makefile
zregvart: So that the next `go test...` doesn't overwrite the coverage report from the previous `go test...`

internal/replacer/replacer_test.go
zregvart: Timeout [exceeded](https://github.com/hacbs-contract/ec-cli/actions/runs/3195743393/jobs/5216779685) for the test in `replacer_test.go`, so I needed to reintroduce the `integration` tag.

PR #207, opened 2022-10-05
HACBS-1180 Add 'exclude_rules' field

robnester-rh:
This commit adds the 'exclude_rules' field in the data.json config map alongside the existing 'non_blocking_checks' field, which is now deprecated.


zregvart:
@robnester-rh are you thinking this would be a first step into getting the configuration needed for the ec-policies and then adding the changes in to the ECP CRD? My initial thought was that we should change the ECP CRD first and then apply it here, but this approach could work also...

robnester-rh:
> @robnester-rh are you thinking this would be a first step into getting the configuration needed for the ec-policies and then adding the changes in to the ECP CRD? My initial thought was that we should change the ECP CRD first and then apply it here, but this approach could work also...

yes, this change won't be "useful" until the work in 1181 is completed, but this can be completed now.

zregvart (approved):

simonbaird (approved):
lgtm

PR #218, opened 2022-10-14
Simplify policy source handling

simonbaird:
The motivation is to be able to use oci containers as policy sources.

From commit message:

    Instead of having a multi-field struct to represent a policy source,
    use a simple string type with a conftest pull compatible go-getter
    style url.
    
    Also
    - Removes some extra processing related to normalizing policy urls
    - Removes some obsolete policy dir handling, since all policies are
      placed under the ./policy directory, we don't need to keep track
      of the directory where each policy source can be found.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/google/addlicense v1.0.0
 	github.com/google/go-containerregistry v0.11.0
 	github.com/google/go-github/v45 v45.2.0
-	github.com/hacbs-contract/enterprise-contract-controller v0.0.0-20221012083301-6cc9977bd09d
+	github.com/hacbs-contract/enterprise-contract-controller v0.0.0-20221014180348-9dd7c428a70a
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/in-toto/in-toto-golang v0.3.4-0.20220709202702-fa494aaa0add
 	github.com/leanovate/gopter v0.2.9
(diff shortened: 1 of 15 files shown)


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/hacbs-contract/ec-cli
 
 go 1.18
 
+// replace github.com/hacbs-contract/enterprise-contract-controller => /home/sbaird/code/enterprise-contract-controller
```
simonbaird, line 5: One of the multiple reasons CI doesn't pass...

zregvart:
[This bit](https://github.com/hacbs-contract/ec-cli/blob/main/internal/acceptance/kubernetes/kubernetes.go#L43) configure the wire mock stub to serve the ECP CR.
The changes to `eval.feature` look good. I’m guessing the issue is with running a different version of the ecc (i.e. not from your fork). We should be okay with changing the ecc definition and then getting a PR to use the new version in the cli, the version is pinned…

simonbaird:
> Does running:
> 
> ```shell
> go get github.com/hacbs-contract/enterprise-policy-controller # perhaps with a `-u`
> # or
> go mod tidy
> ```

Sorry, I didn't explain it well. The `// replace github.com/hacbs-contract/enterprise-contract-controller => /home/sbaird/code/enterprise-contract-controlle` actually works great locally, but can't work in Github CI. So I mean that this is a reason it can't pass here in in Github. Once https://github.com/hacbs-contract/enterprise-contract-controller/pull/17 is merged this problem goes away.

simonbaird:
..and the other reason it was related to mocking the cluster ECP. I figured out what the problem was, largely pebkac.

simonbaird:
https://github.com/hacbs-contract/enterprise-contract-controller/pull/17 is merged, and I updated go.mod to use the latest version. This should probably pass CI now.

zregvart (approved):

PR #219, opened 2022-10-17
Track required tasks

lcarva:
https://issues.redhat.com/browse/HACBS-1239

### DIFF
--- a/cmd/track_bundle.go
+++ b/cmd/track_bundle.go
@@ -18,18 +18,12 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 	"os"
 
-	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/spf13/cobra"
-	"github.com/tektoncd/pipeline/pkg/remote/oci"
-
-	"github.com/hacbs-contract/ec-cli/internal/image"
-	"github.com/hacbs-contract/ec-cli/internal/tracker"
 )
 
-type trackBundleFn func(context.Context, []string, string, tracker.Collector) ([]byte, error)
+type trackBundleFn func(context.Context, []string, string) ([]byte, error)
 
 func trackBundleCmd(track trackBundleFn) *cobra.Command {
 	var data = struct {
@@ -56,7 +50,12 @@ or a digest is required.
 The output is meant to assist enforcement of policies that ensure the
 most recent Tekton Bundle is used. As such, each entry contains an
 "effective_on" date which is set to 30 days from today. This indicates
-the Tekton Bundle usage should be updated within that period.`,
+the Tekton Bundle usage should be updated within that period.
+
+Additionally, the common set of Tasks referenced by all "important"
+Pipeline definitions are deemed required and displayed as such. An
+"important" Pipeline definition is defined as one that does NOT include
+the label "skip-hacbs-test" set to the value "true".`,
 		Example: `Track multiple bundles:
 
   ec track bundle --bundle <IMAGE1> --bundle <IMAGE2>
@@ -75,13 +74,15 @@ Extend an existing tracking file with a new bundle and save changes:
 		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
 
-			out, err := track(cmd.Context(), data.Bundles, data.Input, tektonBundleCollector)
+			out, err := track(cmd.Context(), data.Bundles, data.Input)
 			if err != nil {
 				return err
 			}
 
 			if data.OutputFile == "" {
-				fmt.Println(string(out))
+				if _, err := cmd.OutOrStdout().Write(out); err != nil {
+					return err
+				}
 			} else {
 				f, err := os.Create(data.OutputFile)
 				if err != nil {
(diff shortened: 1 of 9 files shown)


lcarva:
Apologies for the size of this PR. It came out much larger than I had anticipated.

internal/tracker/client.go
```diff
@@ -0,0 +1,59 @@
...
+func NewClient(ctx context.Context) Client {
+	if client, ok := ctx.Value(clientContextKey).(Client); ok && client != nil {
+		return client
+	}
+	return exoClient{}
+}
+
+type exoClient struct{}
+
+func (c exoClient) GetTektonObject(ctx context.Context, bundle, kind, name string) (runtime.Object, error) {
```
joejstuart, line 53: Can't this function and `GetImage` be called without exoClient? I'm not recognizing the need for the interface.

lcarva: It can. This is a pattern that we started using so the client implementation can be replaced with a different one. The use case in mind is supporting unit tests. This was inspired by [this article](https://incident.io/blog/golang-client-mocks). 

The alternative is to create global variables to functions that make an external call, and replace those during unit testing with something that returns mocked data. The problem with this approach, of course, is that it relies on global states which can have unintended side-effects.

internal/tracker/bundle_info.go
```diff
@@ -0,0 +1,143 @@
...
+type bundleInfo struct {
+	ref image.ImageReference
+	// Name of the Task definitions found in the bundle.
+	tasks []string
+	// Name of the Pipeline definitions found in the bundle.
+	pipelines []string
+	// Set of common Tasks found across Pipelines definitions in the bundle.
+	commonPipelineTasks commonTasksRecord
+	// Set of collection where the bundle should be tracked under.
+	collections sets.String
```
zregvart, line 40: It's odd for me to see a set here, seems to me that this is used as a enum instead.

zregvart (commented):
Not sure, put some thoughts in the comments, my first gut reaction is that this is a bit over complicated...

cmd/track_bundle.go
```diff
@@ -56,7 +50,12 @@ or a digest is required.
 The output is meant to assist enforcement of policies that ensure the
 most recent Tekton Bundle is used. As such, each entry contains an
 "effective_on" date which is set to 30 days from today. This indicates
-the Tekton Bundle usage should be updated within that period.`,
+the Tekton Bundle usage should be updated within that period.
+
+Additionally, the commont set of Tasks referenced by all "important"
```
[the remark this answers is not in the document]
lcarva: Thanks!

internal/tracker/bundle_info.go
zregvart: I think I've now realized that the purpose of this, the set threw me of. I guess, this could be a something like:

```go
const (
	PIPELINE			int = 1
	TASK				int = 2
	PIPELINE_AND_TASK	int = 3
)
```

```suggestion
	type int
```

Perhaps I'm overreacting based on the first (wrong) impression...

internal/tracker/client.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/tektoncd/pipeline/pkg/remote/oci"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type Client interface {
+	GetTektonObject(ctx context.Context, bundle, kind, name string) (runtime.Object, error)
```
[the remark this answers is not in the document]
lcarva: The interface does return a runtime.Object. I can rename it to `GetTektonPipelineObject` but that seems misleading. To do, I think it would be best to change the interface to return a Pipeline instead of an Object. However, that would require adding a bit of code in `exoClient.GetTektonObject` to perform the conversion. I think we should avoid doing that because that code is not covered by unit tests only by acceptance tests.

I'm inclined to leave it as is unless you feel strongly about changing.

internal/tracker/bundle_info.go
lcarva: It's a "private" attribute and meant to be used internally by the `bundleInfo` struct. A set simplifies the code quite a bit so I'd prefer not to move away from that.

internal/tracker/bundle_info.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/tektoncd/pipeline/pkg/remote/oci"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/hacbs-contract/ec-cli/internal/image"
+)
+
+type bundleInfo struct {
+	ref image.ImageReference
+	// Name of the Task definitions found in the bundle.
+	tasks []string
```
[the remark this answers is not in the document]
lcarva: Actually... `tasks` is not needed at all. I'm doing a bit of refactoring to `newBundleInfo` to also reduce the need for the `pipelines` attribute. Stay tuned.

internal/tracker/client.go
joejstuart: That makes sense. Thanks for the explanation. I've dealt with this by passing the functions to the struct that needs them, so I think in this case the `bundleInfo` struct would have fields `GetTektonObject` and `GetImage`, but I'm not sure how you'd test the `Track` method using that pattern.

internal/tracker/tracker.go
```diff
@@ -23,22 +23,47 @@ import (
...
 
-type Tracker map[string]map[string][]Record
+type commonTasksRecord struct {
+	Tasks       []string  `json:"tasks"`
+	EffectiveOn time.Time `json:"effective_on"`
+	initialized bool
+}
+
+func (r *commonTasksRecord) updateTasksIntersection(newTasks sets.String) {
+	if !r.initialized {
```
zregvart, line 53: Perhaps, remove the need for `initialized
```suggestion
	if len(r.Tasks) == 0 {
```

internal/tracker/client.go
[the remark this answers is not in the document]
zregvart: My main concern is that this is a leaky abstraction and a bit overly generic. Let's leave it as is and refactor if we feel like it.

internal/tracker/bundle_info.go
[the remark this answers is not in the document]
zregvart: Much clearer, thanks!

zregvart (approved):
The usual, unusual nitpickery

internal/tracker/tracker.go
lcarva: Actually... a pipeline that has no tasks is probably something we want to ignore. I'll switch the logic (and get rid of `initialized`).

lcarva: Done.

PR #226, opened 2022-10-20
Add `-trimpath` to `go build`

zregvart:
With this local paths won't be present in the built binary or stack traces.


robnester-rh (commented):
LGTM

robnester-rh (approved):

lcarva (approved):

PR #233, opened 2022-10-24
Use k8s.io/client-go directly, configure logs

zregvart:
To reduce the depenencies a little we can use k8s.io/client-go directly and not via sigs.k8s.io/controller-runtime. The klog logs are now forwarded to logrus instead of going to stdout. The caveat being that all klog logs are logged with INFO category, though they do comply with the logging level set, as the filtering is done to make sure the correct logging level is propagated -- the sink bridge logs with `Info` regardless.

### DIFF
--- a/internal/kubernetes/client.go
+++ b/internal/kubernetes/client.go
@@ -22,9 +22,13 @@ import (
 
 	ecc "github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	controllerruntime "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 type contextKey string
@@ -36,7 +40,13 @@ type Client interface {
 }
 
 type kubernetesClient struct {
-	client client.Client
+	client dynamic.Interface
+}
+
+var kubeconfig string
+
+func AddKubeconfigFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&kubeconfig, "kubeconfig", "", "path to the Kubernetes config file to use")
 }
 
 func WithClient(ctx context.Context, client Client) context.Context {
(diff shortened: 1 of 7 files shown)


simonbaird (approved):

internal/kubernetes/client.go
```diff
@@ -36,7 +40,13 @@ type Client interface {
 }
 
 type kubernetesClient struct {
-	client client.Client
+	client dynamic.Interface
+}
+
+var kubeconfig string
+
+func AddKubeconfigFlag(cmd *cobra.Command) {
```
robnester-rh, line 48: I think that flag declarations should be in the `cmd` package, not in internal, no?

zregvart: I'm still not sold on this being the best idea, the reason I added this was because we did have the `controller-runtime` [sneak](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.12.3/pkg/client/config/config.go#L38-L40) the `-kubeconfig` flag on us, which left it undocumented for us.

I've considered:

- adding the flag to the cmd and have it mutate a public state in `internal/kubernetes`
- adding the flag to the cmd and pass the state towards `internal/kubernetes` via function arguments
- having public state relating to `internal/kubernetes` within `cmd`
- mutating the `cmd` from `init` in `internal/kubernetes`
- (what I implemented in the end) invoking from `cmd` a mutation on `cmd` in `internal/kubernetes`

My thought process was, the state (`kubeconfig` var) should be private to `internal/kubernetes`, not really a concern of other packages. Passing it through for every invocation leading to the `internal/kubernetes/client.go` is a nuisance -- we have three code paths determining the the policy configuration (flags, inline spec, fetched from k8s) it seemed bad to introduce a k8s concern into all the code paths (see `internal/policy/policy.go::NewPolicy`). And finally, invoking `cmd` from `internal/kubernetes` leads to cyclic error.

Suggestions on how to handle this nicer would be much appreciated!

internal/kubernetes/client.go
```diff
@@ -36,7 +40,13 @@ type Client interface {
 }
 
 type kubernetesClient struct {
-	client client.Client
+	client dynamic.Interface
+}
+
+var kubeconfig string
```
lcarva, line 46: It would be great if we didn't have to introduce a global state like this. I'm surprised that github.com/kubernetes-sigs/controller-runtime also takes this approach!

internal/kubernetes/client.go
lcarva: I follow how you approached this solution, and appreciate all the considerations. I don't have another idea other than stuffing a client into the context early on (but that somehow feels a bit abusive?).

What if we don't provide the flag at all? The `KUBECONFIG` environment variable works just fine. We could document that this is how you can specify an alternate path. Wdyt?

zregvart: To me, having a documented flag with client-go (what's suggested here) is better than having an undocumented flag with controller-runtime (what we have now).
I think using the environment variable is functionally and architecturally equivalent of using a global, static and exported variable. The package-bound non-exported variable should fare better than that with respect to architecture.

PR #238, opened 2022-10-27
Refactor Afero usage

zregvart:
Having a global `afero.Fs` variable and running tests concurrently will lead to test flakiness, best refactor that to make sure that we have a single `afero.Fs` variable per package and reduce any flakiness to the tests affecting that single package.


joejstuart (approved):

simonbaird (approved):

robnester-rh (approved):
LGTM

PR #239, opened 2022-10-27
Evaluate each policy source independently

zregvart:
This changes the policy evaluation to evaluate each policy source independently, instead of downloading from all policy sources and then evaluate once with using all policy files from all policy sources.

In addition to that handling of the hard coded data source from the https://github.com/hacbs-contract/ec-policies is now in the evaluator package so that the downloader package has a single responsibility -- to download from the provided sources.

Along with that the logic of where to place the policy files has been moved to the source package for the same reason.

Includes #238, so review that one first

Ref. https://issues.redhat.com/browse/HACBS-1311


internal/policy/source/source.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+//PolicyUrl is a string containing a go-getter style source url compatible with conftest pull
+type PolicyUrl string
+
+// GetPolicies clones the repository for a given PolicyUrl
+func (p *PolicyUrl) GetPolicy(ctx context.Context, workDir string, showMsg bool) (string, error) {
+	sourceUrl := string(*p)
+
+	dest := uniquePolicyDestination(workDir, sourceUrl)
```
joejstuart, line 56: Can you just call `p.PolicyUrl()` to get `sourceUrl`?

zregvart: Tots, I've added `PolicyUrl` method, didn't refactor this bit

robnester-rh (approved):
LGTM

joejstuart (approved):

internal/evaluator/conftest_evaluator.go
```diff
@@ -33,207 +33,201 @@ import (
...
 
-	jsonData, marshalErr := json.MarshalIndent(config, "", "    ")
-	if marshalErr != nil {
+	configJSON, err := json.MarshalIndent(config, "", "    ")
+	if err != nil {
 		return err
 	}
 	// Check to see if the data.json file exists
-	exists, err = afero.Exists(utils.AppFS, dataFilePath)
+	exists, err := afero.Exists(fs, configFilePath)
```
simonbaird, line 184: More of a golang question than a review comment.. It seems like like the exists var was never declared here. How does that work?

simonbaird:
I tested it locally, works fine afaict.

```
ec validate image --image $IMG --policy '{"publicKey":"sigkey.pub","sources":["github.com/hacbs-contract/ec-policies//policy"],"configuration":{"excludeRules":["test","tasks","attestation_task_bundle"]}}' | yq -P
success: false
components:
  - name: Unnamed
    containerImage: quay.io/cuipinghuo/single-container-app:9f5d549dd64aacf10e3baac90972dfd5df788324
    violations:
      - msg: It just feels like a bad day to do a release
        metadata:
          code: bad_day
          effective_on: "2022-01-01T00:00:00Z"
    warnings: []
    success: false
```

simonbaird (approved):

internal/evaluator/conftest_evaluator.go
robnester-rh: the `afero.Exists()` method returns two values, a boolean and an error. This is just assigning those returns to variable names to evaluate further.

PR #240, opened 2022-10-28
Add option to condense conftest output

joejstuart:
This adds an option to shorten the report output. The output is restricted to one error message per short_name that's defined in the policy annotation.

### DIFF
--- a/cmd/validate_image.go
+++ b/cmd/validate_image.go
@@ -47,7 +47,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		output              string
 		spec                *appstudioshared.ApplicationSnapshotSpec
 		policy              *ecc.EnterpriseContractPolicySpec
-		shortReport         bool
+		summary             bool
 	}{
 		policyConfiguration: "ec-policy",
 	}
@@ -99,6 +99,10 @@ Return a non-zero status code on validation failure:
 
   ec validate image --image registry/name:tag --strict
 
+Return a summary of test results:
+
+  ec validate image --image registry/name:tag --policy my-policy --summary
+
 Use an EnterpriseContractPolicy resource from the currently active kubernetes context:
 
   ec validate image --image registry/name:tag --policy my-policy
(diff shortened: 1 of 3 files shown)


internal/applicationsnapshot/report.go
```diff
@@ -30,13 +31,50 @@ type Component struct {
 	Success    bool            `json:"success"`
 }
 
-type report struct {
+type fullReport struct {
```
zregvart, line 34: I think the struct can still be called report...
```suggestion
type report struct {
```

internal/applicationsnapshot/report.go
```diff
@@ -30,13 +31,50 @@ type Component struct {
...
 	Success    bool        `json:"success"`
 	Components []Component `json:"components"`
 }
 
+type shortReport struct {
+	Components []shortComponent `json:"components"`
+	Success    bool             `json:"success"`
+}
+
+type shortComponent struct {
```
zregvart, line 44: Maybe:
```suggestion
type componentSummary struct {
```

internal/applicationsnapshot/report.go
```diff
@@ -30,13 +31,50 @@ type Component struct {
...
+	TotalViolations int                 `json:"total_violations"`
+	TotalWarnings   int                 `json:"total_warnings"`
+}
+
+func NewReport(components []Component, shortReport bool) (string, error, bool) {
+	var j []byte
+	var err error
+	var success bool
+	if shortReport {
+		report := condensedReport(components)
```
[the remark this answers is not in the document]
joejstuart: absolutely, just pushed the changes. thanks!

cmd/validate_image.go
```diff
@@ -98,14 +99,22 @@ Return a non-zero status code on validation failure:
...
+
 Use an EnterpriseContractPolicy resource from the currently active kubernetes context:
 
   ec validate image --image registry/name:tag --policy my-policy
 
 Use an EnterpriseContractPolicy resource from a different namespace:
 
   ec validate image --image registry/name:tag --policy my-namespace/my-policy
+  
+Use an EnterpriseContractPolicy resource from a different namespace:
```
lcarva, line 114: This example looks identical to the one above (including the description)

cmd/validate_image.go
```diff
@@ -220,6 +230,7 @@ Use an inline EnterpriseContractPolicy spec
 		"write output to a file. Use empty string for stdout, default behavior")
 	cmd.Flags().BoolVarP(&data.strict, "strict", "s", data.strict,
 		"return non-zero status on non-successful validation")
+	cmd.Flags().BoolVarP(&data.summary, "summary", "c", data.summary, "print condensed output of conftest")
```
lcarva, line 233: I think mentioning `conftest` here is not very helpful to the user. Maybe, something like `condense violation and warning messages`?

joejstuart: I agree. updated.

cmd/validate_image.go
joejstuart: yep, copy/paste mistake.

zregvart (approved):
Looks good! I think one semi-obvious followup we can do is to have a more elaborate `--output` that not only takes into account different formats (JSON/YAML) but also can do `summary`, `template`, `jsonpath` -- similar to kubectl (created [HACBS-1353](https://issues.redhat.com/browse/HACBS-1353))

PR #243, opened 2022-11-01
Use indirect filesystem usage via afero

zregvart:
This helps with writing tests against an in-memory filesystem. The design focuses on passing the `afero.Fs` instance as parameter from the `cmd` package where the instance is determined from the `context.Context` using the helper `fs(context.Context) afero.Fs` function. Having it pass as a parameter makes it very evident when writing tests that the `afero.Fs` can be replaced, and no dependency cycles are introduced.
Some parts of the code, namely parts that use `go-git`, are not changed and still access the filesystem directly, this is because `go-git` doesn't support afero directly.
Perhaps a followup could be to try using go-billy-afero[1].

Ref. https://issues.redhat.com/browse/HACBS-1337

[1] https://github.com/jonfriesen/go-billy-afero


lcarva:
I'm failing to see the benefits of this approach which may be obvious to others.

simonbaird (approved):
I think probably the consistency (since it appears we're using afero in a couple of places already) and the testing benefits outweigh the extra complexity, so I'll give it a lgtm.

PR #245, opened 2022-11-03
github workflow for tekton

joejstuart:


.github/workflows/tekton.yaml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: setup kind cluster
+        uses: container-tools/kind-action@v2
+
+      - name: Run ec task
+        run: |
```
lcarva, line 40: Can we toss all of this into a bash script so we can run it against a local kind cluster?

joejstuart: Absolutely, I still have some cleanup to do.

deployments/ecp-policy.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: EnterpriseContractPolicy
+metadata:
+  creationTimestamp: null
+  name: ec-policy
+spec:
+  description: My custom enterprise contract policy configuration
+  sources:
+    - git::https://github.com/hacbs-contract/ec-policies.git
+  exceptions:
+    nonBlocking:
```
zregvart, line 11: For followup: we should be using the non-deprecated `configuration/excludeRules` here

.github/workflows/tekton.yaml
```diff
@@ -0,0 +1,72 @@
...
+          # install tekton and wait for controllers to start
+          tekton_version=v0.38.4
+          kubectl apply --filename https://storage.googleapis.com/tekton-releases/pipeline/previous/$tekton_version/release.yaml
+          kubectl wait pods -n tekton-pipelines -l pipeline.tekton.dev/release=$tekton_version --for=condition=Ready --timeout=90s
+
+          # setup resources for the ec task to run
+          kubectl create secret generic cosign-public-key --from-file=cosign.pub=hack/cosign.pub
+          kustomize build https://github.com/hacbs-contract/enterprise-contract-controller/config/crd?ref=main | kubectl apply -f -
```
zregvart, line 48: I prefer using `kubectl kustomize`, one less binary to make sure is available...

.github/workflows/tekton.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: setup kind cluster
+        uses: container-tools/kind-action@v2
+
+      - name: Run ec task
+        run: |
+          # install tekton and wait for controllers to start
+          tekton_version=v0.38.4
+          kubectl apply --filename https://storage.googleapis.com/tekton-releases/pipeline/previous/$tekton_version/release.yaml
+          kubectl wait pods -n tekton-pipelines -l pipeline.tekton.dev/release=$tekton_version --for=condition=Ready --timeout=90s
```
zregvart, line 44: We could probably wait more towards the end and proceed with the creation of other resources; probably doesn't make much of a difference...

deployments/verify-enterprise-contract-v2-taskrun.yaml
```diff
@@ -0,0 +1,44 @@
...
+spec:
+  params:
+  - name: HOMEDIR
+    value: /tekton/home
+  - name: IMAGES
+    value: |
+      {
+        "components": [
+          {
+            "containerImage": "quay.io/hacbs-contract-demo/single-container-app@sha256:84550a596e40024e6ea10ce5aab9ad5b0b91a14cb27161654722ab1d11c70ac2"
```
zregvart, line 25: Would be cool to pull this from `hack/images.txt` so we don't need to think about updating it here when that changes

deployments/verify-enterprise-contract-v2-taskrun.yaml
```diff
@@ -0,0 +1,44 @@
...
+          }
+        ]
+      }
+  - name: POLICY_CONFIGURATION
+    value: $(context.taskRun.namespace)/ec-policy
+  - name: PUBLIC_KEY
+    value: k8s://$(context.taskRun.namespace)/cosign-public-key
+  - name: REKOR_HOST
+    value: ""
```
zregvart, line 34: This is the default, so just wondering if we need it here?

zregvart (approved):
Very nice! In terms of followup on this in a separate PR, we could install everything via kustomize; i.e. all those separate kubectl apply/create could be merged into a single `kustomization.yaml` file.

Another followup could be to get this workflow triggered via `repository_dispatch` on changes to ec-policies repository.

I think it is more beneficial to get this merged and then modify as needed.

.github/workflows/tekton.yaml
```diff
@@ -0,0 +1,72 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: setup kind cluster
+        uses: container-tools/kind-action@v2
+
+      - name: Run ec task
+        run: |
+          # install tekton and wait for controllers to start
+          tekton_version=v0.38.4
```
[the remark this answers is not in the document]
joejstuart: I might have to come back to this. v0.38.4 is the latest version that works with the version of kubernetes that's being used.

PR #248, opened 2022-11-10
create task bundle and use catalog layout

joejstuart:
This PR is to change the directory structure of the task to replicate the tekton catalog and to package the task into a tekton bundle. 
- Move the EC task, taskRun and policy to the tasks directory. 
- The tasks directory follows the tekton catalog layout (https://github.com/tektoncd/catalog/tree/main/task/buildah) . 
- After CI passes, a task bundle is created and pushed to an OCI (quay) registry.


simonbaird:
It's not very clear what's going on. Maybe some explanatory commit messages would help?

lcarva:
The tasks directory and the value of the label `app.kubernetes.io/version` should match. Can we just set the label to 0.1? It has a new home, so I think it's ok to reset the version.

task/tests/run.sh
```diff
@@ -0,0 +1,46 @@
...
+status=$(kubectl get $tr -o jsonpath='{.status.conditions[*].status}')
+
+if [[ "$status" == "False" ]]; then
+  echo "task {$tr} failed"
+  echo "debugging info"
+  kubectl describe $tr
+  exit 1
+fi
+
+tkn bundle push ${OCI_REPO}:$TASK_VERSION -f "${TASK_DIR}/${TASK_FILE}"
```
lcarva, line 46: Maybe as a follow up, but I think we want to push the bundle to a unique location first, say the file digest as the tag. Then use something like `skopeo copy` to tag it with the version tag. If we don't do that, the previous image won't have any tags and quay.io will delete that image, breaking any references by digest.

.github/workflows/tekton.yaml
```diff
@@ -27,18 +27,32 @@ name: Checks
...
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v3
 
       - name: setup kind cluster
         uses: container-tools/kind-action@v2
 
-      - name: Run ec task
+      - name: install tkn cli
```
zregvart, line 37: Found [this action](https://github.com/marketplace/actions/install-tkn), which does basically the same as here. I do like having versions managed in a single place, perhaps in a followup we can see about using the [tools approach](https://github.com/hacbs-contract/ec-cli/blob/9b3412a103f0ec3b70cb4ceaeb4750ffdcac85e8/internal/tools/tools.go) and version via `go.mod`.

zregvart (approved):
Looks good. I think we could (also?) push to GitHub Container registry (as we do/did for the Tekton Chains fork). We could to this all in a followup.

PR #249, opened 2022-11-10
Flexible output formats

lcarva:
```
Add flexible output options to validate image

This change allows the ec-cli to produce a report output to multiple
locations in different formats.

https://issues.redhat.com/browse/HACBS-1353
```

### DIFF
--- a/cmd/validate_image.go
+++ b/cmd/validate_image.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"sync"
 
 	ecc "github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
@@ -45,10 +44,10 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		strict              bool
 		input               string
 		filePath            string
-		output              string
+		outputFile          string
+		output              []string
 		spec                *appstudioshared.ApplicationSnapshotSpec
 		policy              *ecc.EnterpriseContractPolicySpec
-		summary             bool
 	}{
 		policyConfiguration: "ec-policy",
 	}
@@ -100,10 +99,6 @@ Return a non-zero status code on validation failure:
 
   ec validate image --image registry/name:tag --strict
 
-Return a summary of test results:
-
-  ec validate image --image registry/name:tag --policy my-policy --summary
-
 Use an EnterpriseContractPolicy resource from the currently active kubernetes context:
 
   ec validate image --image registry/name:tag --policy my-policy
@@ -113,7 +108,13 @@ Use an EnterpriseContractPolicy resource from a different namespace:
   ec validate image --image registry/name:tag --policy my-namespace/my-policy
 
 Use an inline EnterpriseContractPolicy spec
-  ec validate image --image registry/name:tag --policy '{"publicKey": "<path/to/public/key>"}'`,
+  ec validate image --image registry/name:tag --policy '{"publicKey": "<path/to/public/key>"}'
+
+Write output in JSON format to a file
+  ec validate image --image registry/name:tag --output json=<path>
+
+Write output in YAML format to stdout and in HACBS format to a file
+  ec validate image --image registry/name:tag --output yaml --output hacbs=<path>`,
 		PreRunE: func(cmd *cobra.Command, args []string) (allErrors error) {
 			ctx := cmd.Context()
 			if s, err := applicationsnapshot.DetermineInputSpec(
@@ -187,26 +188,19 @@ Use an inline EnterpriseContractPolicy spec
 					components = append(components, r.component)
 				}
 			}
-
-			report, err, success := applicationsnapshot.NewReport(components, data.summary)
-
 			if allErrors != nil {
-				return multierror.Append(allErrors, err)
+				return allErrors
 			}
 
-			if len(data.output) > 0 {
-				if err := ioutil.WriteFile(data.output, []byte(report), 0644); err != nil {
-					return multierror.Append(allErrors, err)
-				}
-				fmt.Printf("Report written to %s\n", data.output)
-			} else {
-				_, err := cmd.OutOrStdout().Write([]byte(report))
-				if err != nil {
-					return multierror.Append(allErrors, err)
-				}
+			if len(data.outputFile) > 0 {
+				data.output = append(data.output, fmt.Sprintf("%s=%s", applicationsnapshot.JSON, data.outputFile))
+			}
+			report := applicationsnapshot.NewReport(components)
+			if err := report.WriteAll(data.output, cmd.OutOrStdout(), fs(cmd.Context())); err != nil {
+				return err
 			}
 
-			if data.strict && !success {
+			if data.strict && !report.Success {
 				// TODO: replace this with proper message and exit code 1.
 				return errors.New("success criteria not met")
 			}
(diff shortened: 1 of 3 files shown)


internal/applicationsnapshot/report.go
```diff
@@ -50,28 +59,27 @@ type componentSummary struct {
...
-		j, err = json.Marshal(report)
-	}
-
-	if err != nil {
-		return "", err, false
-	}
-
-	return string(j), nil, success
+// hacbsReport represents the standardized HACBS_TEST_OUTPUT format.
+type hacbsReport struct {
```
lcarva, line 63: @dirgim fyi

internal/applicationsnapshot/report.go
```diff
@@ -31,8 +39,9 @@ type Component struct {
 	Success    bool            `json:"success"`
 }
 
-type report struct {
-	Success    bool        `json:"success"`
+type Report struct {
```
joejstuart, line 42: any reason to make this public?

lcarva: The `NewReport` function now returns an instance of Report instead of `[]byte`. I think technically this could be private, but it does seem odd to return a private type from a public function. I can change it back if you feel strongly about this.

joejstuart: I guess it depends, if access (from outside of the package) to `Report` should always come from `NewReport` I would make it private. Otherwise, it's not a big deal.

internal/applicationsnapshot/report.go
dirgim: Interesting, we have the same struct here (will be updated with the `int` fields soon): https://github.com/redhat-appstudio/integration-service/blob/main/helpers/integration.go#L42
Does it make sense to import it from there?

internal/applicationsnapshot/report.go
```diff
@@ -132,3 +172,94 @@ func condensedMsg(results []output.Result) map[string][]string {
 	}
 	return shortNames
 }
+
+// toHacbs returns a version of the report that conforms to the
+// HACBS_TEST_OUTPUT format.
+func (r *Report) toHacbs() hacbsReport {
```
zregvart, line 178: nitpickery:
```suggestion
func (r *Report) toHACBS() hacbsReport {
```

internal/applicationsnapshot/report.go
```diff
@@ -132,3 +172,94 @@ func condensedMsg(results []output.Result) map[string][]string {
...
+	case 1:
+		format = parts[0]
+	case 2:
+		format = parts[0]
+		path = parts[1]
+	}
+
+	if len(format) == 0 {
+		format = JSON
+	}
```
zregvart, line 242: nitpickery

```suggestion
	parts := strings.SplitN(target, "=", 2)
	var format, path string
	switch len(parts) {
		case 0:
			format = JSON
		case 1:
			format = parts[0]
		case 2:
			format = parts[0]
			path = parts[1]
	}
```

internal/applicationsnapshot/report.go
```diff
@@ -132,3 +172,94 @@ func condensedMsg(results []output.Result) map[string][]string {
...
+}
+
+// fileWriter is a simple struct that allows writing to a file on-demand.
+type fileWriter struct {
+	path string
+}
+
+// Write the given data to a certain path.
+func (w fileWriter) Write(data []byte) (int, error) {
+	file, err := os.OpenFile(w.path, os.O_WRONLY|os.O_CREATE, 0644)
```
zregvart, line 259: Could use afero here

zregvart (commented):
We could be testing the file output as well.

cmd/validate_image.go
```diff
@@ -113,7 +108,13 @@ Use an EnterpriseContractPolicy resource from a different namespace:
...
 
 Use an inline EnterpriseContractPolicy spec
-  ec validate image --image registry/name:tag --policy '{"publicKey": "<path/to/public/key>"}'`,
+  ec validate image --image registry/name:tag --policy '{"publicKey": "<path/to/public/key>"}'
+
+Write output in JSON format to a file
+  ec validate image --image registry/name:tag --output json=<path>
+
+Write output in YAML format to stdout and in HACBS format to a file
+  ec validate image --image registry/name:tag --output yaml --output hacbs=<path>`,
```
simonbaird, line 117: I like it.

deployments/verify-enterprise-contract.yaml
```diff
@@ -114,10 +117,13 @@ spec:
...
-        - "$(results.REPORT.path)"
-        - "--summary"
+        - "--strict=$(params.STRICT)"
+        - "--output"
+        - "yaml"
+        - "--output"
+        - "json=$(results.REPORT.path)"
+        - "--output"
+        - "hacbs=$(results.HACBS_TEST_OUTPUT.path)"
       env:
```
simonbaird, line 127: Nice!

cmd/validate_image.go
simonbaird: I was initially thinking maybe the json vs yaml would orthogonal to the structure, e.g. what if I want the hacbs shaped output in yaml format? But now I feel like this is fine. It's easy enough to expand the output format, `hacbs-yaml`, `hacbs-xml` etc if we feel the need.

cmd/validate_image.go
```diff
@@ -225,12 +219,12 @@ Use an inline EnterpriseContractPolicy spec
 		"path to ApplicationSnapshot Spec JSON file")
 	cmd.Flags().StringVarP(&data.input, "json-input", "j", data.input,
 		"JSON represenation of an ApplicationSnapshot Spec")
-	cmd.Flags().StringVarP(&data.output, "output-file", "o", data.output,
-		"write output to a file. Use empty string for stdout, default behavior")
+	cmd.Flags().StringSliceVar(&data.output, "output", data.output,
+		"write output to a file in a specific format. Use empty string path for stdout. May be used multiple times. Possible formats are json, yaml, hacbs, and summary")
```
simonbaird, line 223: Is summary just the one word "failure" or "success"? If so then maybe that could be made more clear somehow.

simonbaird (approved):
Some comments and notes but no blockers afaict, assuming you consider the other reviewers comments and nitpicks.

internal/applicationsnapshot/report.go
lcarva: That's what I had initially, but then I decided to test this: `--output=`. This hits `case 1` but sets the format to an empty string.

internal/applicationsnapshot/report.go
lcarva: It's true. Maybe it's a controversial opinion, but I don't think a public method/function should ever have a private return type in its signature. It can make the callee code quite odd (for instance, callee can't declare a variable of that type).

If we want to hide the implementation, then I think the proper solution is to create a public interface which is used as the return type in the method/function signature. Then we can return a private type that conforms to that interface. I think this is overkill here. After all, this is all part of internal packages.

cmd/validate_image.go
lcarva: Yeah. The reason I added yaml is that it's much easier on the eyes and makes it good for throwing in the logs. I could've gone with something like `jsonpretty`. yaml just seemed like a neater solution here.

cmd/validate_image.go
lcarva: It's the "condensed" version of the output. I think we should probably remove it, but didn't want to make that decision here.

PR #267, opened 2022-11-16
Rollback gci and golangci-lint versions

zregvart:
At some point we diverged in the version of gci from the version golangci-lint is using. That led to issues with gci not reporting any issues and golangci-lint reporting bogus issues. Oddly enough I could only manifest this issues locally, the GitHub Checks workflow never failed. Given that we're not using any of the newer features of either gci or golangci-lint, I opted to rollback their versions  instead.

gci from 0.8.2 to 0.8.0 and golangci-lint from 1.50.0 to 1.50.1.


zregvart:
Turns out it was a Golang version issue. The problem would manifest with 1.18 but not with 1.18.7 which is used on GitHub Workflows.

PR #271, opened 2022-11-17
Install git-core instead of git

zregvart:
This way we don't have `perl` and `emacs-filesystem` in the image.


lcarva (approved):
Thank you!

simonbaird (approved):

PR #272, opened 2022-11-17
Use our fork of the release action

zregvart:
GitHub issues deprecation warnings in the release workflow and there is a PR upstream that is not merged (yet). This uses our fork with that PR on top to resolve those warnings.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,7 +61,13 @@ jobs:
           git push -f --tags
 
       - name: Release
-        uses: softprops/action-gh-release@v1
+        # we're using our own fork with
+        # https://github.com/softprops/action-gh-release/pull/275 on top of it,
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
           git push -f --tags
 
       - name: Release
-        uses: softprops/action-gh-release@v1
+        uses: hacbs-contract/action-gh-release@v2
```
lcarva, line 64: Can we add a TODO comment here? It would also be great to add a link to the upstream issue we're expecting to be resolved.

lcarva (approved):

lcarva (approved):

.github/workflows/release.yaml
zregvart: Good point :)

PR #275, opened 2022-11-18
Effective time flag

simonbaird:
Add a new flag, `--effective-time` and use its value instead of the current time when evaluating policies.

[HACBS-1219](https://issues.redhat.com/browse/HACBS-1219)

### DIFF
--- a/features/validate_image.feature
+++ b/features/validate_image.feature
@@ -114,7 +114,79 @@ Feature: evaluate enterprise contract
     }
     """
 
+  Scenario: future failure is converted to a warning
+    Given a key pair named "known"
+    Given an image named "acceptance/ec-happy-day"
+    Given a valid image signature of "acceptance/ec-happy-day" image signed by the "known" key
+    Given a valid Rekor entry for image signature of "acceptance/ec-happy-day"
+    Given a valid attestation of "acceptance/ec-happy-day" signed by the "known" key
+    Given a valid Rekor entry for attestation of "acceptance/ec-happy-day"
+    Given a git repository named "future-deny-policy" with
+      | main.rego | examples/future_deny.rego |
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/ec-happy-day --policy {"sources":["git::http://${GITHOST}/git/future-deny-policy.git"]} --public-key ${known_PUBLIC_KEY} --rekor-url ${REKOR} --strict"
+    Then the exit status should be 0
+    Then the standard output should contain
+    """
+    {
+      "success": true,
+      "key": ${known_PUBLIC_KEY_JSON},
+      "components": [
+        {
+          "name": "Unnamed",
+          "containerImage": "localhost:(\\d+)/acceptance/ec-happy-day",
+          "violations": [],
+          "warnings": [
+            {
+              "metadata": {
+                "effective_on": "2099-01-01T00:00:00Z"
+              },
+              "msg": "Fails in 2099"
+            }
+          ],
+          "success": true,
+          "signatures": ${ATTESTATION_SIGNATURES_JSON}
+        }
+      ]
+    }
+    """
+
   @focus
+  Scenario: future failure is a deny when using effective-date flag
+    Given a key pair named "known"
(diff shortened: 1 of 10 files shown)


simonbaird:
I don't love how there are 300 one line diffs in 200 files, so I'm thinking about how to improve that situation.

simonbaird:
Rebased and also changed the design quite a bit. Still doesn't have test coverage, so still in draft.

simonbaird:
When adding the acceptance test I realized the patch doesn't pass the effective time in to the `Evaluate` function in `internal/evaluator/conftest_evaluator.go`. I had been just adding it to config.json and assuming that was enough, which is no longer true.

So there's more work to do and this will stay in draft a little longer.

features/validate_image.feature
```diff
@@ -114,7 +114,79 @@ Feature: evaluate enterprise contract
...
+            }
+          ],
+          "success": true,
+          "signatures": ${ATTESTATION_SIGNATURES_JSON}
+        }
+      ]
+    }
+    """
+
   @focus
```
lcarva, line 153: Should this be removed?

internal/evaluator/conftest_evaluator.go
```diff
@@ -143,34 +146,39 @@ func createConfigJSON(fs afero.Fs, dataDir string, p *policy.Policy) error {
...
 		}
 	}
-	// Check to see that we've actually added any values to the policyConfig struct.
-	// If so, we'll update the config map. Otherwise, this is skipped.
-	if (policyConfig{} != *pc) {
-		config["config"] = map[string]interface{}{
-			"policy": pc,
-		}
+
+	// Now that the future deny logic is handled in the ec-cli and not in rego,
```
lcarva, line 173: This. I really like the fact that the time we're evaluating against does not change between each image being validated.

internal/policy/policy_test.go
```diff
@@ -55,6 +56,9 @@ func (c *FakeKubernetesClient) FetchEnterpriseContractPolicy(ctx context.Context
 }
 
 func TestNewPolicy(t *testing.T) {
+	timeNowStr := "2022-11-23T16:30:00Z"
+	timeNow, _ := time.Parse(time.RFC3339, timeNowStr)
```
lcarva, line 60: For our own future sanity, can we fail the test if `time.Parse` does return an error?

lcarva (approved):

internal/acceptance/cli/cli.go
```diff
@@ -147,6 +147,9 @@ func ecCommandIsRunWith(ctx context.Context, parameters string) (context.Context
 		logger.Log(err)
 	}
 
+	logger.Logf("Stdout: %v", stdout.String())
+	logger.Logf("Stderr: %v", stderr.String())
+
```
zregvart, line 152: There is already `logOutput` that could be used instead, it prints much more. Having this here the stdout/stderr will be printed twice on unexpected exit status, might be a bit confusing

zregvart (approved):

zregvart:
Let's merge this and address the minor issues in a followup PR

internal/evaluator/conftest_evaluator.go
simonbaird: +1, I can imagine some potential weirdness if one conftest invocation ran a millisecond before midnight and the next ran a millisecond after.

internal/acceptance/cli/cli.go
simonbaird: Looks like this has been dealt with via ed0db955c00d0c51c0f1eb4ec96b584ff0aa2fec .

features/validate_image.feature
simonbaird: Done in e3688ebdc7f68dab00e9859b9a21ff4d9555a2a0

internal/policy/policy_test.go
simonbaird: 8e42aa0b6b4e1aa55dab15dbdfdb63b822399a04

simonbaird:
Thanks for all the fixups :sweat_smile:

PR #276, opened 2022-11-21
`hack/setup-environment.sh` improvements

zregvart:
Now the `hack/setup-environment.sh` in addition to configuring Tekton Pipelines and the EnterpriseContractPolicy custom resource and permissions, also configures an image registry that the ec-cli image and the verify-enterprise-contract Tekton Task bundle can be pushed to.

The versions of the Tekton Pipeline and EnterpriseContractPolicy custom resource are governed by the versions used in the `go.mod` file.

Additionally a namespace named `work` is created where the required permissions are granted to the `default` ServiceAccount.

For each component (Tekton Pipeline, ECC, registry, work) a kustomize directory is created to keep the descriptors together and help with further customization.

Ref. https://issues.redhat.com/browse/HACBS-1424 - this is the refactor needed for enabling the tests on pull requests


zregvart:
Added:

[Task test improvements](https://github.com/hacbs-contract/ec-cli/pull/276/commits/885e231b9a0b314ae641b2029c05047bd76802b9)

This moves the workflow from `tekton.yaml` to `checks.yaml`, makes sure that the Task test runs with the configurable image and waits for the Task to finish either successfully or unsuccessfully.

[Make sure to download all dependencies](https://github.com/hacbs-contract/ec-cli/pull/276/commits/497db7e6132dbae353cfcc5f07b72a267508171c)

This includes downloading dependencies from internal/tools.

.github/workflows/checks.yaml
```diff
@@ -72,3 +74,30 @@ jobs:
         with:
           files: ./coverage-acceptance.out
           flags: acceptance
+
+  Task:
+    runs-on: ubuntu-latest
+    env:
```
joejstuart, line 80: Would it make sense to add `ECC_VERSION: ${{ github.sha }}` here to avoid the clone in setup-environment.sh?

joejstuart (approved):
Looks great!

PR #277, opened 2022-11-21
Expose attestation signatures in output

lcarva:
https://issues.redhat.com/browse/HACBS-1280


zregvart (approved):
Lovely :)

zregvart:
Oh, one note, there is support for regex-based comparison in the acceptance feature files, like this:

https://github.com/hacbs-contract/ec-cli/blob/905ed7fd1855d12cbe84ac9de9e9b6f8746ee015/features/validate_image.feature#L37

So in the acceptance test we don't need to care about having the exact output, instead we can check the structure. That makes the acceptance tests catch unexpected changes in the output, so we can see backward compatibility issues.

PR #279, opened 2022-11-22
Update task image tag to match the ec-bundle tag

joejstuart:
This PR makes sure the task image version is updated on each merge to main


.github/workflows/release.yaml
```diff
@@ -82,7 +87,10 @@ jobs:
         run: podman login -u ${{ secrets.BUNDLE_PUSH_USER  }} -p ${{ secrets.BUNDLE_PUSH_PASS }} quay.io
 
       - name: Create image from snapshot
-        run: make push-snapshot-image IMAGE_TAG=${{ github.sha }}
+        run: make push-snapshot-image IMAGE_TAG=$TAG IMAGE_REPO=$IMAGE_REPO
+
+      - name: Create and push the tekton bundle tagged with sha
+        run: make task-bundle TASK_TAG=$TAG TASK=<( yq e '.spec.steps[].image? = "$IMAGE_REPO:$TAG"' $TASK)
```
lcarva, line 93: I think embedding an yq expression in this file will make it kind of hard to debug things. 

I'm also surprised that this works: `'.spec.steps[].image? = "$IMAGE_REPO:$TAG"'` 
I would've expected the variable replacement to not occur because of the outer single quotes.

.github/workflows/release.yaml
```diff
@@ -82,7 +87,10 @@ jobs:
...
 
-      - name: Create and push the tekton bundle
-        run: make task-bundle-snapshot TASK_TAG=${{ github.sha }}
+      - name: Create and push the tekton bundle tagged with snapshot
+        run: make task-bundle TASK_TAG=snapshot TASK=<( yq e '.spec.steps[].image? = "$IMAGE_REPO:snapshot"' $TASK)
```
lcarva, line 96: I don't really think building the bundle twice is a good idea. Let's build it once and tag that image as needed.

joejstuart: yeah, I was thinking making the bundle tag match the image tag, but I think having the image tag of "snapshot" is just a bad idea either way. I reverted this.

zregvart (approved):

zregvart:
Let's merge and see a release, we can followup if anything goes awry (I don't think it will).

PR #286, opened 2022-11-25
Development environment with Tekton Chains

zregvart:
Adds `hack/setup-dev-environment.sh` that will create a Kind cluster named `ec` if it already doesn't exist, install Tekton Pipielines, Tekton Chains, ECC, an image registry and setup a "work" namespace with the required permissions.

Also adds a `dev` target to the Makefile to push the ec and the task image to the development cluster.

Also includes: [Simplify the kustomize test setup](https://github.com/hacbs-contract/ec-cli/commit/8e8fad5ae5324bd8d97403c27fe77f6697dbb53a)

Replaces the use of `kubectl kustomize` with invoking the `kustomize` directly (version managed by `internal/tools/go.mod`). Instead of scripting in the `hack/setup-environment.sh` to determine the versions of Tekton and enterprise-contract-controller, that is now performed by running KRM functions via kustomize generators. This opens a way to install all required resources using a single kustomize command and a kustomization that references individual components.


zregvart:
Still want to test this a bit. I'm not too keen on the duplication of `cosign.pub` it would be nicer to have a single directory containing the public and private keys.

.fake-kube-config
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
lcarva, line 1: NIT: Let's add a comment in this file to explain its purpose.

Makefile
```diff
@@ -171,13 +171,25 @@ dist-image-push: dist-image  $(subst image_,push_image_,$(ALL_SUPPORTED_IMG_OS_A
 	@for img in $(ALL_IMAGE_REFS); do TARGETOS=$$(echo $$img | sed -e 's/.*:[^-]\+-\([^-]\+\).*/\1/'); TARGETARCH=$${img/*-}; podman manifest add $(IMAGE_REPO):$(IMAGE_TAG) $(PODMAN_OPTS) $$img --os $${TARGETOS} --arch $${TARGETARCH}; done
 	@podman manifest push $(IMAGE_REPO):$(IMAGE_TAG) $(IMAGE_REPO):$(IMAGE_TAG)
 
+.PHONY: dev
+dev: IMAGE_REPO=localhost:5000/ec
+dev: PODMAN_OPTS=--tls-verify=false
```
lcarva, line 176: I'm a perpetual Makefile noob. What does it mean when the target is repeated?

hack/chains/chains.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: v1
+kind: TektonResourcePlugin
```
lcarva, line 18: What's this?

hack/chains/cosign.key
```diff
@@ -0,0 +1,11 @@
+-----BEGIN ENCRYPTED COSIGN PRIVATE KEY-----
```
lcarva, line 1: I know this is for testing, but I'd feel a lot better if we didn't commit any private keys. Can we generate this on demand?

hack/ecc/ecc.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: v1
+kind: EnterpriseContractControllerResourcePlugin
```
lcarva, line 18: What's this?

hack/setup-test-environment.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
lcarva, line 1: This seems to duplicate a lot of `setup-dev-environment.sh`. Maybe on the next iteration we can combine the two?

lcarva (approved):
LGTM!

Anything I mentioned can be addressed in a follow up PR if needed.

Makefile
zregvart: It's just a [weird syntax](https://www.gnu.org/software/make/manual/make.html#Target_002dspecific) to set the variables per-target. It makes for better UX for the `dev` use case to have these preset

hack/chains/chains.yaml
zregvart: In kustomize one can have plugins, this one is called [exec KRM function](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/exec_krm_functions/), the `kind` and most other fields are ceremonial here, what's important is that we can now invoke a shell script and dynamically contribute objects (YAML files) to the set of objects Kustomize has defined statically. Kustomize is very intentionally anti-templating, so we can't point to a resource with `${version}` as a variable within its URL. Note that this won't work with `kubectl kustomize` as it explicitly doesn't include the exec plugins feature. What we arrive at with this is that we have a Kustomize-pure solution, making it easy to be incorporated in acceptance tests, i.e. we can use Kustomize's go API and point to a directory here and have the same environment provisioned from go in addition to shell script. Rather neat I would say.

hack/chains/cosign.key
zregvart: The need for a shared private key is so that the golden images and their attestations have a known key as signature. I've been trying to create images that we can use in https://quay.io/organization/hacbs-contract-demo/ that we could point ec to. I'd like for those to be usable and re-buildable (via `hack/rebuild.sh`) over time. For self-contained workflows like (acceptance) tests, we could generate keys, but I don't think it's worth it

hack/ecc/ecc.yaml
zregvart: See comment above -- ceremony to get the KRM exec plugin invoked from Kustomize

hack/setup-test-environment.sh
zregvart: Some of it is, there could be a `environment.sh` that has the shared bits in -- I didn't think it was worth it right now. This one uses a different Kustomize directory (development vs test) to install Chains in addition to Tekton and the image repository, doesn't build/push the images and waits on Chains in addition to Tekton & image repository to be available.

Makefile
zregvart: Oh, and I think if one don't have 50cm long gray beard you're not legally allowed to write Makefiles, so we're all Makefile noobs at this point.

hack/chains/cosign.key
lcarva: Can we store the private key as a secret here in GitHub?

joejstuart (approved):

hack/chains/cosign.key
zregvart: Hmm, we need to provide it to Chains running the HACBS stack, currently only the a dev cluster someone would spin up on their own is a viable option. With some investment we could run a near-HACBS stack in the development environment on Kind and then have it run on schedule in GitHub Actions to rebuild the golden images, though we're some ways off from that...

zregvart:
I'll merge this, the issue with the private key we can solve once we have the prerequisites in place: Kind environment that can run HACBS pipelines...

PR #304, opened 2022-12-07
Abstraction over kubernetes cluster type

zregvart:
In the acceptance tests we can now introduce different kinds of clusters using the `internal/acceptance/kubernetes/types.Cluster` abstraction. Currently the `stub` implementation using WireMock is provided, and this will be extended to include `kind`-based implementation.


internal/acceptance/kubernetes/kubernetes.go
```diff
@@ -14,95 +14,102 @@
...
 
-// AddStepsTo adds Gherkin steps to the godog ScenarioContext
+// AddStepsTo adds cluster-related steps to the context
 func AddStepsTo(sc *godog.ScenarioContext) {
-	sc.Step(`^stub apiserver running$`, stubApiserverRunning)
-	sc.Step(`^policy configuration named "([^"]*)" with specification$`, stubPolicy)
+	sc.Step(`^a stub cluster running$`, startAndSetupState(stub.Start))
+	sc.Step(`^policy configuration named "([^"]*)" with specification$`, createNamedPolicy)
+
+	// removes all git repositories from the filesystem
```
lcarva, line 97: This comment seems inaccurate

lcarva (approved):

internal/acceptance/kubernetes/kubernetes.go
zregvart: That's what you get for copy-pasting, thanks for spotting this!

PR #307, opened 2022-12-08
Use default auth when checking image access

lcarva:
This change ensures ec-cli can leverage any existing authentication from the underlying system when querying the registry to check image access.

An example where this is a must is when validating an image from an OpenShift internal registry which must always be authenticated.

### DIFF
--- a/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
+++ b/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path"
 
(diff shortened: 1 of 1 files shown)


lcarva:
Let's hold off merging this. Not sure if this fully solves the issue.

zregvart (approved):

zregvart:
> Let's hold off merging this. Not sure if this fully solves the issue.

Looks good though…

lcarva:
> Looks good though…

Yeah. I was running into another issue. (Bad policy). Let's merge this.

PR #311, opened 2022-12-12
Update k8s.io/client-go from 0.25.4 to 0.26.0 and fix sets

lcarva:
The new version of `k8s.io/client-go` deprecates the type specific sets in favor of generics.


zregvart:
The set changes are already on https://github.com/hacbs-contract/ec-cli/pull/310

lcarva:
> The set changes are already on #310

Awesome! Thanks for fixing it.

PR #316, opened 2022-12-19
Small, but important, track bundle fixes

lcarva:
* Keep oldest entry during deduplication
* Fix deduplication logic to take into account newly added entries
    
https://issues.redhat.com/browse/HACBS-1513

### DIFF
--- a/internal/tracker/tracker.go
+++ b/internal/tracker/tracker.go
@@ -214,7 +214,11 @@ func filterRecords(records []bundleRecord, prune bool) []bundleRecord {
 	filtered := make([]bundleRecord, 0, len(records))
 	keys := map[string]bool{}
 	for _, r := range records {
-		key := r.Repository + r.Digest
+		// NOTE: Newly added records will have a repository, but existing ones
+		// will not. This is expected because the output does not persist the
+		// repository for each record. Instead, the repository is the attribute
+		// which references the list of records.
+		key := r.Digest
 		if _, ok := keys[key]; ok {
 			continue
 		}
@@ -230,7 +234,7 @@ func filterRecords(records []bundleRecord, prune bool) []bundleRecord {
 	return filtered
 }
 
-// filterRecords reduces the list of required tasks by removing superfulous
+// filterRequiredTasks reduces the list of required tasks by removing superfulous
 // entries. If prune is true, it skips any entry that is no longer acceptable.
 // Any entry with an EffectiveOn date in the future, and the entry with the most
 // recent EffectiveOn date *not* in the future are considered acceptable.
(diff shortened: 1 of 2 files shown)


internal/tracker/tracker_test.go
```diff
@@ -481,224 +564,6 @@ func mustCreateFakeEmptyPipelineObject() runtime.Object {
 	return &pipeline
 }
 
-func TestFilterBundles(t *testing.T) {
```
lcarva, line 484: Removed this because I believe it is now covered by the newly added cases in TestTrack. I found this test a bit misleading because it didn't catch the issue where the `Repository` attribute is not set when loading data from YAML.

internal/tracker/tracker_test.go
```diff
@@ -306,6 +314,81 @@ required-tasks:
...
       - git-clone
       - summary
+`,
+		},
+		{
+			name: "prefer older entries with same digest",
+			urls: []string{
+				"registry.com/one:1.0@" + sampleHashOne.String(),
+			},
+			input: `---
```
lcarva, line 324: As a follow up, I'd like to apply heredoc to these. Skipped for this PR to minimize changes.

zregvart (approved):
I'd go with sort by `EffectiveOn` rather than assume that the YAML file would be only modified by `ec track bundle`, but that's just nitpickery. Let's merge this and see if we ever come across this...

internal/tracker/tracker.go
```diff
@@ -211,26 +211,34 @@ func (t *Tracker) filterBundles(prune bool) {
...
-		key := r.Repository + r.Digest
+	for i := len(records) - 1; i >= 0; i-- {
+		r := records[i]
+		// NOTE: Newly added records will have a repository, but existing ones
+		// will not. This is expected because the output does not persist the
+		// repository for each record. Instead, the repository is the attribute
+		// which references the list of records.
+		key := r.Digest
 		if _, ok := keys[key]; ok {
 			continue
```
joejstuart, line 224: Nit: Could get rid of the continue statement also to clean it up more.

```golang
for i := len(records) - 1; i >= 0; i-- {
	r := records[i]
	key := r.Digest
	if !keys[key] {
		keys[key] = true
                unique = append([]bundleRecord{r}, unique...)
	}
}
```

joejstuart (approved):

lcarva:
> I'd go with sort by `EffectiveOn` rather than assume that the YAML file would be only modified by `ec track bundle`, but that's just nitpickery. Let's merge this and see if we ever come across this...

@zregvart , we have a story to sort things in rego, https://issues.redhat.com/browse/HACBS-1507, but that won't help here. I'll take your suggestion to avoid having to sort the data here as well, at least for now.

PR #319, opened 2022-12-19
Support new format policy sources

simonbaird:
It required some changes to how evaluators are organized. There is now one evaluator for each "source group".

Also remove the "hard coded" data source since it's now possible to list it explicitly. This will require the adding the main EC data souce explicitly to default config in the e2e tests.

Reference: https://issues.redhat.com/browse/HACBS-1397


features/validate_image.feature
```diff
@@ -236,3 +244,112 @@ Feature: evaluate enterprise contract
       ]
     }
     """
+    #
+    # In this example the result is the same but in this example the there's only one "source
+    # group" which means the conftest evaluator is run just once with the three sources fetched
+    Given policy configuration named "ec-policy-variation" with specification
```
zregvart, line 250: Usually one wouldn't do multiple Given-When-Then sets within the same scenario, breaks the single responsibility and makes testing the individual cases harder. This would be a good case for using a [Rule](https://cucumber.io/docs/gherkin/reference/#rule) instead, and having the setup of the git repositories in a `Background`.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -101,33 +102,45 @@ func NewApplicationSnapshotImage(ctx context.Context, fs afero.Fs, url string, p
 		os.Setenv("SIGSTORE_TRUST_REKOR_API_PUBLIC_KEY", "1")
 	}
 
-	policySources, err := fetchPolicySources(p)
-	if err != nil {
-		log.Debug("Failed to fetch the policy sources from the ECP!")
-		return nil, err
-	}
+	// Return an evaluator for each of these
+	for _, sourceGroup := range p.EnterpriseContractPolicySpec.Sources {
```
zregvart, line 106: One improvement, in a followup, could be that the work within this loop is done via in parallel via gorutines

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -101,33 +102,45 @@ func NewApplicationSnapshotImage(ctx context.Context, fs afero.Fs, url string, p
...
 	return a, nil
 }
 
 // fetchPolicySources returns an array of policy sources
-func fetchPolicySources(spec *policy.Policy) ([]source.PolicySource, error) {
-	policySources := make([]source.PolicySource, 0, len(spec.Sources))
-	for _, sourceUrl := range spec.Sources {
-		url := source.PolicyUrl{Url: sourceUrl, Kind: source.PolicyKind}
+func fetchPolicySources(s *ecc.Source) ([]source.PolicySource, error) {
+	policySources := make([]source.PolicySource, 0, len(s.Policy)+len(s.Data))
```
zregvart, line 133: Log the name of the PolicySource being fetched?

zregvart (commented):
Needs to update [ecp-policy.yaml](https://github.com/hacbs-contract/ec-cli/blob/main/task/0.1/tests/ecp-policy.yaml), and for some reason [this](https://github.com/hacbs-contract/ec-cli/blob/main/hack/ecc/ecc.sh) resolved to the older version.

simonbaird:
Thanks @zregvart. I'll move it to draft until everything is passing and the comments are addressed.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
simonbaird: Actually I decided to take out the logging here since it's mostly duplicating the logging done in the calling function.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
simonbaird: I added a comment to suggest that.

hack/simple-demo.sh
```diff
@@ -25,7 +25,16 @@ EAg+paWtlme2KNcostCsmIbwz+bc2aFV+AxCOpRjRpp3vYrbS5KhkmgC1Q==
 POLICY='{
   "publicKey": "'${PUBLIC_KEY//$'\n'/\\n}'",
   "sources": [
-    "github.com/hacbs-contract/ec-policies//policy"
+    {
+      "name": "EC Policies",
+      "policy": [
+        "github.com/hacbs-contract/ec-policies//policy/lib",
```
lcarva, line 31: Oh interesting! Does this actually work??

lcarva (approved):

hack/simple-demo.sh
simonbaird: Yeah, I wanted to see if it worked like this, and seems like it does.

PR #327, opened 2023-01-02
Move tools an acceptance to repository root

zregvart:
This sets up the acceptance tests as a separate go module and moves acceptance and tools from `internal` to repository root to better reflect them as modules.


lcarva:
This seems pretty straight forward. If using go 1.18 or newer, it doesn't seem to require any special usage. It just works.

Are there any downsides to this approach? I wonder if there will be any issues with dependabot.

zregvart:
> Are there any downsides to this approach? I wonder if there will be any issues with dependabot.

Seems that dependabot [doesn't support workspaces](https://github.com/dependabot/dependabot-core/issues/6012) right now, so that's a downside -- I've seen `go mod tidy` in a module update the `go.work.sum` file.

Other than that, I was hoping that we could work with dependencies a bit more independent, but that is not the case, so having modules without workspace could be a bit more helpful at the moment. For example, the acceptance tests could run with a different version of docker go API than the CLI, which would help with Testcontainers update (#325). Downside of that, is that I don't think it's good for the projects to drift in versions, an example of that would be tekton/cli vs tekton/pipeline version issue that occurred in #324.

I haven't seen anything that would impede the day to day workflow. We don't _really_ need workspaces, the modules don't depend on each other. An interesting development could be if we were to merge the https://github.com/hacbs-contract/enterprise-contract-controller and https://github.com/hacbs-contract/ec-cli into a single git repository, here workspace would be very beneficial, so if we would like to consider that, this would be a good first step.

simonbaird:
I can see the "move acceptance tests to their own directory with their own go.mod file" change, but what part of this change is the "create go workspace"?

zregvart:
> but what part of this change is the "create go workspace"?

In essence the `go.work` and `go.work.sum` files (https://github.com/hacbs-contract/ec-cli/pull/327/files#diff-70f7e3c361ae136c655130ab253927b5f6e5c987797a095db1bbb1e790930484). The repercussions are in the `go.mod` files. With `go work sync` the dependencies get hoisted (or that's my understanding) from modules and then the (newest?) version gets synced down to other modules -- resulting in some version changes that I did not expect when starting on this.

We don't inter-depend between modules so no other changes are visible. If we were, the `replace` in `go.mod` that would point to a local directory would be removed.

simonbaird:
Aha, thanks.

lcarva:
Have we decided to drop this, at least for now?

zregvart:
> Have we decided to drop this, at least for now?

I'll remove the `go.work` and `go.work.sum` files & update the title/commit message.

.github/workflows/checks.yaml
```diff
@@ -43,7 +43,6 @@ jobs:
       - name: Download go dependencies
         run: |
           go mod download
-          go mod download -modfile internal/tools/go.mod
```
lcarva, line 46: For my own understanding, why is this removed instead of changed to the new path?

lcarva (approved):

lcarva (approved):

PR #328, opened 2023-01-03
Print make target

zregvart:
Adds printing of make target to add a clue as to what target is running. This is as helpful as it can be made, some targets are run in parallel which obfuscates the output a bit.


lcarva (approved):
I'll trust you on this!

zregvart:
Easy to revert, let's try and see if folk find it valuable...

PR #332, opened 2023-01-05
Consolidate attestations into single input file

lcarva:
When given multiple input files, the conftest evaluator processes each input file separately. This causes issues when evaluating policy rules that expect the list of attestations to contain certain attestation types.

A single input file allows policy rules to make a decision based on full set of attestations available.

Also, since ec-cli already consolidates the output, it makes sense to consolidate the input as well.

https://issues.redhat.com/browse/HACBS-1573

### DIFF
--- a/internal/evaluation_target/application_snapshot_image/application_snapshot_image_test.go
+++ b/internal/evaluation_target/application_snapshot_image/application_snapshot_image_test.go
@@ -139,43 +139,84 @@ func createSimpleAttestation(statement *in_toto.Statement) oci.Signature {
 	return signature
 }
 
-func TestWriteInputFiles(t *testing.T) {
+const simpleAttestationJSONText = `{
+	"_type": "",
+	"predicateType": "https://slsa.dev/provenance/v0.2",
+	"subject": null,
+	"predicate": {
+		"buildType": "https://tekton.dev/attestations/chains/pipelinerun@v2",
+		"builder": {
+			"id": ""
+		},
+		"invocation": {
+			"configSource": {}
+		}
+	}
+}`
+
+func TestWriteInputFile(t *testing.T) {
+	cases := []struct {
+		name     string
+		snapshot ApplicationSnapshotImage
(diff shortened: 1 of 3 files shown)


internal/image/validate.go
```diff
@@ -109,7 +109,7 @@ func ValidateImage(ctx context.Context, fs afero.Fs, url string, p *policy.Polic
 		return out, nil
 	}
 
-	inputs, err := a.WriteInputFiles(ctx, fs)
+	inputs, err := a.WriteInputFile(ctx, fs)
```
simonbaird, line 112: Since it went from "Files" to "File", maybe change the var name to match.

```suggestion
	input, err := a.WriteInputFile(ctx, fs)
```

internal/evaluation_target/application_snapshot_image/application_snapshot_image_test.go
```diff
@@ -139,43 +139,84 @@ func createSimpleAttestation(statement *in_toto.Statement) oci.Signature {
...
+  "predicate": {
+    "buildType": "https://tekton.dev/attestations/chains/pipelinerun@v2",
+    "builder": {
+    "id": ""
+    },
+    "invocation": {
+    "configSource": {}
+    }
+  }
+}`
```
simonbaird, line 155: I'm bike shedding here, but it feels odd to have the two space indent here. Would tabs work?

simonbaird (approved):

internal/image/validate.go
lcarva: oops missed this one. Fixed. Thanks!

internal/evaluation_target/application_snapshot_image/application_snapshot_image_test.go
lcarva: That's a good point. It's JSON so it doesn't matter. Switched to tabs to stay consistent with the rest of the file.

PR #337, opened 2023-01-09
update PIPELINE_BUNDLE in rebuild.sh and ECP to use new source schema

cuipinghuo:
https://issues.redhat.com/browse/HACBS-1587
https://issues.redhat.com/browse/HACBS-1588

update PIPELINE_BUNDLE in rebiuld.sh to use 
PIPELINE_BUNDLE=quay.io/redhat-appstudio-tekton-catalog/pipeline-hacbs-docker-build:devel

update hack/simple-demo.sh to use ec-golden-image:latest 

update EnterpriseContractPolicy with new source schema

### DIFF
--- a/hack/demo.sh
+++ b/hack/demo.sh
@@ -45,14 +45,12 @@ metadata:
   name: ec-demo
 spec:
   description: Demo Enterprise Contract policy configuration
-  configuration:
-    exclude:
-    - not_useful
-    - test:conftest-clair
   sources:
-    - name: Default EC policy
-      policy:
-      - quay.io/hacbs-contract/ec-release-policy:latest
+  - data:
+    - quay.io/hacbs-contract/ec-policy-data:latest
+    name: Default EC policy
+    policy:
+    - quay.io/hacbs-contract/ec-release-policy:latest
 EOF
 
 while read -r IMG
--- a/task/0.1/tests/ecp-policy.yaml
+++ b/task/0.1/tests/ecp-policy.yaml
@@ -22,9 +22,8 @@ metadata:
 spec:
   description: My custom enterprise contract policy configuration
   sources:
-    - policy:
-      - quay.io/hacbs-contract/ec-release-policy:latest
-  configuration:
-    exclude:
-      - not_useful
-      - test:conftest-clair
+  - data:
+    - quay.io/hacbs-contract/ec-policy-data:latest
+    name: ec-policy
+    policy:
+    - quay.io/hacbs-contract/ec-release-policy:latest
--- a/hack/rebuild.sh
+++ b/hack/rebuild.sh
@@ -29,7 +29,7 @@ HACK_DIR="$(dirname "${BASH_SOURCE[0]}")"
 PIPELINE_SERVICE_ACCOUNT=pipeline
 
 # What pipeline bundle to use
-PIPELINE_BUNDLE=quay.io/redhat-appstudio/hacbs-templates-bundle:devel
+PIPELINE_BUNDLE=quay.io/redhat-appstudio-tekton-catalog/pipeline-hacbs-docker-build:devel
 
 # Where to push the image(s)
 IMAGE_REPOSITORY=quay.io/hacbs-contract-demo
--- a/hack/simple-demo.sh
+++ b/hack/simple-demo.sh
@@ -15,12 +15,15 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-IMAGE=${IMAGE:-"quay.io/cuipinghuo/single-container-app:9f5d549dd64aacf10e3baac90972dfd5df788324"}
+IMAGE=${IMAGE:-"quay.io/redhat-appstudio/ec-golden-image:latest"}
 
+#ec-golden-image is signed with staging public key, to verify, use the below public key
+#(https://raw.githubusercontent.com/redhat-appstudio/infra-deployments/main/components/pipeline-service/public/tekton-chains-signing-secret.pub)
 PUBLIC_KEY=${PUBLIC_KEY:-"-----BEGIN PUBLIC KEY-----
-MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEPfwkY/ru2JRd6FSqIp7lT3gzjaEC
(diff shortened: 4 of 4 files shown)


lcarva (approved):

hack/demo.sh
```diff
@@ -48,16 +48,21 @@ spec:
...
   sources:
-    - name: Default EC policy
-      policy:
-      - quay.io/hacbs-contract/ec-release-policy:latest
+  - data:
+    - github.com/hacbs-contract/ec-policies.git//data
+    name: ec-demo
+    policy:
+      - github.com/hacbs-contract/ec-policies.git//policy/lib
+      - github.com/hacbs-contract/ec-policies.git//policy/release
```
simonbaird, line 59: I think it's better to keep using the quay.io url. That's what we're moving towards in other places. That bundle includes the lib dir, so you only need the one url.

hack/demo.sh
```diff
@@ -48,16 +48,21 @@ spec:
   configuration:
     exclude:
     - not_useful
-    - test:conftest-clair
+    include:
+    - slsa_provenance_available
```
simonbaird, line 52: Specifying one include like this means use *only* that item. Is that what you're intending here?

simonbaird (changes requested):

hack/demo.sh
cuipinghuo: I miss understand the include here, so deleting it now.

hack/demo.sh
cuipinghuo: change back to quay.io/hacbs-contract/ec-release-policy:latest

hack/demo.sh
```diff
@@ -48,10 +48,11 @@ spec:
   configuration:
     exclude:
     - not_useful
```
zregvart, line 50: We removed `not_useful` (see https://github.com/hacbs-contract/ec-policies/pull/236), perhaps it's a good opportunity to remove it here as well

task/0.1/tests/ecp-policy.yaml
```diff
@@ -21,10 +21,12 @@ metadata:
   name: ec-policy
 spec:
   description: My custom enterprise contract policy configuration
-  sources:
-    - policy:
-      - quay.io/hacbs-contract/ec-release-policy:latest
   configuration:
     exclude:
       - not_useful
```
zregvart, line 26: Let's remove it, we should not need it any more
```suggestion
```

hack/demo.sh
```diff
@@ -48,10 +48,11 @@ spec:
...
-    - test:conftest-clair
```
zregvart, line 51: We removed `not_useful` (see https://github.com/hacbs-contract/ec-policies/pull/236), perhaps it's a good opportunity to remove it here as well

```suggestion
```

zregvart (approved):
Thanks!

cuipinghuo:
@zregvart Could you help to merge?

hack/demo.sh
```diff
@@ -48,10 +48,11 @@ spec:
...
   sources:
-    - name: Default EC policy
-      policy:
+  - data:
+    - github.com/hacbs-contract/ec-policies.git//data
```
simonbaird, line 53: It's pretty new, but you could consider using the bundle for this too, i.e.:
```suggestion
    - quay.io/hacbs-contract/ec-policy-data:latest
```

hack/demo.sh
simonbaird, line 50: Repeating Zoran's suggestion, the `not_useful` doesn't exist any more, so no need to exclude it. Not sure about conftest-clair. I remember it used to regularly fail due to a time out. Does it pass more often now?

hack/demo.sh
simonbaird, line 53: This should work now if you want to use the bundle here also.

```suggestion
    - quay.io/hacbs-contract/ec-policy-data:latest
```

simonbaird (approved):
Lgtm, but see my comments/questions.

hack/demo.sh
simonbaird: (Delete my other comment about yaml indenting. It's already inconsistent I think, so don't worry it.)

zregvart:
Saw the [error uploading to codecov](https://github.com/hacbs-contract/ec-cli/actions/runs/3877962552/jobs/6621314563#step:6:30) so I restarted the CLI job. Even though we should be save to merge this by ignoring the codecov issue.

@cuipinghuo do you want to address the `not_useful` removal, or should we put that off for a followup PR?

zregvart (approved):

zregvart:
I think this should be ready to merge, let's just double-check that the Checks/Task doesn't fail for some odd reason...

task/0.1/tests/ecp-policy.yaml
cuipinghuo: Thanks,removed.

hack/demo.sh
cuipinghuo: Changed to quay.io/hacbs-contract/ec-policy-data:latest.And the bundle works. Thanks!

cuipinghuo:
> merge

cuipinghuo:
> I think this should be ready to merge, let's just double-check that the Checks/Task doesn't fail for some odd reason...



> Checks



> Saw the [error uploading to codecov](https://github.com/hacbs-contract/ec-cli/actions/runs/3877962552/jobs/6621314563#step:6:30) so I restarted the CLI job. Even though we should be save to merge this by ignoring the codecov issue.
> 
> @cuipinghuo do you want to address the `not_useful` removal, or should we put that off for a followup PR?

@zregvart I removed it, PTAL.

hack/demo.sh
cuipinghuo: I didn't see conftest-clair fails now, to see the latest ec validation result, run [simple-demo.sh](https://github.com/hacbs-contract/ec-cli/blob/main/hack/simple-demo.sh) which is with [quay.io/redhat-appstudio/ec-golden-image:latest](https://quay.io/repository/redhat-appstudio/ec-golden-image?tab=tags&tag=latest])

zregvart: I think conftest-clair was problematic in the past, timeouts against the API and such. IIRC the test stream now runs with a standalone clair tool, so that improved the reliability.

PR #339, opened 2023-01-10
Reverts to using the upstream release action

zregvart:
We needed to fork the `softprops/action-gh-release` to get fixes merged. These have now been addressed upstream, so we no longer need to use our fork.


lcarva (approved):

zregvart:
Hopefully nothing is broken :crossed_fingers:, as we can only see that once the release workflow runs. Easy to revert if it is...

PR #370, opened 2023-01-27
Update staging public key in simple demo script

simonbaird:


lcarva (approved):
Verified it matches: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/public/tekton-chains-signing-secret.pub

Hey, it might be a good idea to just curl that. Wdyt?

zregvart (approved):

PR #373, opened 2023-01-31
Add ec inspect policy command

simonbaird:
Still very light on test coverage but perhaps we could merge early and often.

Cleaned up and simplified from original version.

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -20,19 +20,16 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func fetchCmd() *cobra.Command {
+func inspectCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "fetch",
-		Short: "Fetch policies and authorization data",
+		Use:   "inspect",
+		Short: "Inspect policy rules",
 	}
-
 	return cmd
 }
 
 func init() {
-	fetch := fetchCmd()
-	fetch.AddCommand(commitAuthorizationCmd())
-	fetch.AddCommand(k8sResourceAuthorizationCmd())
-	fetch.AddCommand(fetchPolicyCmd())
-	RootCmd.AddCommand(fetch)
+	inspect := inspectCmd()
+	inspect.AddCommand(inspectPolicyCmd())
+	RootCmd.AddCommand(inspect)
 }
--- a/cmd/inspect_policy.go
+++ b/cmd/inspect_policy.go
@@ -1,0 +1,118 @@
+// Copyright 2022 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Define the `ec inspect policy` command
+package cmd
+
+import (
+	"encoding/json"
+
+	hd "github.com/MakeNowJust/heredoc"
+	"github.com/open-policy-agent/opa/ast"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+
+	"github.com/hacbs-contract/ec-cli/internal/opa"
+	"github.com/hacbs-contract/ec-cli/internal/policy/source"
+	"github.com/hacbs-contract/ec-cli/internal/utils"
+)
+
+func inspectPolicyCmd() *cobra.Command {
+	var (
+		sourceUrls   []string
+		destDir      string
+		outputFormat string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "policy --source <source-url>",
+		Short: "Read policies from source urls and show information about the rules inside them",
+
+		Long: hd.Doc(`
+			Read policies from a source url and show information about the rules inside them.
+
+			This fetches policy sources similar to the 'ec fetch policy' command, but once
+			the policy is fetched the equivalent of 'opa inspect' is run against the
+			downloaded policies.
+
+			This can be used to extract information about each rule in the policy source,
+			including the rule annotations which include the rule's title and description
+			and custom fields used by ec to filter the results produced by conftest.
+
+			Note that this command is not typically required to verify the Enterprise
+			Contract. It has been made available for troubleshooting and debugging purposes.
+		`),
+
+		Example: hd.Doc(`
+			Print a list of rules and their descriptions from the latest Stonesoup release policy:
+
+			  ec inspect policy --source quay.io/hacbs-contract/ec-release-policy
+
+			Display details about the latest Stonesoup release policy in json format:
+
+			  ec inspect policy --source quay.io/hacbs-contract/ec-release-policy -o json | jq
+		`),
+
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if destDir == "" {
+				workDir, err := utils.CreateWorkDir(afero.NewOsFs())
+				if err != nil {
+					log.Debug("Failed to create work dir!")
+					return err
+				}
+				destDir = workDir
+			}
+
+			allResults := make(map[string][]*ast.AnnotationsRef)
+			for _, url := range sourceUrls {
+				s := &source.PolicyUrl{Url: url, Kind: source.PolicyKind}
+
+				// Download
+				policyDir, err := s.GetPolicy(cmd.Context(), destDir, false)
+				if err != nil {
+					return err
+				}
+
+				// Inspect
+				result, err := opa.InspectDir(afero.NewOsFs(), policyDir)
+				if err != nil {
+					return err
+				}
+
+				// Collect results
+				allResults[s.PolicyUrl()] = result
+			}
+
+			out := cmd.OutOrStdout()
+			if outputFormat == "json" {
+				return json.NewEncoder(out).Encode(allResults)
+			}
+			return opa.OutputText(out, allResults)
+		},
+	}
+
+	cmd.Flags().StringArrayVarP(&sourceUrls, "source", "s", []string{}, "policy source url. multiple values are allowed")
+	cmd.Flags().StringVarP(&destDir, "dest", "d", "", "use the specified destination directory to download the policy. if not set, a temporary directory will be used")
+	cmd.Flags().StringVarP(&outputFormat, "output", "o", "text", "output format, either json or text.")
+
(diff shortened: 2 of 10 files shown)


simonbaird:
Example usage:

```
$ go run main.go inspect-policy --source github.com/hacbs-contract/ec-policies/policy --work-dir -o yaml | head
github.com/hacbs-contract/ec-policies/policy:
- annotations:
    custom:
      failure_msg: Unexpected kind '%s'
      short_name: unexpected_kind
    description: A sanity check to confirm the input data has the kind "Pipeline"
    scope: rule
    title: Input data has unexpected kind
  location:
    col: 1
```

internal/opa_inspect/opa_inspect.go
```diff
@@ -0,0 +1,151 @@
...
+func inspectSingle(path, module string) ([]*ast.AnnotationsRef, error) {
+	mod, err := ast.ParseModuleWithOpts(path, module, ast.ParserOptions{ProcessAnnotation: true})
+	if err != nil {
+		return nil, err
+	}
+
+	as, x := ast.BuildAnnotationSet([]*ast.Module{mod})
+	if len(x) > 0 {
+		return nil, err
+	}
```
zregvart, line 39: I wrote this, now looking at it it looks dodgy, `err` is `nil` here, `x` should be named `errs`, this should be (from the top of my head), something like:

```suggestion
	as, errs := ast.BuildAnnotationSet([]*ast.Module{mod})
	if len(errs) > 0 {
		return nil, errors.New(errs.Error())
	}
```

internal/opa_inspect/opa_inspect.go
```diff
@@ -0,0 +1,151 @@
...
+		results = append(results, r...)
+	}
+
+	return results, nil
+}
+
+func isWarnOrDeny(rule *ast.AnnotationsRef) bool {
+	ruleName := rule.Path[len(rule.Path)-1].Value.String()
+	// Fixme: No idea how to compare the ast.Value to a string in a sane
+	// way here. https://pkg.go.dev/github.com/open-policy-agent/opa/ast#Value
```
zregvart, line 71: Looking at the docs, does this work? Note lots of pointers in this:

```suggestion
	ruleName := rule.GetRule().Head.Name
```

internal/opa_inspect/opa_inspect.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	return filteredResults, nil
+}
+
+func shortPath(fullPath string, destDir string) string {
+	return strings.TrimPrefix(strings.TrimPrefix(fullPath, destDir), "/")
+}
+
+func InspectDir(dir string, raw bool) ([]*ast.AnnotationsRef, error) {
```
zregvart, line 97: I think we'd prefer using afero here

internal/policy/source/source.go
```diff
@@ -46,6 +46,7 @@ type downloaderFunc interface {
 type PolicySource interface {
 	GetPolicy(ctx context.Context, dest string, showMsg bool) (string, error)
 	PolicyUrl() string
+	PolicyDir() string
```
zregvart, line 49: Not sure about this addition, the PolicyDir is a concern of the policy engine, not of the policy source

simonbaird:
Thanks Zoran. I didn't go through your suggestions yet, will do that soon.

cmd/fetch.go
```diff
@@ -23,7 +23,7 @@ import (
 func fetchCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "fetch",
-		Short: "Fetch data for policy evaluation",
+		Short: "Fetch something",
```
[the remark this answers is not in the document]
simonbaird: Done.

cmd/inspect.go
```diff
@@ -0,0 +1,36 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
```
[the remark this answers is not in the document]
simonbaird: I think it's a good suggestion, but I'd rather stick with the current convention in this PR, and maybe do a reorg of all the cmd files in a future PR.

cmd/inspect_policy.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+
+			fmt.Print(string(output))
+			return nil
+		},
+	}
+
+	cmd.Flags().StringArrayVarP(&sourceUrls, "source", "s", []string{}, "policy source url. multiple values are allowed")
+	cmd.Flags().StringVarP(&destDir, "dest", "d", ".", "use the specified download destination directory. ignored if --work-dir is set")
+	cmd.Flags().BoolVarP(&useWorkDir, "work-dir", "w", false, "use a temporary work dir as the download destination directory")
```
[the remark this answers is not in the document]
simonbaird: Initially I was being lazy, but I think it is useful to use a tmp dir. Maybe I'll make that the default if dest is not specified and remove the flag.

cmd/inspect_policy.go
```diff
@@ -0,0 +1,150 @@
...
+				}
+				output = []byte(outputStr)
+			}
+
+			fmt.Print(string(output))
+			return nil
+		},
+	}
+
+	cmd.Flags().StringArrayVarP(&sourceUrls, "source", "s", []string{}, "policy source url. multiple values are allowed")
```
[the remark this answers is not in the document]
simonbaird: That's a good idea. In the interests of merging this one sooner, I think it can be a new story.

cmd/inspect_policy.go
```diff
@@ -0,0 +1,150 @@
...
+
+			var (
+				output []byte
+				err    error
+			)
+			switch outputFormat {
+			case "json":
+				if output, err = json.Marshal(allResults); err != nil {
+					return err
+				}
```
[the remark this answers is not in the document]
simonbaird: Changed the json, but I couldn't make it work like that for the yaml. Added a todo about it.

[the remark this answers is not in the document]
simonbaird: Decided to remove the yaml entirely..

cmd/inspect_policy.go
```diff
@@ -0,0 +1,150 @@
...
+
+			fmt.Print(string(output))
+			return nil
+		},
+	}
+
+	cmd.Flags().StringArrayVarP(&sourceUrls, "source", "s", []string{}, "policy source url. multiple values are allowed")
+	cmd.Flags().StringVarP(&destDir, "dest", "d", ".", "use the specified download destination directory. ignored if --work-dir is set")
+	cmd.Flags().BoolVarP(&useWorkDir, "work-dir", "w", false, "use a temporary work dir as the download destination directory")
+	cmd.Flags().BoolVarP(&raw, "raw", "r", false, "return annotations for every rule instead of just warns and denies")
```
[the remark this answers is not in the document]
simonbaird: Done.

internal/policy/source/source.go
simonbaird: Yeah, that's a reasonable point. It's quite convenient to stash it there though. Anyway, since it's not actually needed yet, I took it out of this PR.

simonbaird:
I'm still being lazy about the tests, but I think I'll pull it out of draft. Recent changes:

- Move the "inspect output to text" code to its own package (`internal/rego_text`).
- Remove the yaml output option since json is fine, and I was having trouble making it work as I wanted. Also you can use `yq -P` very easily.
- Simplify the command flags based on Zoran's suggestion
- Remove the commit about storing the dest dir for each policy source url.
- Misc tweaks.

simonbaird:
> have an intermediate struct/array in opa_inspect

I think it's probably a good idea, but I think it can be done in future. Once we try to integrate this data with the `ec validate` it might be more clear what is needed.

internal/opa_inspect/opa_inspect.go
simonbaird: Fixed.

internal/opa_inspect/opa_inspect.go
simonbaird: Still todo.

cmd/inspect.go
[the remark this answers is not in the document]
zregvart: Created [HACBS-1736](https://issues.redhat.com/browse/HACBS-1736)

internal/opa_inspect/opa_inspect.go
```diff
@@ -0,0 +1,151 @@
...
+	return results, nil
+}
+
+func isWarnOrDeny(rule *ast.AnnotationsRef) bool {
+	r := rule.GetRule()
+	if r == nil {
+		return false
+	}
+	ruleName := r.Head.Name.String()
+	return ruleName == "deny" || ruleName == "warn"
```
lcarva, line 77: Should we use the same convention used by conftest here? e.g. `deny`, `deny_*`

internal/opa_inspect/opa_inspect.go
```diff
@@ -0,0 +1,151 @@
...
+func hasAnnotations(rule *ast.AnnotationsRef) bool {
+	return rule.Annotations != nil
+}
+
+// Interesting rules are warns, denies, and anything with an annotation,
+// which includes some package scoped annotations not assocated with a rule
+func interestingRulesOnly(results []*ast.AnnotationsRef) ([]*ast.AnnotationsRef, error) {
+	filteredResults := make([]*ast.AnnotationsRef, 0, len(results))
+	for _, rule := range results {
+		if isWarnOrDeny(rule) || hasAnnotations(rule) {
```
lcarva, line 89: It would be useful to document this behavior in a more "public" place. I did not expect rules without annotations to be excluded but I agree that it does make sense to do so.

lcarva (approved):
Cool!

lcarva:
FYI, go prescribes a somewhat unique way to name packages: https://go.dev/blog/package-names

Something like `opa_inspect` could be called just `opa`. This makes the usage side of things a bit nicer: `opa.InspectDir`.

internal/opa_inspect/opa_inspect.go
zregvart: [This](https://github.com/open-policy-agent/conftest/blob/55834493ce76e76241479ede4e3c711010a7bf5c/policy/engine.go#L490-L498) is what conftest does

simonbaird:
> FYI, go prescribes a somewhat unique way to name packages: https://go.dev/blog/package-names
> 
> Something like `opa_inspect` could be called just `opa`. This makes the usage side of things a bit nicer: `opa.InspectDir`.

Thanks. I did move things around a little based on that:

```
internal/opa/inspect.go
internal/opa/output.go
```

internal/opa_inspect/opa_inspect.go
simonbaird: The presence of the --raw flag did provide some clue about it before.

But note: Warns/denies without annotations would show in the json. And in the very latest revision (a short time ago), the also now appear in the text output. The things I'm mainly excluding are the `lib.useful_thing` and the `_some_helper` type of rules.

PR #375, opened 2023-01-31
Required pipeline tasks

joejstuart:
Add new key to the acceptable bundles list `pipeline-required-tasks`  This key will contain the tasks that exist in each pipeline definition found in a bundle. 
The existing key `required-tasks` will be kept and used as the default.


features/track_bundle.feature
```diff
@@ -19,6 +19,11 @@ Feature: track bundles
         - digest: ${REGISTRY_acceptance/bundle:tag_HASH}
           effective_on: "[0-9]{4}-[0-9]{2}-[0-9]{2}T00:00:00Z"
           tag: tag
+    pipeline-required-tasks:
+      pipeline1:
+       effective_on: "[0-9]{4}-[0-9]{2}-[0-9]{2}T00:00:00Z"
+       tasks:
+         - git-clone
```
zregvart, line 26: Seems that the indenting is causing acceptance test failure
```suggestion
        effective_on: "[0-9]{4}-[0-9]{2}-[0-9]{2}T00:00:00Z"
        tasks:
          - git-clone
```

internal/tracker/bundle_info.go
```diff
@@ -71,31 +72,56 @@ func newBundleInfo(ctx context.Context, ref image.ImageReference, tasks commonTa
 	return &info, nil
 }
 
-// updateCommonTasks updates the commonPipelineTasks attributes with the tasks found
-// in the given pipeline.
-func (info *bundleInfo) updateCommonTasks(ctx context.Context, pipelineName string) error {
+// getBundlePipeline gets a pipeline from a bundle by the pipeline name
+func (info *bundleInfo) getBundlePipeline(ctx context.Context, pipelineName string) (error, v1beta1.PipelineObject) {
```
zregvart, line 76: Odd to see it the other way around

```suggestion
func (info *bundleInfo) getBundlePipeline(ctx context.Context, pipelineName string) (v1beta1.PipelineObject, error) {
```

internal/tracker/bundle_info.go
```diff
@@ -71,31 +72,56 @@ func newBundleInfo(ctx context.Context, ref image.ImageReference, tasks commonTa
...
+	if taskKey := getRequiredTaskKey(pipelineObject); taskKey != "" {
+		info.pipelineTasks = map[string][]string{taskKey: sets.List(pipelineTaskNames)}
+	}
+
+	info.commonPipelineTasks.updateTasksIntersection(pipelineTaskNames)
 	return nil
 }
 
+// get the key used in "pipeline-required-tasks" from the pipelineObject
+func getRequiredTaskKey(pipelineObject v1beta1.PipelineObject) string {
```
zregvart, line 115: Perhaps this should return a `bool` and be named `holdsRequiredTasks`?

zregvart (approved):
Looks good, some nitpickery inside. Should we, perhaps in a followup, use a different annotation/label, I don't fully understand what `pipelines.openshift.io/strategy` is for and I could not find any documentation of it.

internal/tracker/bundle_info.go
joejstuart: So just check for the existence of the key? Where does it get the label value then?

internal/tracker/bundle_info.go
joejstuart: Yes, good catch. fixed

internal/tracker/bundle_info.go
zregvart: I could have missed other uses of `getRequiredTaskKey`; I was thinking about the way the function is used on line 106, basically as a conditional for selecting pipeline tasks, so reflecting that in the name/signature would be helpful.

PR #381, opened 2023-02-02
Make policy.Policy an interface

zregvart:
Having a globally mutable `policy.Policy` makes changes to it difficult, having to keep track of what modifies internal state or how individual `policy.Policy` are constructed. This makes `policy.Policy` an interface to ease the maintenance.

### DIFF
--- a/internal/image/validate_test.go
+++ b/internal/image/validate_test.go
@@ -22,13 +22,11 @@ import (
 	"encoding/json"
 	"errors"
 	"testing"
-	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	gcr "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/types"
-	"github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
 	"github.com/in-toto/in-toto-golang/in_toto"
 	v02 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
 	conftestOutput "github.com/open-policy-agent/conftest/output"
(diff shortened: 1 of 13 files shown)


internal/image/validate_test.go
```diff
@@ -116,11 +114,8 @@ func TestValidateImage(t *testing.T) {
 		t.Run(c.name, func(t *testing.T) {
 			fs := afero.NewMemMapFs()
 
-			p := &policy.Policy{
-				EnterpriseContractPolicySpec: v1alpha1.EnterpriseContractPolicySpec{},
-				CheckOpts:                    &cosign.CheckOpts{},
-				EffectiveTime:                time.Now().UTC(),
-			}
+			p, err := policy.NewOfflinePolicy(context.TODO(), "")
```
lcarva, line 117: There's a context a couple of lines below. Maybe do this after and use that context?

internal/policy/policy.go
```diff
@@ -36,21 +36,54 @@ import (
...
+type Policy interface {
+	PublicKeyPEM() ([]byte, error)
+	CheckOpts() (*cosign.CheckOpts, error)
+	WithSpec(spec ecc.EnterpriseContractPolicySpec) Policy
+	Spec() ecc.EnterpriseContractPolicySpec
+	EffectiveTime() time.Time
+}
+
+type policy struct {
 	ecc.EnterpriseContractPolicySpec
```
lcarva, line 48: Should this be `spec ecc.EnterpriseContractPolicySpec` now? Not sure if there's any value in "extending" that type.

internal/policy/policy.go
```diff
@@ -36,21 +36,54 @@ import (
...
+	pk, err := p.checkOpts.SigVerifier.PublicKey()
 	if err != nil {
 		return []byte{}, err
 	}
 	return cryptoutils.MarshalPublicKeyToPEM(pk)
 }
 
+func (p *policy) CheckOpts() (*cosign.CheckOpts, error) {
+	if p.checkOpts == nil {
+		return nil, errors.New("no check options configured")
```
lcarva, line 64: This seems a bit unusual. Is returning an error the right thing here?

lcarva (approved):
Neat!

internal/policy/policy.go
zregvart: Either that or a panic, `p.checkOpts` might not get set...

internal/policy/policy.go
zregvart: We break backward compatibility when `--policy` is an inline policy otherwise; less we copy all the fields from `EnterpriseContractPolicySpec` into `policy`.

lcarva: Yeah... We could add something like `NewFromECPSpec`. Probably not worth doing it at this time.

lcarva (approved):

simonbaird (approved):

PR #383, opened 2023-02-03
Cleanup workdir when no longer needed

zregvart:
Let's not fill up the temporary disk space with `ec-work-*` directories. There is a feature toggle built in to help debugging, if `EC_DEBUG` environment variable is set, the temp working directory will not be deleted.


lcarva (approved):
It would be great to document the `EC_DEBUG` env var somewhere (README.md?)

simonbaird (approved):

zregvart:
Added note to the README, merging now...

PR #391, opened 2023-02-07
Effective time refinements

zregvart:
Effective time can now be specified with the following designators:
 - "now" for current time (default)
 - "attestation" for the youngest attestation time, determined from `predicate.metadata.buildFinishedOn`
 - date as a RFC3339 string (2001-02-03T04:05:06+07:00)
 - only date (2001-02-03)

Ref. https://issues.redhat.com/browse/HACBS-1332

### DIFF
--- a/acceptance/image/image.go
+++ b/acceptance/image/image.go
@@ -25,6 +25,7 @@ import (
 	"io"
 
 	"github.com/cucumber/godog"
+	jsonpatch "github.com/evanphx/json-patch/v5"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
@@ -32,6 +33,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/random"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/sigstore/cosign/pkg/cosign"
 	"github.com/sigstore/cosign/pkg/oci/static"
 	cosigntypes "github.com/sigstore/cosign/pkg/types"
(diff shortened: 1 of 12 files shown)


acceptance/image/image.go
```diff
@@ -388,9 +404,39 @@ func JSONAttestationSignaturesFrom(ctx context.Context) string {
...
+
+	return &modified, nil
+}
+
 // AddStepsTo adds Gherkin steps to the godog ScenarioContext
 func AddStepsTo(sc *godog.ScenarioContext) {
 	sc.Step(`^an image named "([^"]*)"$`, createAndPushImage)
 	sc.Step(`^a valid image signature of "([^"]*)" image signed by the "([^"]*)" key$`, createAndPushImageSignature)
 	sc.Step(`^a valid attestation of "([^"]*)" signed by the "([^"]*)" key$`, createAndPushAttestation)
+	sc.Step(`^a valid attestation of "([^"]*)" signed by the "([^"]*)" key with$`, createAndPushAttestationWithPatches)
```
simonbaird, line 441: `...with patches$` maybe?

simonbaird (approved):

internal/image/validate.go
```diff
@@ -18,8 +18,14 @@ package image
 
 import (
 	"context"
+	"encoding/json"
+	"sort"
+	"time"
 
 	conftestOutput "github.com/open-policy-agent/conftest/output"
+	"github.com/qri-io/jsonpointer"
```
lcarva, line 26: This library seems to be unmaintained? Last commit almost two years ago.

internal/image/validate.go
```diff
@@ -127,3 +135,67 @@ func resolveAndSetImageUrl(url string, asi *application_snapshot_image.Applicati
...
+		return times[i].After(times[j])
+	})
+
+	attestationTime := times[0]
+
+	if log.IsLevelEnabled(log.DebugLevel) {
+		log.Debugf("Determined attestation time: %s", attestationTime.Format(time.RFC3339))
+	}
+
+	return context.WithValue(ctx, policy.AttestationTime, &attestationTime)
```
lcarva, line 200: Passing parameters in the context is usually not something we want to do unless we have to. Is this really necessary? An explicit setter on the Policy could be an alternative.

internal/image/validate.go
zregvart: Seems good enough for what we need, we're pulling it via [qri-io/jsonschema](https://github.com/qri-io/jsonschema) as well, which serves us well. I like that it has no additional dependencies...

internal/image/validate.go
zregvart: Yeah, let me refactor that

lcarva (approved):

PR #392, opened 2023-02-07
Support OCI input/output in track bundle

zregvart:
Adds `oci:` prefix support to `--input` and `--output` parameters, so now in addition to files OCI image registries are supported.

Ref. https://issues.redhat.com/browse/HACBS-1658

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -35,7 +35,8 @@
             "mode": "test",
             "program": "${workspaceFolder}/acceptance",
             "args": [
-                "-persist"
+                "-persist",
+                "-tags=@focus"
             ]
         },
         {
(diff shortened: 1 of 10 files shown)


.vscode/launch.json
```diff
@@ -35,7 +35,8 @@
             "mode": "test",
             "program": "${workspaceFolder}/acceptance",
             "args": [
-                "-persist"
+                "-persist",
+                "-tags=@focus"
```
lcarva, line 39: Intentional?

internal/tracker/oci.go
```diff
@@ -0,0 +1,107 @@
...
+		mediaType, err := layer.MediaType()
+		if err != nil {
+			return nil, err
+		}
+
+		if mediaType != openPolicyAgentData {
+			continue
+		}
+
+		if i > 0 {
```
lcarva, line 66: If there are other layers with different unexpected mediaTypes, you could end up with something like this:

```yaml
foo: bar
---
---
spam: maps
```

I think that should be ok because the code should handle an empty doc properly.

lcarva (approved):

.vscode/launch.json
zregvart: Yeah, I find myself running single scenario more often than all scenarios, I can revert if this is uncommon...

internal/tracker/oci.go
zregvart: I could cover that case, but I don't think we'll have more than one layer regardless (this should be fine:tm:)

zregvart:
I'm going to merge this as it is needed for next steps in [HACBS-1658](https://issues.redhat.com/browse/HACBS-1658)

PR #393, opened 2023-02-07
FIX bug where non-existent collection succeeds

robnester-rh:
This commit addresses situations wherein there are zero successes, failures, or warnings. This may be due to specifying a non-existent collection, or specifying an empty policy / ruleset. The error text indicates that the user should check the inputs provided.

### DIFF
--- a/acceptance/examples/allow_all.rego
+++ b/acceptance/examples/allow_all.rego
@@ -1,0 +1,4 @@
+# Simplest never-failing policy
+package main
+
+allow := []
--- a/features/validate_image.feature
+++ b/features/validate_image.feature
@@ -306,6 +306,31 @@ Feature: evaluate enterprise contract
     }
     """
 
+  # Demonstrate that a validation with no failures, warnings, or successes constitutes a failure as nothing was actually evaluated.
+  Scenario: no failures, warnings, or successes
+    Given a key pair named "known"
+    Given an image named "acceptance/ec-happy-day"
+    Given a valid image signature of "acceptance/ec-happy-day" image signed by the "known" key
+    Given a valid Rekor entry for image signature of "acceptance/ec-happy-day"
+    Given a valid attestation of "acceptance/ec-happy-day" signed by the "known" key
+    Given a valid Rekor entry for attestation of "acceptance/ec-happy-day"
+    Given a git repository named "happy-day-policy" with
+      | main.rego | examples/allow_all.rego |
+    Given policy configuration named "ec-policy" with specification
+    """
+    {
+      "sources": [
+        {
+          "policy": [
+            "git::http://${GITHOST}/git/happy-day-policy.git"
+          ]
+        }
+      ]
+    }
+    """
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/ec-happy-day --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --rekor-url ${REKOR} --strict"
+    Then the exit status should be 1
+
   # Demonstrate data sources and using the same rules with different data
   Scenario: policy and data sources
     Given a key pair named "known"
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -160,6 +160,20 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) ([]out
 
 	results = append(results, runResults...)
 
+	// Evaluate total successes, warnings, and failures. If all are 0, then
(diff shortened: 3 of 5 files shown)


internal/evaluator/conftest_evaluator.go
```diff
@@ -160,6 +160,20 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) ([]out
 
 	results = append(results, runResults...)
 
+	// Evaluate total successes, warnings, and failures. If all are 0, then
+	// we have effectively failed, because no tests were actually ran due to
+	// input error, etc.
+	var totalSuccesses, totalWarnings, totalFailures int
```
lcarva, line 166: nit: a single counter should be sufficient

lcarva (approved):

internal/evaluator/conftest_evaluator.go
robnester-rh: ack

zregvart (approved):

lcarva (approved):

acceptance/examples/allow_all.rego
```diff
@@ -0,0 +1,4 @@
+# Simplest never-failing policy
+package main
+
+allow := []
```
simonbaird, line 4: Is this file used anywhere now? If not then I'd suggest deleting it.

features/validate_image.feature
```diff
@@ -306,6 +306,31 @@ Feature: evaluate enterprise contract
...
+  # Demonstrate that a validation with no failures, warnings, or successes constitutes a failure as nothing was actually evaluated.
+  Scenario: no failures, warnings, or successes
+    Given a key pair named "known"
+    Given an image named "acceptance/ec-happy-day"
+    Given a valid image signature of "acceptance/ec-happy-day" image signed by the "known" key
+    Given a valid Rekor entry for image signature of "acceptance/ec-happy-day"
+    Given a valid attestation of "acceptance/ec-happy-day" signed by the "known" key
+    Given a valid Rekor entry for attestation of "acceptance/ec-happy-day"
+    Given a git repository named "happy-day-policy" with
+      | main.rego | examples/allow_all.rego |
```
simonbaird, line 318: Does it still pass if this is changed to happy_day.rego?

acceptance/examples/allow_all.rego
simonbaird: Update: I can see it is used, but I wonder if it needs to be.

internal/evaluator/conftest_evaluator_test.go
```diff
@@ -181,6 +181,40 @@ func TestConftestEvaluatorEvaluate(t *testing.T) {
...
+	p, err := policy.NewOfflinePolicy(ctx, "")
+	assert.NoError(t, err)
+
+	evaluator, err := NewConftestEvaluator(ctx, afero.NewMemMapFs(), []source.PolicySource{
+		testPolicySource{},
+	}, p)
+
+	assert.NoError(t, err)
+	actualResults, err := evaluator.Evaluate(ctx, inputs)
+	assert.Error(t, err)
```
simonbaird, line 214: Could we have an assertion here that checks the error is the error we're expecting? Something like this maybe:

    assert.Equal(t, "no successes, warnings, or failures, check input", err.Error())

simonbaird: This might be it:

    assert.ErrorContains(t, err, "no successes, warnings, or failures, check input")

simonbaird (approved):
Approve, but consider my suggestion about confirming the expected error in the tests.

internal/evaluator/conftest_evaluator_test.go
robnester-rh: ack

features/validate_image.feature
robnester-rh: No.

acceptance/examples/allow_all.rego
robnester-rh: Yes. My understanding from discussing with zregvart is that, since there is no deny rule, this triggers a 0 warning, 0 failure, 0 success count situation whereas `happy_day.rego` does not.

robnester-rh:
> Approve, but consider my suggestion about confirming the expected error in the tests.

Thank you for the suggestion. I've updated the code to reflect your suggestion!

acceptance/examples/allow_all.rego
zregvart: Yeah, this is to help test that `ec validate` will not succeed, `allow` is not supported by Conftest, so this tests that at least one supported rule has been evaluated.

PR #406, opened 2023-02-09
Docs urls in inspect output

simonbaird:


simonbaird:
It's just the top commit. Draft because it should be rebased after the other PR is merged.

simonbaird:
Rebased and moved out of draft.

internal/opa/output.go
```diff
@@ -83,13 +85,22 @@ func prepTemplateFields(a *ast.AnnotationsRef) TemplateFields {
 		// conventions i.e. "package policy.<type>.<name>"
 		shortPath = trimmedPath
 	}
+
+	shortName := getShortName(a)
+
+	var docsUrl string
+	if len(pathStrings) > 4 && shortName != "" && (strings.HasPrefix(trimmedPath, "policy.pipeline.") || strings.HasPrefix(trimmedPath, "policy.release.")) {
+		docsUrl = fmt.Sprintf("https://hacbs-contract.github.io/ec-policies/%s_policy.html#%s__%s", pathStrings[2], pathStrings[3], shortName)
```
simonbaird, line 93: Later we could have a package annotation with the default docs url prefix. For now - hard coded.

internal/opa/output.go
```diff
@@ -83,13 +85,22 @@ func prepTemplateFields(a *ast.AnnotationsRef) TemplateFields {
 		// conventions i.e. "package policy.<type>.<name>"
 		shortPath = trimmedPath
 	}
+
+	shortName := getShortName(a)
+
+	var docsUrl string
+	if len(pathStrings) > 4 && shortName != "" && (strings.HasPrefix(trimmedPath, "policy.pipeline.") || strings.HasPrefix(trimmedPath, "policy.release.")) {
```
lcarva, line 92: Why greater than 4?

lcarva (approved):

internal/opa/output.go
simonbaird: Hmm yeah, probably should be `len(..) == 4`

simonbaird: Actually 5, because it still include "data" as element 0.

internal/opa/output.go
```diff
@@ -47,12 +47,14 @@ type TemplateFields = struct {
...
 	ShortName   string
+	DocsUrl     string
 }
 
 var templates = map[string]string{
 	"text": hd.Doc(`
 		{{ .TrimmedPath }}.{{ .ShortName }} ({{ .WarnDeny }})
-		{{ .Title }}
+		{{ if .DocsUrl }}{{ .DocsUrl }}
+		{{ end }}{{ .Title }}
```
zregvart, line 57: Not important at all, just wanted to mention this is another way of doing this:
```suggestion
		{{ with .DocsUrl }}{{ . }}
		{{ end }}{{ .Title }}
```

internal/opa/output.go
```diff
@@ -83,13 +85,22 @@ func prepTemplateFields(a *ast.AnnotationsRef) TemplateFields {
 		// conventions i.e. "package policy.<type>.<name>"
 		shortPath = trimmedPath
 	}
+
+	shortName := getShortName(a)
+
+	var docsUrl string
```
zregvart, line 91: It'd be nice to have the URL in Rego in a annotation, perhaps as a followup...

zregvart (approved):

internal/opa/output.go
lcarva: Ah so this is checking if it includes a certain prefix?. If so, maybe something like this is more easily read: `len(pathStrings) > len("myprefix")`. Or even better `strings.HasPrefix(pathStrings, "myprefixX")`, if `X` is always the same value.

internal/opa/output.go
simonbaird: I like it.

internal/opa/output.go
simonbaird: Agreed. I added some commentary along those lines.

simonbaird:
Thanks for the reviews. I made a few tweaks. Will stop fussing with this and probably merge when it goes green.

PR #408, opened 2023-02-10
Add created by metadata to the data layer

zregvart:
The metadata key will hold the whole invocation of the ec-cli used to track the bundle, from this we can (in the future) determine what Tekton bundles were used to create the OPA/Conftest data bundle and with that (perhaps) determine the delta that is needed to generate the correct (i.e. without gaps) acceptable bundles and required tasks list.

Also adds a test for the `conftest pull` compatiblity.

Ref. https://issues.redhat.com/browse/HACBS-1658


zregvart:
Created as a draft as I do want to add a bit of documentation to this.

PR #411, opened 2023-02-14
create owners file for artifacthub

joejstuart:
This is for adding the task to artifacthub. 

https://issues.redhat.com/browse/HACBS-1579

### DIFF
--- a/task/0.1/README.md
+++ b/task/0.1/README.md
@@ -9,9 +9,8 @@ kubectl apply -f https://raw.githubusercontent.com/hacbs-contract/ec-cli/main/ta
 ### Required
 * **IMAGES**: A JSON formatted list of images. 
 ### Optional
-* **POLICY_CONFIGURATION**: Name of the policy configuration (EnterpriseContractPolicy
-        resource) to use. `namespace/name` or `name` syntax supported. If
-        namespace is omitted the namespace where the task runs is used.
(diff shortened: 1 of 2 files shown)


task/0.1/README.md
```diff
@@ -1,2 +1,43 @@
...
+## Install the task
+kubectl apply -f https://raw.githubusercontent.com/hacbs-contract/ec-cli/main/task/0.1/verify-enterprise-contract.yaml
+
+## Parameters
+### Required
+* **IMAGES**: A JSON formatted list of images. 
+### Optional
+* **POLICY_CONFIGURATION**: Name of the policy configuration (EnterpriseContractPolicy
+        resource) to use. `namespace/name` or `name` syntax supported. If
+        namespace is omitted the namespace where the task runs is used.
```
zregvart, line 14: We also support inline policy here, perhaps something like:
```suggestion
* **POLICY_CONFIGURATION**: Name or inline policy in JSON configuration to use. For name `namespace/name` or `name` syntax supported. If
        namespace is omitted the namespace where the task runs is used. For inline policy provide the [specification](https://hacbs-contract.github.io/ecc/main/reference.html#k8s-api-github-com-hacbs-contract-enterprise-contract-controller-api-v1alpha1-enterprisecontractpolicyspec) as JSON.
```

zregvart (approved):
We might document the inline policy option, or we might not, not 100% sure.

PR #412, opened 2023-02-15
ADD Antora extension to parse Tekton Task yamls

robnester-rh:
This commit adds a first pass at an Antora extension that parses Tekton Task YAML to generate asciidoc documentation.


robnester-rh:
The CLI acceptance test failure isn't related to any changes in this PR.

docs/antora.yml
```diff
@@ -23,6 +23,6 @@ nav:
 ext:
   collector:
     run:
-      command: make reference-docs
+      command: make all-docs
     scan:
-      dir: dist/reference
+      dir: dist
```
lcarva, line 28: Will the other files in this directory, e.g. `ec_linux_amd64`, cause any issues?

internal/task_documentation/task_documentation.go
```diff
@@ -0,0 +1,146 @@
...
+	var err error
+	defer func() {
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+	}()
+
+	// YAML
+	if *yamlDir != "" {
```
lcarva, line 49: If it is an empty string, should it exit with an error?

internal/task_documentation/task_documentation.go
```diff
@@ -0,0 +1,146 @@
...
+		if err := findTektonTasks(*startPath, *taskDir, *yamlDir); err != nil {
+			return
+		}
+	}
+}
+
+func findTektonTasks(basePath, taskDir, dstDir string) error {
+	var err error
+	if basePath == "" {
+		basePath, err = getGitRoot()
```
lcarva, line 62: Why not just assume current working directory?

internal/task_documentation/task_documentation.go
```diff
@@ -0,0 +1,146 @@
...
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+const DirectoryPermissions = 0755
+
+var yamlDir = flag.String("yaml", "", "Location of the generated YAML files")
+var startPath = flag.String("startPath", "", "Location of path to start from")
+var taskDir = flag.String("taskDir", "task", "Location of tekton tasks to scrape")
```
lcarva, line 36: Do we need both `taskDir` and `startPath`? Why not just take one path parameter?

internal/task_documentation/task_documentation.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+	return out.Close()
```
lcarva, line 121: Isn't `out.Close()` handled by the `defer` a few lines above?

internal/task_documentation/task_documentation.go
```diff
@@ -0,0 +1,146 @@
...
+	task := TektonTask{}
+
+	data, err := os.ReadFile(file)
+	if err != nil {
+		return false, err
+	}
+
+	err = yaml.Unmarshal(data, &task)
+	if err != nil {
+		return false, err
```
lcarva, line 139: If it's a yaml file, but not a kubernetes resource, do we want to return an error? Maybe swallowing the error here makes sense.

internal/task_documentation/task_documentation.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	if task.APIVersion == "tekton.dev/v1beta1" && task.Kind == "Task" {
```
lcarva, line 142: Upstream is already on `v1`. The `params` and `results` attributes are the same there. Maybe accept `v1` here as well?

lcarva (approved):

.github/workflows/release-reference-antora-extension.yaml
```diff
@@ -26,7 +26,7 @@ name: Release reference-antora-extension
 
 
 jobs:
-  release-antora-extension:
+  release-reference-antora-extension:
```
zregvart, line 29:
```suggestion
  release-antora-extensions:
```

We could keep the same workflow and use matrix strategy to run it twice in two different directories, keeps the maintenance overhead low. See an example [here](https://github.com/hacbs-contract/ec-policies/blob/398a79a4830ee89d917aa1ad6f5a3fb78e238f3b/.github/workflows/docs.yaml#L12-L18)

Makefile
```diff
@@ -57,6 +57,16 @@ reference-docs: ## Generate reference documentation input YAML files
...
 
+.PHONY: tekton-task-docs
+tekton-task-docs: ## Generate reference documentation input YAML files
+	@rm -rf dist/task
+	@go run internal/task_documentation/task_documentation.go -yaml dist/task
+
+.PHONY: all-docs
+all-docs: ## Generate reference documentation input YAML files
+	@$(MAKE) reference-docs
+	@$(MAKE) tekton-task-docs
```
zregvart, line 68: This is a commonplace pattern of expressing dependencies between targets

```suggestion
all-docs: reference-docs tekton-task-docs ## Generate reference documentation input YAML files
```

docs/antora.yml
zregvart: Best keep this separate, e.g. if we need to delete all the generated `.adoc` files we can remove just `dist/reference` and not care if there are any other files in `dist`. Might be better to have two collectors here.

docs/modules/ROOT/pages/task.adoc
```diff
@@ -0,0 +1,4 @@
+= ec command line task documentation
```
zregvart, line 1:
```suggestion
= Enterprise Contract Tekton Tasks
```

We'll have another one in [HACBS-1765](https://issues.redhat.com/browse/HACBS-1765), and possibly different versions in the future.

docs/modules/ROOT/nav.adoc
```diff
@@ -3,4 +3,5 @@
 {{#each reference}}
 *** xref:{{ path }}[{{ name }}]
 {{/each}}
-*** xref:verify-enterprise-contract.adoc[verify-enterprise-contract] task   
+** xref:task.adoc[Tasks]
```
zregvart, line 6: Perhaps?
```suggestion
** xref:tasks.adoc[Tasks]
```

internal/task_documentation/task_documentation.go
zregvart: Make `-startPath` required?

internal/task_documentation/task_documentation.go
```diff
@@ -0,0 +1,146 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package main
```
zregvart, line 17: Wouldn't it be simpler if we had a symlink pointing to the `tasks` directory from `docs`. This is what we do in [ec-policies](https://github.com/hacbs-contract/ec-policies/tree/main/antora/docs) for `policies`?

reference-antora-extension/index.js
```diff
@@ -36,7 +36,7 @@ module.exports.register = function () {
         const template = Handlebars.compile(referenceTemplate.contents.toString());
         const referenceDocs = []
 
-        content.files.filter(f => f.src.extname === '.yaml').forEach(f => {
+        content.files.filter(f => f.src.extname === '.yaml' && f.src.basename.startsWith('ec')).forEach(f => {
```
zregvart, line 39: Might be a foot-gun if we name a task yaml file with with `ec*`. Let's separate by output directories, e.g. use `dist/cli-reference` and `dist/task-reference` instead.  Then this filter could be:
```suggestion
        content.files.filter(f => f.src.extname === '.yaml' && f.src.scanned.startsWith('dist/cli-reference')).forEach(f => {
```

.github/workflows/release-tektonTask-antora-extension.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+---
+name: Release tektonTask-antora-extension
+
+"on":
+  push:
+    branches:
+      - main
+    paths:
+      - 'tektonTask-antora-extension/**'
```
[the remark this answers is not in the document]
robnester-rh: ack

docs/antora.yml
robnester-rh: @zregvart, this isn't for laying down files, it's for scanning. 

To @lcarva's question, no it won't care about other files in the directory.

internal/task_documentation/task_documentation.go
robnester-rh: @lcarva I just feel it's bad to assume. Either we explicitly pass a `startPath` or we start from the top.

internal/task_documentation/task_documentation.go
robnester-rh: My thought here was that if a `taskDir` is provided, we only start to walk from that subdirectory of the `startPath`. If not, we would walk from the root, collecting all Tekton tasks as we walk the entirety of the directory structure under the `startPath`.

internal/task_documentation/task_documentation.go
robnester-rh: So the `defer out.Close()` calls are handling errors on open and create. The `return out.Close()` is ensuring that the file is closed after copying.

internal/task_documentation/task_documentation.go
robnester-rh: I'd prefer to leave it here since the caller can determine if something should be done with the error (maybe the file disappeared from underneath us?, etc)

internal/task_documentation/task_documentation.go
robnester-rh: Good call. Will do.

.github/workflows/release-reference-antora-extension.yaml
robnester-rh: I'll check into that.

Makefile
robnester-rh: ack, I had that previously and changed it thinking that we would want to ensure we didn't have failures in the reference creation. I'll use the previous approach.

reference-antora-extension/index.js
robnester-rh: Agreed. I'll look at doing this. I like the idea of updating `reference` to `cli-reference` and will do so.

docs/modules/ROOT/pages/task.adoc
robnester-rh: ack

internal/task_documentation/task_documentation.go
lcarva: I don't think that's how defer works. `defer out.Close()` will make `out.Close()` be called at the end of the function, regardless of where it returns from.

internal/task_documentation/task_documentation.go
robnester-rh: I suppose that prompts the question: are we dogmatic that all tasks will reside only under the `task` directory? If so, I think that the logic in this file can be simplified.

docs/modules/ROOT/nav.adoc
robnester-rh: ack

internal/task_documentation/task_documentation.go
robnester-rh: You're correct, the `defer` does handle the closing. My thought with the final `return out.Close()` would be a bit of an explicit error catching. If we rely just on the `defer`, this function will provide the error regarding opening the input file or creating the output file to the caller, but if closing the output file has an error, we end up not passing that back if we don't return the result of calling `out.Close()`

internal/task_documentation/task_documentation.go
zregvart: Yeah, I think that's needed for the Artifact Hub (cc @joejstuart)

internal/task_documentation/task_documentation.go
lcarva: I understanding your reasoning, but the [docs](https://pkg.go.dev/os#File.Close) say this:
> Close will return an error if it has already been called. 

The happy path on this function will always call `Close` twice, and will always generate an error on the second call, but that error is ignored. :shrug:

robnester-rh: Well then, no need for that. :man_shrugging:

simonbaird:
I'm not sure why we need the `internal/task_documentation/task_documentation.go` since IIUC we don't need it for the rego or the acceptable bundles. Why not let the dir walking happen in the javascript? Maybe a symlink is needed so the files are visible, but it seems like less complex solution.

simonbaird (dismissed):
Looking for reasoning re the inclusion of a go "dir walk and copy" for the task definitions.

simonbaird:
Ack, thanks, it makes sense. I forgot we have a similar copy mechanism for the command reference.

simonbaird (commented):
Removing my request changes.

internal/task_documentation/task_documentation.go
simonbaird: I think we can decide this. Tasks in the tasks directory makes sense to me.

zregvart:
> Ack, thanks, it makes sense. I forgot we have a similar copy mechanism for the command reference.

Just to make sure, we're on the same page here. We don't copy for the CLI reference, the Corbra structs are transformed and marshalled to YAML and then placed in `dist/reference`. For the Tekton task we already have YAML and all we do is to copy it over to `dist/task`. Having go code where a symlink would do seems over-engineered.

robnester-rh:
> would

Agreed. This approach has been abandoned.

simonbaird:
I tried a build and saw something like this in the nav:

- [Enterprise Contract CLI](https://github.com/hacbs-contract/ec-cli/pull/index.html)
  - [Documentation](https://github.com/hacbs-contract/ec-cli/pull/index.html)
    - [Reference](https://github.com/hacbs-contract/ec-cli/pull/reference.html)
      - [{{ name }}](https://github.com/hacbs-contract/ec-cli/pull/412#{{%20path%20}})
    - [Tasks](https://github.com/hacbs-contract/ec-cli/pull/tasks.html)
      - [{{ name }}](https://github.com/hacbs-contract/ec-cli/pull/412#{{%20path%20}})

Maybe the handlebars isn't doing what it's supposed to in that one place? I'll see if I can figure it out.

robnester-rh:
> I tried a build and saw something like this in the nav:
> 
> * [Enterprise Contract CLI](https://github.com/hacbs-contract/ec-cli/pull/index.html)
>   
>   * [Documentation](https://github.com/hacbs-contract/ec-cli/pull/index.html)
>     
>     * [Reference](https://github.com/hacbs-contract/ec-cli/pull/reference.html)
>       
>       * [{{ name }}](https://github.com/hacbs-contract/ec-cli/pull/412#%7B%7B%20path%20%7D%7D)
>     * [Tasks](https://github.com/hacbs-contract/ec-cli/pull/tasks.html)
>       
>       * [{{ name }}](https://github.com/hacbs-contract/ec-cli/pull/412#%7B%7B%20path%20%7D%7D)
> 
> Maybe the handlebars isn't doing what it's supposed to in that one place? I'll see if I can figure it out.

Ensure that in your hacbs-contract.github.io repo, you’ve amended your antora playbook to point to the proper branch for testing purposes.

docs/modules/ROOT/nav.adoc
```diff
@@ -1,6 +1,3 @@
-* xref:index.adoc[Documentation]
-** xref:reference.adoc[Reference]
-{{#each reference}}
-*** xref:{{ path }}[{{ name }}]
-{{/each}}
-*** xref:verify-enterprise-contract.adoc[verify-enterprise-contract] task   
+include::partial$main_nav.adoc[]
+include::partial$cli_nav.adoc[]
+include::partial$tasks_nav.adoc[]
```
zregvart, line 3: Very neat!

zregvart (approved):
Awesome!

simonbaird:
> Ensure that in your checked out hacbs-contract.github.io, you’ve amended your antora playbook to point to the proper branch for testing purposes.

Ah, thanks.

simonbaird (approved):

PR #424, opened 2023-02-16
count successes on each conftest run

joejstuart:


zregvart (approved):
LGTM, but the unit tests need to be updated, most likely the acceptance tests as well...

internal/applicationsnapshot/report.go
```diff
@@ -35,6 +35,7 @@ type Component struct {
 	Violations []conftestOutput.Result  `json:"violations"`
 	Warnings   []conftestOutput.Result  `json:"warnings"`
 	Success    bool                     `json:"success"`
+	Passed     int                      `json:"passed"`
```
joejstuart, line 38: `Successes` or `Passed`? I'm leaning towards `Passed`

lcarva: Maybe `successes` because that's also a noun like `warnings` and `violations`?

joejstuart: This is a good point. Conftest uses Passed, but I can't find a good reason.

simonbaird: One note: We want to (in the future) actually list details about the successes. So maybe reserve `successes` for that purpose and use something like SuccessCount for this.

simonbaird: `Passed` isn't so bad though, idk.

lcarva: Oh! I thought this was about adding the list of passing rules, not just the count. +1 to Simon's suggestion.

simonbaird (approved):
Lgtm, assuming you can get the tests back to passing.

PR #425, opened 2023-02-16
Support authentication when resolving image digest

lcarva:


lcarva:
This is all that was needed to run `ec validate image` against a private repository. All other places seem to handle authentication already.

simonbaird (approved):
Lgtm. Does it work?

lcarva:
I tested against a private repo, and it does work.

PR #428, opened 2023-02-17
Fix the `IMAGE_TAG` should be `snapshot`

zregvart:


zregvart:
Merging to fix the `snapshot` tag

zregvart:
Well, that was not great. Working on a revert + fix.

PR #437, opened 2023-02-22
Pass afero.Fs via context.Context

zregvart:
There is no need to pass `afero.Fs` to functions that receive `context.Context`, the `afero.Fs` implementation should be contained within the context already.


lcarva (approved):
Super!

simonbaird (approved):

PR #446, opened 2023-02-23
Fix src origin for pages

robnester-rh:
In https://github.com/hacbs-contract/ec-cli/pull/445 the origin field was removed because of issues building
documentation upstream. Issue was resolved by using a valid origin value of 'ec-cli'


simonbaird:
Trying to fix the downstream builds. Let's merge.

PR #478, opened 2023-03-01
Include successful rule outcomes in output

zregvart:
In addition to failures and warnings now successes are added to the output. Given that from Conftest we only receive the number of successful rule outcomes this relies on the inspected AST data from the policy sources.

We can't tell the message a successful rule would generate, this is lost when Conftest discards this and doesn't include it in the result, a fixed message of `"Pass"` is reported.

Ref. https://issues.redhat.com/browse/HACBS-1791

### DIFF
--- a/cmd/validate/image_test.go
+++ b/cmd/validate/image_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"github.com/hacbs-contract/ec-cli/internal/applicationsnapshot"
+	"github.com/hacbs-contract/ec-cli/internal/evaluator"
 	"github.com/hacbs-contract/ec-cli/internal/output"
 	"github.com/hacbs-contract/ec-cli/internal/policy"
 	"github.com/hacbs-contract/ec-cli/internal/utils"
@@ -174,11 +175,21 @@ func Test_ValidateImageCommand(t *testing.T) {
 			AttestationSyntaxCheck: output.VerificationStatus{
 				Passed: true,
 			},
-			PolicyCheck: []conftestOutput.CheckResult{
+			PolicyCheck: evaluator.CheckResults{
 				{
-					FileName:  "test.json",
-					Namespace: "test.main",
-					Successes: 14,
+					CheckResult: conftestOutput.CheckResult{
+						FileName:  "test.json",
+						Namespace: "test.main",
+						Successes: 1,
+					},
+					Successes: []conftestOutput.Result{
+						{
+							Message: "Pass",
+							Metadata: map[string]interface{}{
+								"code": "policy.nice",
+							},
+						},
+					},
 				},
 			},
 			ImageURL: url,
@@ -211,8 +222,10 @@ func Test_ValidateImageCommand(t *testing.T) {
 			"containerImage": "registry/image:tag",
 			"violations": [],
 			"warnings": [],
-			"success": true,
-			"successCount": 14
+			"successes": [
+				{"msg": "Pass", "metadata": {"code": "policy.nice"}}
+			],
+			"success": true
 		  }
 		]
 	  }`, mockPublicKey), out.String())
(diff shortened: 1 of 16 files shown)


internal/evaluator/conftest_evaluator.go
```diff
@@ -190,25 +213,59 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) ([]out
...
-		runResults[i] = result
+		results = append(results, CheckResult{CheckResult: result})
 	}
 
-	results = append(results, runResults...)
+	// set successes, these are not provided in the Conftest results, so we
+	// reconstruct these from the parsed rules, any rule that hasn't been
+	// touched by adding metadata must have succeeded
+
+	// TODO see about multiple resuts, somehow; using results[0] for now
```
simonbaird, line 223:
```suggestion
	// TODO see about multiple results, somehow; using results[0] for now
```

cmd/validate/image_test.go
```diff
@@ -211,8 +214,8 @@ func Test_ValidateImageCommand(t *testing.T) {
 			"containerImage": "registry/image:tag",
 			"violations": [],
 			"warnings": [],
-			"success": true,
-			"successCount": 14
+			"successes": [],
+			"success": true
```
lcarva, line 218: Is this right? I thought we said that if there are no successes, no warnings, and no failures, then it should fail.

simonbaird (approved):
Awesome.

My only idea, (and it could be in a future PR and/or future Story), is maybe we should add a flag `--no-success-output` or `--suppress-success-output` which let's you turn it off if you don't want it. (Though maybe the count can stay, idk..?)

And this made me think, why not also a `--no-warning-output` that works the same way in case you don't care about warnings.

internal/applicationsnapshot/report_test.go
```diff
@@ -150,28 +150,27 @@ components:
...
     success: true
-    successCount: 5
 `
 
 	var components []Component
 	for _, component := range snapshot.Components {
 		c := Component{
-			Violations:   []output.Result{},
-			Success:      true,
-			SuccessCount: 5,
```
lcarva, line 174: Where did the 5 come from? :thinking:

internal/evaluator/conftest_evaluator.go
```diff
@@ -190,25 +213,59 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) ([]out
...
+	if l := len(rules); l > 0 {
+		results[0].Successes = make([]output.Result, 0, l)
+	}
+
+	for code, rule := range rules {
```
lcarva, line 228: I don't follow how this is filtering out rules that did fail/warn.

lcarva (commented):
This looks great. Just a couple of questions to help me understand it better

zregvart:
Thanks for the review, I'll have something smarter to say a bit later

internal/evaluator/conftest_evaluator.go
lcarva: Oh that's confusing! Unrelated to this PR it seems

cmd/validate/image_test.go
zregvart: Updated the test. The result mocked from Conftest had `Successes: 14` but no
shadowed `Successes` array of results, that meant that the output lacked
`successes` but did not fail as the total number of rules was `14`. I've added
one successful result and set `14` to `1`.

internal/evaluator/conftest_evaluator.go
zregvart: We have parsed AST of rego rules, we don't know which ones passed and which ones
failed, so we need to delete the ones that we know failed (or warninged), that
should in theory leave only the ones that succeeded. So very much part of this PR.

internal/applicationsnapshot/report_test.go
zregvart: This test kinda sucks, I'm going to refactor it

zregvart:
Please take note of:

[Refactor output and report test](https://github.com/hacbs-contract/ec-cli/pull/478/commits/968f7ef002ab041c8f36557bd53c29ba2727a048)
Makes the `report_test.go` test more. Also, in a spur of the moment I thought _why are we including empty results in the report?_ and thought that we should not.

lcarva (approved):
This is great!

lcarva:
Let's merge this because I have some changes that will be impacted by it.

PR #479, opened 2023-03-01
Run tests and acceptance tests in parallel

zregvart:
Splits the tests and acceptance tests in two jobs so they'll run in parallel. This should speed up the Checks workflow by a couple of minutes.


zregvart:
Rerunning the acceptance tests on docker flake ([HACBS-1354](https://issues.redhat.com/browse/HACBS-1354)) is now a bit quicker as well

lcarva (approved):

PR #465, opened 2023-02-27
Conditionally include rule title and description

zregvart:
Adds the title and description defined in annotations of policy rules to the output if `--info` parameter is passed to the `ec validate image` command.

Ref. https://issues.redhat.com/browse/HACBS-1735

(includes #464 so tests don't fail)

PR #480, opened 2023-03-02
Convert result "collections" Metadata to []string

zregvart:
The change in #465 converted the `result.Metadata["collections"]` to `[]string` from `[]any` leading to `extractCollections` returning no collections for a given rule. This meant that when filtering by collections was used no failures would be reported.

This changes `result.Metadata["collections"]` to always be `[]string`. Any other given value type will cause `collections` to be removed from `result.Metadata".

Side note: this was not detected by unit tests because the tests did not provide any `.rego` files for the rule inspection to inspect, so no conversion from `[]any` to `[]string` was performed in `addMetadata`. And we have no acceptance tests that specifically test rule filtering by collections.

### DIFF
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -47,6 +47,16 @@ type testRunner interface {
 	Run(context.Context, []string) ([]output.CheckResult, error)
 }
 
+const (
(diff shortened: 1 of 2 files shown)


psturc (approved):
verified these changes with the test that identified the bug

PR #482, opened 2023-03-02
make validate pipeline more generic

joejstuart:
This essentially just renames `validate pipeline` to `validate definition-file`. I thought about leaving `validate pipeline` and then adding a subcommand for each definition file we want to process, but for now having a generic `validate definition-file` seems more efficient. 
https://issues.redhat.com/browse/HACBS-1900


zregvart:
Perhaps `ec validate definition`, `definition-file` seems one-way street, say if we want to validate a bundle, or a cluster resource?

joejstuart:
> Perhaps `ec validate definition`, `definition-file` seems one-way street, say if we want to validate a bundle, or a cluster resource?

Yes, that does sound better. I'll update.

zregvart (approved):

PR #485, opened 2023-03-03
Include policy output in the acceptance test

zregvart:


features/validate_image.feature
```diff
@@ -611,6 +611,7 @@ Feature: evaluate enterprise contract
     }
     """
 
+  @focus
```
lcarva, line 614: Accidental?

lcarva (approved):

PR #489, opened 2023-03-07
Sort results from ec-cli

mbestavros:
This is a very simple sorting implementation on `ec-cli`'s result outputs.

From what I can tell, a lot of the reason result ordering wasn't deterministic in the first place is down to the ordering of _evaluators_ not being deterministic in the first place. You can see how the current code prepares input to `Evaluate()` here: https://github.com/hacbs-contract/ec-cli/blob/main/internal/image/validate.go#L101

This doesn't seem to be an issue internally (at least as far as I can tell), so I went with the approach of sorting the merged lists of successes, warnings, and violations. Note that the sort method is different for violations, since the `msg` field is often more descriptive than the `metadata.code` field. However, the opposite is true for successes.

This is my first Go pull request ever, so I'm **sure** there's a better way to do this. Feedback is welcome!

Ref: https://issues.redhat.com/browse/HACBS-1771

cc @zregvart @simonbaird @lcarva and anyone else interested :smile:


internal/output/output.go
```diff
@@ -198,6 +199,9 @@ func (o Output) Violations() []output.Result {
 	violations = o.AttestationSyntaxCheck.addToViolations(violations)
 	violations = o.addCheckResultsToViolations(violations)
 
+	sort.Slice(violations, func(i, j int) bool {
+		return violations[i].Message < violations[j].Message
```
simonbaird, line 203: Why not sort by the "code" value here also? Or maybe by the code first and message second?

simonbaird:
The acceptance failure seems bogus, but the unit test failures are useful. Do you get same failures with `make test` locally?

internal/output/output.go
mbestavros: @simonbaird `code` seems to be less descriptive for violations than other types. For example, when testing `ec-cli` against `quay.io/cuipinghuo/hermetic_build:0.3`, I get:

```
- msg: Pipeline task 'appstudio-configure-build' does not contain a bundle reference
  metadata:
    code: attestation_task_bundle.disallowed_task_reference
    effective_on: "2022-01-01T00:00:00Z"
- msg: Pipeline task 'appstudio-init' does not contain a bundle reference
  metadata:
    code: attestation_task_bundle.disallowed_task_reference
    effective_on: "2022-01-01T00:00:00Z"
- msg: Pipeline task 'build-container' does not contain a bundle reference
  metadata:
    code: attestation_task_bundle.disallowed_task_reference
    effective_on: "2022-01-01T00:00:00Z"
- msg: Pipeline task 'clair-scan' does not contain a bundle reference
  metadata:
    code: attestation_task_bundle.disallowed_task_reference
    effective_on: "2022-01-01T00:00:00Z"
...
```

Note that all `code` values are the same. In practice, sorting on that value _still_ isn't deterministic - the task names will be mixed up from run to run.

Sorting against `msg`, on the other hand, captures these differences and tends to be more deterministic.

I have tried sorting on both criteria using the `sort.Slice` function sequentially, but the end result seemed to be that the later sort just overwrote the first one, making it redundant. Perhaps there's a much smarter sort method in Go I'm unaware of - if anyone happens to know one, please chime in :)

Compare and contrast that to the success results:

```
- msg: Pass
  metadata:
    code: attestation_task_bundle.empty_task_bundle_reference
- msg: Pass
  metadata:
    code: attestation_task_bundle.missing_required_data
- msg: Pass
  metadata:
    code: attestation_task_bundle.out_of_date_task_bundle
```

In this case, `msg` doesn't tell us anything, but `metadata.code` does - so it makes sense to sort on `metadata.code` for success.

I don't have enough example output for "warnings" to make a call one way or the other, though.

simonbaird: There's some value in sorting by code first and msg second. I think it would be a little weird if violations with the same "code" value were not together in the sorted output.

Sorting by two fields should be doable. You have to put the logic in the sort function. Let me try to demonstrate.

simonbaird, line 203:
```suggestion
                iCode := evaluator.ExtractStringFromMetadata(successes[i], "code")
                jCode := evaluator.ExtractStringFromMetadata(successes[j], "code")
                if iCode == jCode {
                    # Secondary sort by the message
                    return violations[i].Message < violations[j].Message
                }
		return iCode < jCode
```

simonbaird: I didn't test it, but something like that should work.

simonbaird:
Not sure if there's a good way to extract the comparison logic. Can a go function return a function?

I'm imagining something like this:

```
    sort.Slice(violations, sorterFunc(violations))
```

internal/output/output.go
mbestavros: @simonbaird Ah, yep, that makes sense. I toyed with trying to sort on both values in return statement, but I couldn't get it to work, and I didn't want to make the closure unnecessarily long, But this is probably better sort logic anyway, so I'll incorporate it.

mbestavros:
@simonbaird I don't know if that's even necessary - all three slices (`successes`, `warnings`, and `violations`) should behave the same if passed to a separate `sort` function, and we can just pass that function into `sort.Slice()`.

mbestavros:
@simonbaird How does it look now?

internal/evaluator/conftest_evaluator.go
```diff
@@ -279,6 +279,18 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) (Check
 	return results, nil
 }
 
+// Helper function to sort Result lists.
```
simonbaird, line 282: Nice! I like it. Could this function be moved over to `internal/output/output.go`? It looks like it's only used there, so maybe it can be just a private function in that file now.

simonbaird:
Better get the tests passing, it might encourage more reviewers. :smile:

internal/evaluator/conftest_evaluator.go
mbestavros: @simonbaird That's actually how I originally wrote it, but putting it there means we'd need to export `extractStringFromMetadata()` from `conftest_evaluator.go` so it can be accessed from `internal/output/output.go`.

I could probably just replicate `extractStringFromMetadata()` in the function if we really want to minimize exports. But that (and tests) are a job for tomorrow :wink:

zregvart: I think sorting in the `internal/output` is the way to go, never mind the duplication of `result.Metadata["code"]`

zregvart, line 282: Nitpick, seems like it will get refactored regardless. The convention is to start with the function name (see https://go.dev/blog/godoc)

```suggestion
// SortResults sorts Result slices.
```

internal/evaluator/conftest_evaluator.go
```diff
@@ -279,6 +279,18 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) (Check
...
+func SortResults(i, j int, slice []output.Result) bool {
```
zregvart, line 283: For a function named `SortResults` I was expecting a signature like:

```suggestion
func SortResults(results []output.Result) []output.Result {
```

internal/evaluator/conftest_evaluator.go
```diff
@@ -279,6 +279,18 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) (Check
...
+	// First, sort by shortcode
+	iCode := extractStringFromMetadata(slice[i], "code")
+	jCode := extractStringFromMetadata(slice[j], "code")
+	if iCode == jCode {
+		// If inconclusive, sort by the message
+		return slice[i].Message < slice[j].Message
+	}
+	return iCode < jCode
```
zregvart, line 291: I think the sort order should be code, term and then message. Terms are further classifications of codes, e.g. for `test.test_result_unsupported` code there can be a number of `term`s

zregvart (commented):
I would have expected some changes to the acceptance tests, I guess we compare JSON for content not for order. Might be good as a followup to add explicit check for sorted output in the acceptance test

internal/evaluator/conftest_evaluator.go
mbestavros: @zregvart How exactly do you mean by `term`s? To use your example of `test.test_result_unsupported`, would the `term` be `test_result_unsupported`?

lcarva: `term` is an additional piece of metadata. It's dynamically added to a policy rule violation on the rego side of things. Here's what it looks like: https://github.com/hacbs-contract/ec-policies/blob/main/policy/release/cve_test.rego#L38-L51

mbestavros: Thanks @lcarva!

internal/evaluator/conftest_evaluator.go
simonbaird: Personally I don't mind, whichever feels right.

internal/evaluator/conftest_evaluator.go
simonbaird: Yeah, maybe this one could be named `ResultCompare` or something like that. And the comment could be changes to be more like `// ResultCompare is for use with sort.Slice to define the sort order`. Or, you could refactor and write a different SortResults that does have the signature Zoran was expecting.

internal/evaluator/conftest_evaluator.go
mbestavros: When I add in sorting via `term` as well (in the revision I just pushed), I end up using the `extractStringFromMetadata()` function enough that I think it's worth exporting, IMO.

internal/evaluator/conftest_evaluator.go
mbestavros: I've done the latter - I think it's cleaner code anyway. The interface should match @zregvart's proposal while encapsulating the same sort logic as was there previously.

internal/evaluator/conftest_evaluator.go
mbestavros: Updated, thanks!

mbestavros:
To all following, I've refactored the sort function so it makes more sense and updated tests to work with the new sorting. I can also look into a new test to ensure everything is sorted, but that will take more time.

internal/output/output.go
```diff
@@ -198,6 +199,7 @@ func (o Output) Violations() []output.Result {
 	violations = o.AttestationSyntaxCheck.addToViolations(violations)
 	violations = o.addCheckResultsToViolations(violations)
 
+	violations = sortResults(violations)
```
simonbaird, line 202: :star_struck: tidy!!

simonbaird (commented):
Awesome!

One nitpick: The PR description is out of date now.

simonbaird:
I'm not sure what the deal is with `Pipelines as Code CI / task-verify-ec-on-pull-request`. Anyone know if we're expecting that to pass?

simonbaird:
I found this:

```
Pipeline tekton-ci/tekton-bundle-builder can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "sanity-label-check": GET https://quay.io/v2/auth?scope=repository%3Aredhat-appstudio-tekton-catalog%2Ftask-sanity-label-check%3Apull&service=quay.io: unexpected status code 502 Bad Gateway:
```

PR #492, opened 2023-03-07
Add --info flag to ec task definition

simonbaird:
The param sets the --info flag and is enabled by default.

The motivation is so that the titles and descriptions for all rules,
including passing checks as well as violations and warnings, are
exposed in the log output of the task where they can (presumably) be
scraped by the UI team and used to produce user-facing displays of
the EC results.

Previously the titles and descriptions were not available in the ec
output, so they needed to be slurped in from a static json file. Now
the information is available directly in the ec output, the json
file should not be needed any more.

JIRA: HACBS-1966

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -74,6 +74,11 @@ spec:
         paths can be provided by using the ":" separator.
       default: ""
 
+    - name: INFO
+      type: string
+      description: Include rule titles and descriptions in the output. Set to "false" to disable it.
+      default: "true"
+
     - name: STRICT
       type: string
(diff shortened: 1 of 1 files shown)


tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -115,6 +115,8 @@ spec:
         - "$(params.REKOR_HOST)"
         # NOTE: The syntax below is required to negate boolean parameters
         - "--strict=$(params.STRICT)"
+        # Include rule descriptions and titles in the output
+        - "--info"
```
simonbaird, line 119: I realized we can do `--info=$(params.INFO)` similar to strict. Will revise the PR.

joejstuart (approved):

zregvart:
Should we have a `--terse` instead?

zregvart (approved):

PR #494, opened 2023-03-08
Accept YAML as input

lcarva:
https://issues.redhat.com/browse/HACBS-1898

### DIFF
--- a/cmd/validate/image_test.go
+++ b/cmd/validate/image_test.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"testing"
 
+	hd "github.com/MakeNowJust/heredoc"
 	conftestOutput "github.com/open-policy-agent/conftest/output"
 	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
 	"github.com/spf13/afero"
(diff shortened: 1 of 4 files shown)


cmd/validate/image_test.go
```diff
@@ -122,6 +123,39 @@ func Test_determineInputSpec(t *testing.T) {
...
+                ---
+                application: app1
+                components:
+                - name: nodejs
+                  containerImage: quay.io/hacbs-contract-demo/single-nodejs-app:877418e
+                - name: petclinic
+                  containerImage: quay.io/hacbs-contract-demo/spring-petclinic:dc80a7f
+                - name: single-container-app
+                  containerImage: quay.io/hacbs-contract-demo/single-container-app:62c06bf
+                `),
```
simonbaird, line 139: Indent looks off, but I'm guessing it's spaces instead of tabs. Consider putting real tabs in maybe?

lcarva: Tabs are not allowed in YAML.

simonbaird: Oh, I did not know that..

cmd/validate/image_test.go
```diff
@@ -79,12 +80,12 @@ func Test_determineInputSpec(t *testing.T) {
 		{
 			name: "faulty ApplicationSnapshot string",
 			arguments: data{
-				input: "/",
+				input: "{",
 			},
-			err: "invalid character '/' looking for beginning of value",
+			err: "unable to parse ApplicatonSnapshot from input: error converting YAML to JSON: yaml: line 1: did not find expected node content",
```
simonbaird, line 85: Typo maybe?

```suggestion
			err: "unable to parse ApplicationSnapshot from input: error converting YAML to JSON: yaml: line 1: did not find expected node content",
```

simonbaird (approved):

cmd/validate/image_test.go
lcarva: Well.. it turns out that `hd.Doc` does remove the leading tabs. Updated.

cmd/validate/image_test.go
lcarva: Yes. Thank you. Updated!

cmd/validate/image_test.go
simonbaird: Ah yes. The unusual case where some tabs followed by some spaces is correct. :smile: .

PR #498, opened 2023-03-09
Don't output passing rule info for skipped rules

simonbaird:
If the passing rule is not part of the specified collection, or is excluded for some other reason, don't include it in the output.

(I think it works, but creating as draft because I expect test coverage is needed.)

### DIFF
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -255,6 +255,16 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) (Check
 			result.Metadata["collections"] = rule.Collections
 		}
 
+		if !c.isResultIncluded(result) {
+			log.Debugf("Skipping result success: %#v", result)
+			continue
+		}
+
+		// Todo maybe: We could also call isResultEffective here for the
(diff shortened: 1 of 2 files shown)


internal/evaluator/conftest_evaluator.go
```diff
@@ -255,6 +255,16 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) (Check
...
 
+		if !c.isResultIncluded(result) {
+			log.Debugf("Skipping result success: %#v", result)
+			continue
+		}
+
+		// Todo maybe: We could also call isResultEffective here for the
+		// success and skip it if the rule is not yet effective. This would
+		// require collecting the effective_on value from the custom annotation
+		// in rule.RuleInfo.
```
zregvart, line 266: I think this is okay, effective on is a cutoff date, one can be in compliance before the date, so it is okay to show that the policy passed.

zregvart (approved):

internal/evaluator/conftest_evaluator.go
simonbaird: Yeah, agreed.

PR #500, opened 2023-03-09
Upgrade to Cosign 2.0.0-(snapshot), Sigstore 1.6.0

zregvart:
With the update the `SIGSTORE_REKOR_PUBLIC_KEY` is no longer part of the Cosign codebase, and the Rekor public key can be now specified via `cosign.CheckOpts`. This opts to provide the Rekor public key if it was specified via `REKOR_PUBLIC_KEY` environment variable. Providing `--rekor-url` without `REKOR_PUBLIC_KEY` containing the PEM encoded Rekor public key results in image signature validation failure.


zregvart:
Created as draft so we can discuss having a snapshot version of Cosign or should we wait for 2.0.1 release. We need the snapshot because Sigstore 1.6.0 is not used in 2.0.0.

features/version.feature
```diff
@@ -13,7 +13,7 @@ Feature: ec cli version subcommand
     OPA                         v.+
     Conftest                    v.+
     Red Hat AppStudio \(shared\)  v.+
-    Cosign                      v.+
+    Cosign                      (v.+|N\/A)
```
zregvart, line 16: This (and below) is odd, we don't get a version when using snapshot dependencies

zregvart:
I'm confident in our test suite, let's merge this :) Happy to address any review issues in followup PRs.

PR #501, opened 2023-03-09
Move and adapt configuration from ec-policies here

zregvart:
Ref. https://issues.redhat.com/browse/HACBS-1807

### DIFF
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -1,0 +1,252 @@
+= Policy Configuration
+
+When `ec` command line requires configuration, e.g. for `ec validate image`
+command, the configuration can be provided as an inline JSON or YAML string,
+such as in this example:
+
+[,bash]
+----
+ec validate image --policy '{
+    "configuration": {
+        "collections": ["minimal"]
+    }
+    "sources": [
+        {
+            "policy": ["oci::https://quay.io/hacbs-contract/ec-release-policy:latest"],
+            "data": ["oci::https://quay.io/hacbs-contract/ec-policy-data:latest"]
+        }
+    ]
+}' ...
+----
+
+The configuration can be provided as a reference to a
+xref:ecc:ROOT:reference.adoc[EnterpriseContractPolicy] Kubernetes custom
+resource by either name (e.g. "my-policy") in the current namespace, or by
+providing namespace and the name separaded by slash (`/`) character, such as in
+this example:
+
+[,bash]
+----
+ec validate image --policy my-namespace/my-policy
+----
+
+If the explicit policy is not provided via the `--policy` parameter, the default
+policy named `default` is loaded from `enterprise-contract-service` namespace of
+the cluster accessed using the current Kubernetes client configuration.
+
+== Including and excluding rules
+
+By default, all rules are included.
+
+You can modify the defaults by setting `config.policy.include` and
+`config.policy.exclude`. Each of those values should be a list of strings.
+
+The strings in the list should be one of the the following:
+
+A "package name"::
+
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -0,0 +1,264 @@
...
+    "configuration": {
+        "collections": ["minimal"]
+    }
+    "sources": [
+        {
+            "policy": ["oci::https://quay.io/hacbs-contract/ec-release-policy:latest"],
+            "data": ["oci::https://quay.io/hacbs-contract/ec-policy-data:latest"]
+        }
+    ]
+} ...'
```
lcarva, line 19: Should it be `}' ...` ?

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -0,0 +1,264 @@
...
+----
+
+As a reference to a xref:ecc:ROOT:reference.adoc[EnterpriseContractPolicy]
```
lcarva, line 22: I think this sentence needs to start with something like `A configuration can also be a reference ...`.

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -0,0 +1,264 @@
+= Policy Configuration
+
+When `ec` command line requires configuration, e.g. for `ec validate image`
+command, the configuration can be provided as an inline JSON or YAML string,
+such as in this example:
+
+[,bash]
```
lcarva, line 7: I wonder if it would be good to mention the workaround we have for when the policy is on a local file, e.g. `ec validate image --policy "$(cat policy.yaml)". At least, until HACBS-1975 is place.

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -0,0 +1,264 @@
...
+"*"::
+
+The `"\*"` acts as a wildcard to match any package. Note that it does not match
+partial names, so you can't specify `"s*"` to match every package starting with
+"s" for example.
+
+You can also specify `"<packagename>.*"` and it works the same as just
+`"<packagename>"` to represent every rule in a package.
+
+== Specifying that individual tests are non-blocking
```
lcarva, line 83: I think this can be omitted now since it's covered by the `:term` options above. `test` is no longer a special case. The "Specifying that certain tests should be non-blocking" below is still relevant and highlights this.

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -0,0 +1,264 @@
...
+=== Including only some rules from a package
+
+You can specify both `include` and `exclude` to pick out just the
+rules you want.
+
+This example specifies that only the `unacceptable_task_bundle` rule from the
+`attestation_task_bundle` package should be included. The other rules in the
+`attestation_task_bundle` would be excluded.
+
+Notice the higher specificity include rule takes precedence over the exclude
```
lcarva, line 160: Wait. I don't know if this is true!

lcarva (approved):
Seems good! I think most of my comments have to do with already existing issues. We can address those later.

docs/modules/ROOT/pages/configuration.adoc
zregvart: I think we should add support for files and feel the pain in the meantime

docs/modules/ROOT/pages/configuration.adoc
lcarva: It is not. Filed https://issues.redhat.com/browse/HACBS-1986 to address it.

lcarva (approved):

PR #502, opened 2023-03-09
validate a directory of definition files

joejstuart:
This change adds support for a directory of definition files as a value for parameter `--file`. It also adds an optional parameter `--namespace` to specify the policy namespace to execute. Without the parameter, conftest evaluates all packages including what is under `lib/`

### DIFF
--- a/internal/definition/validate.go
+++ b/internal/definition/validate.go
@@ -73,7 +73,12 @@ func detectFiles(ctx context.Context, fpath string) ([]string, error) {
 
 	defer file.Close()
 
-	if ok, _ := afero.IsDir(fs, fpath); ok {
+	dir, err := afero.IsDir(fs, fpath)
+	if err != nil {
+		return nil, err
+	}
+
+	if dir {
 		files, err := afero.ReadDir(fs, fpath)
 		if err != nil {
 			return nil, err
@@ -84,7 +89,6 @@ func detectFiles(ctx context.Context, fpath string) ([]string, error) {
 		}
 	} else {
 		defFiles = append(defFiles, fpath)
-
 	}
 
 	return defFiles, nil
--- a/internal/definition/validate.go
+++ b/internal/definition/validate.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
@@ -30,19 +31,19 @@ import (
 	"github.com/hacbs-contract/ec-cli/internal/utils"
 )
 
-var def_file = definition.NewDefinition
+var definitionFile = definition.NewDefinition
 var pathExists = afero.Exists
 
 // ValidatePipeline calls NewPipelineEvaluator to obtain an PipelineEvaluator. It then executes the associated TestRunner
 // which tests the associated pipeline file(s) against the associated policies, and displays the output.
 func ValidateDefinition(ctx context.Context, fpath string, sources []source.PolicySource, namespace []string) (*output.Output, error) {
-	defFiles, err := fileDetection(ctx, fpath)
+	defFiles, err := detectFiles(ctx, fpath)
 	if err != nil {
 		return nil, err
 	}
-	p, err := def_file(ctx, defFiles, sources, namespace)
+	p, err := definitionFile(ctx, defFiles, sources, namespace)
 	if err != nil {
-		log.Debug("Failed to create pipeline definition file!")
+		log.Debug("Failed to create definition file!")
 		return nil, err
 	}
 
@@ -56,7 +57,7 @@ func ValidateDefinition(ctx context.Context, fpath string, sources []source.Poli
 }
 
 // detect if a file or directory was passed. if a directory, gather all files in it
-func fileDetection(ctx context.Context, fpath string) ([]string, error) {
+func detectFiles(ctx context.Context, fpath string) ([]string, error) {
 	fs := utils.FS(ctx)
 	exists, err := pathExists(fs, fpath)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


internal/definition/validate.go
```diff
@@ -18,30 +18,81 @@ package definition
...
+	fileInfo, err := file.Stat()
+	if err != nil {
+		return nil, err
+	}
+
+	// IsDir is short for fileInfo.Mode().IsDir()
+	if fileInfo.IsDir() {
+		files, err := os.ReadDir(fpath)
+		if err != nil {
+			log.Fatal(err)
```
lcarva, line 86: I don't think we should exit the application this far in. Can we just return an error here instead?

internal/definition/validate.go
```diff
@@ -18,30 +18,81 @@ package definition
...
+
+	// This returns an *os.FileInfo type
+	fileInfo, err := file.Stat()
+	if err != nil {
+		return nil, err
+	}
+
+	// IsDir is short for fileInfo.Mode().IsDir()
+	if fileInfo.IsDir() {
+		files, err := os.ReadDir(fpath)
```
lcarva, line 84: I suppose we don't want to recursively read the directory?

internal/definition/validate.go
```diff
@@ -18,30 +18,81 @@ package definition
...
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		for _, f := range files {
+			defFiles = append(defFiles, fmt.Sprintf("%s/%s", fpath, f.Name()))
```
lcarva, line 90: https://pkg.go.dev/path/filepath#Join

internal/definition/validate_test.go
```diff
@@ -101,7 +108,8 @@ func Test_ValidatePipeline(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			def_file = tt.defFunc
-			output, err := ValidateDefinition(ctx, tt.fpath, []source.PolicySource{})
+			pathExists = tt.exists
```
lcarva, line 111: The nice thing about using afero, is that you can load an in-memory filesystem. Rather than mocking pathExists, you can change the test to actually create the file, or not. That's what most of the other tests do. It's the reason we use `utils.FS(ctx)`.

internal/definition/validate.go
joejstuart: I thought of that. I don't know. It can be troublesome. I guess I can be convinced if you feel strongly about it.

lcarva: My thinking is that all the repos I've seen that have multiple task definitions do so in different directories. It's probably good enough for now. This can be a feature added later on if its need become obvious.

internal/evaluation_target/definition/definition.go
```diff
@@ -18,35 +18,22 @@ package definition
...
-func NewDefinition(ctx context.Context, fpath string, sources []source.PolicySource) (*Definition, error) {
-	fs := utils.FS(ctx)
-	exists, err := pathExists(fs, fpath)
-	if err != nil {
-		return nil, err
-	}
-	if !exists {
-		return nil, fmt.Errorf("fpath '%s' does not exist", fpath)
-	}
+func NewDefinition(ctx context.Context, fpath []string, sources []source.PolicySource, namespace []string) (*Definition, error) {
```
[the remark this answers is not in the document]
joejstuart: It does do that. I can add a test though. The thing is, since we include the policy/lib directory, conftest evaluates the definition file against all namespaces under policy/. So you end up with this. On one hand it's great because you can see all failed and successful policies/packages, but you also see conftest runs that don't matter. To address this, I can merge all the namespaces and drop the `namespace` field or we can require the `namespace` param, but support something like `policy.pipeline` which includes everything under pipeline. 

```yaml
[
 {
    "filename": "../build-definitions/task/buildah/0.1/buildah.yaml",
    "violations": [],
    "warnings": [],
    "success": true,
    "namespace": "lib.image"
  },
  {
    "filename": "../build-definitions/task/buildah/0.1/buildah.yaml",
    "violations": [
      {
        "msg": "Unexpected kind 'Task'",
        "metadata": {
          "code": "basic.unexpected_kind",
          "description": "A sanity check to confirm the input data has the kind \"Pipeline\"",
          "effective_on": "2022-01-01T00:00:00Z",
          "title": "Input data has unexpected kind"
        }
      }
    ],
    "warnings": [],
    "success": false,
    "namespace": "policy.pipeline.basic"
  }
]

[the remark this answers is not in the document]
Comment (joejstuart) [reply internal/evaluation_target/definition/definition.go:36]: I'll add. I think allowing conftest to run against `lib` feels like the wrong thing.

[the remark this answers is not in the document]
Comment (lcarva) [reply internal/evaluation_target/definition/definition.go:36]: `validate image` runs against all namespaces and we had to do something special to handle this. Check it out to see how it was addressed.

**Review by zregvart — COMMENTED** [review]
Looks okay, have a look at what Luiz mentioned. As a followup and for parity we might want to report the successes as well, this is a bit of work and requires parsing the Rego files to gather the rules and then artificially add them to the results...

File: internal/definition/validate.go [hunk 6]
```diff
@@ -18,30 +18,81 @@ package definition
...
+	"github.com/spf13/afero"
 
 	"github.com/hacbs-contract/ec-cli/internal/evaluation_target/definition"
 	"github.com/hacbs-contract/ec-cli/internal/output"
 	"github.com/hacbs-contract/ec-cli/internal/policy/source"
+	"github.com/hacbs-contract/ec-cli/internal/utils"
 )
 
 var def_file = definition.NewDefinition
+var pathExists = afero.Exists
```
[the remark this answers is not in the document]
joejstuart: To mock it. https://github.com/hacbs-contract/ec-cli/pull/502/files#diff-b267ad03361ef7b25c45e9d67bc38806c84964343614451920e5e231dba8fa7aR111

internal/evaluation_target/definition/definition.go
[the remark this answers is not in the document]
joejstuart: the report output has been updated. https://github.com/hacbs-contract/ec-cli/pull/502/commits/0b56973ec2c02d3a93394915d9c7a476c21ba588

zregvart (approved):
LGTM!

internal/definition/validate_test.go
joejstuart: Replaced OS commands with Afero. Very neat!

PR #504, opened 2023-03-10
Attestation type wrapping SLSA v0.2 Provenance

zregvart:
This unifies the parsing of in-toto attestations containing SLSA v0.2 Provenance payloads. This is the only supported attestation format with `github.com/hacbs-contract/ec-cli/internal/attestation.SLSAProvenanceFromLayer` as the only entry point.

Future refinements might introduce a layer of indirection that for a given source (currently OCI layer, other sources could be possible) returns differing payload types. For simplicity, this is not done here.

Ref. https://issues.redhat.com/browse/HACBS-1768


internal/attestation/attestation.go
```diff
@@ -0,0 +1,38 @@
...
+package attestation
+
+import (
+	"github.com/hacbs-contract/ec-cli/internal/output"
+	e "github.com/hacbs-contract/ec-cli/pkg/error"
+)
+
+var (
+	AT001 = e.NewError("AT001", "No attestation found", e.ErrorExitStatus)
+	AT002 = e.NewError("AT002", "Mailformed attestation data", e.ErrorExitStatus)
```
lcarva, line 26: `s/Mailformed/Malformed`

internal/attestation/attestation.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2022 Red Hat, Inc.
```
lcarva, line 1: We should probably use the current year for new files.

internal/attestation/slsa_provenance_02_test.go
```diff
@@ -0,0 +1,337 @@
...
+		statement in_toto.ProvenanceStatementSLSA02
+		err       e.Error
+	}{
+		{
+			name: "nil layer",
+			err:  AT001,
+		},
+		{
+			name:  "media type error",
+			layer: &mockLayer{&mock.Mock{}},
```
lcarva, line 95: This is the same for all cases, maybe just have it done in the sub-test body? Same for `TestEntitySignature`.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -199,11 +198,12 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
 	// Extract the signatures from the attestations here in order to also validate that
 	// the signatures do exist in the expected format.
 	for _, att := range attestations {
-		signatures, err := entitySignatureFromAttestation(att)
+		sp, err := attestation.SLSAProvenanceFromLayer(att)
```
lcarva, line 201: A further step would be to make it so `SLSAProvenanceFromLayer` is only called once per attestation for a particular image. Maybe by stuffing the results somewhere in the `ApplicationSnapshotImage` object?

This could be done at a later time if execution time needs to be optimized.

pkg/error/error.go
```diff
@@ -90,6 +91,21 @@ func (e ecError) Error() string {
 	return fmt.Sprintf("%s: %s, at %s:%d, caused by: %s", e.code, e.message, e.file, e.line, e.cause)
 }
 
+// Alike teturns true if two ecErros share the same code and cause
```
lcarva, line 94: `s/teturns/returns`

lcarva (approved):

zregvart:
I think this should be okay to merge now. Happy to address anything in followup PRs.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
lcarva: Thanks for going the extra mile to incorporate this! :pray:

PR #506, opened 2023-03-10
ADD support for tekton-lint validation

robnester-rh:
This commit adds support for tekton-lint in the Makefile as well for github workflows when doing a pull request or push to the main branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,16 @@ lint-fix: ## Fix linting issues automagically
 #	@go run -modfile tools/go.mod ./internal/lint -fix $$(go list ./... | grep -v '/acceptance/')
 	@go run -modfile tools/go.mod github.com/daixiang0/gci write -s standard -s default -s "prefix(github.com/hacbs-contract/ec-cli)" .
 
+.PHONY: tekton-lint
+tekton-lint: ## Run tekton-lint for 'tasks' subdirectory.
+ifeq ($(GITHUB_ACTIONS),) # If not running as a Github action, check if tekton-lint is installed. If not, provide link.
+	@which tekton-lint &> /dev/null || (echo "tekton-lint doesn't seem to be installed. Please see https://github.com/IBM/tekton-lint for installation instructions." && exit 1)
+endif
+# We execute tekton-lint for all yaml files contained within the tasks subdirectory, it's smart enough to ignore non-Tekton yaml files.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -112,6 +112,15 @@ lint-fix: ## Fix linting issues automagically
 #	@go run -modfile tools/go.mod ./internal/lint -fix $$(go list ./... | grep -v '/acceptance/')
 	@go run -modfile tools/go.mod github.com/daixiang0/gci write -s standard -s default -s "prefix(github.com/hacbs-contract/ec-cli)" .
 
+.PHONY: tekton-lint
+tekton-lint: ## Check for tekton-lint, if not found provide link. Otherwise, execute tekton-lint for all yaml in the tasks subdirectory, it's smart enough to ignore non-Tekton yaml files. Any warnings are currently considered failures.
+    ifdef $(GITHUB_ACTIONS)
```
lcarva, line 117: Is this check inverted? It would be helpful to do it when running locally, and not on GitHub. (Or maybe I'm misreading this.)

lcarva (approved):

Makefile
robnester-rh: Fixed!

Makefile
```diff
@@ -112,6 +112,16 @@ lint-fix: ## Fix linting issues automagically
 #	@go run -modfile tools/go.mod ./internal/lint -fix $$(go list ./... | grep -v '/acceptance/')
 	@go run -modfile tools/go.mod github.com/daixiang0/gci write -s standard -s default -s "prefix(github.com/hacbs-contract/ec-cli)" .
 
+.PHONY: tekton-lint
+tekton-lint: ## Run tekton-lint for 'tasks' subdirectory.
```
zregvart, line 116: Should have a list of dependencies, something like:
```suggest
tekton-lint: $(wildcard tasks/*/*/*/*.yaml)
```

The `tekton-lint` target should be included in the `lint` target.

zregvart (commented):
We already have a [lint](https://github.com/hacbs-contract/ec-cli/blob/main/.github/workflows/lint.yaml) workflow and a [`lint` target](https://github.com/hacbs-contract/ec-cli/blob/094c331ffcdef80eb9284ccda7024fb5103d04bd/Makefile#L95) in the Makefile. It would be much better to run a single workflow/target.

PR #513, opened 2023-03-14
Add an ec inspect policy-data command

simonbaird:


cmd/inspect/inspect.go
```diff
@@ -28,4 +28,5 @@ func init() {
 		Short: "Inspect policy rules",
 	}
 	InspectCmd.AddCommand(inspectPolicyCmd())
+	InspectCmd.AddCommand(inspectPolicyDataCmd())
```
zregvart, line 31: When I imagined this I was thinking it would be part of the same `ec inspect policy` command. I guess this works as well.

cmd/inspect/inspect_policy_data.go
```diff
@@ -0,0 +1,159 @@
...
+		`),
+
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !slices.Contains(validFormats, outputFormat) {
+				return fmt.Errorf("invalid value for --output '%s'. accepted values: %s", outputFormat, strings.Join(validFormats, ", "))
+			}
+
+			if destDir == "" {
+				workDir, err := utils.CreateWorkDir(afero.NewOsFs())
```
zregvart, line 75: A `defer` to cleanup the workdir could be nice. This should take the fs from context via `util.FS`

cmd/inspect/inspect_policy_data.go
```diff
@@ -0,0 +1,159 @@
...
+
+					if d.IsDir() {
+						return nil
+					}
+
+					fileExt := strings.ToLower(filepath.Ext(path))
+					if slices.Contains([]string{".yaml", ".yml", ".json"}, fileExt) {
+						log.Debugf("Found data file %s", path)
+
+						contents, err := afero.ReadFile(afero.NewOsFs(), path)
```
zregvart, line 107: Use `util.FS` here to get the filesystem implementation

cmd/inspect/inspect_policy_data.go
```diff
@@ -0,0 +1,159 @@
...
+					if readErr != nil {
+						return readErr
+					}
+
+					if d.IsDir() {
+						return nil
+					}
+
+					fileExt := strings.ToLower(filepath.Ext(path))
+					if slices.Contains([]string{".yaml", ".yml", ".json"}, fileExt) {
```
zregvart, line 104: We could try to read any file here and if `yaml.Unmarshal` succeeds use its content

cmd/inspect/inspect_policy_data.go
```diff
@@ -0,0 +1,159 @@
...
+
+						yamlData := make(map[string]interface{})
+
+						// Should work for both yaml and json
+						err = yaml.Unmarshal(contents, &yamlData)
+						if err != nil {
+							return err
+						}
+
+						// If we find a duplicate key, throw an error. I think that is what Conftest would do.
```
zregvart, line 120: There is a [bundle.Merge](https://pkg.go.dev/github.com/open-policy-agent/opa@v0.50.0/bundle#Merge) function in OPA, I wonder if we should use that. The downside is that we would have two ways of loading data. I do assume that Conftest delegates to this in the end. Something like this might work (from the top of my head):

```go
// in the loop over bundles
var complete *bundle.Bundle
r := bundle.NewCustomReader(bundle.NewDirectoryLoader(dir))
b, err := r.Read()

if complete == nil {
	complete = b
} else {
	complete.Merge(b)
}
// dump in YAML
yaml.Marshal(complete.Data)
```

zregvart (commented):
Neat

cmd/inspect/inspect.go
simonbaird: Actually I flip-flopped between a few options here, since `ec inspect policy --data-source <sourceurl>` could also make sense, or even some kind of autodetection of the source "kind", but I decided on a separate command, partly to avoid handling the possibility of the user specifying both policy and data source urls in one `ec inspect` command.

cmd/inspect/inspect_policy_data.go
simonbaird: I think actually Conftest does its own style of data loading different from OPA. In particular, for OPA the directory structure is significant, so you get nested keys for files in nested directories, but Conftest just flattens everything IIUC. I was poking around [here](https://github.com/open-policy-agent/conftest/blob/master/builtins/parse_config.go) trying to understand this.

That said I might experiment with this and see what we get.

zregvart: We don't want to diverge from Conftest, that would be confusing and work in vain...

cmd/inspect/inspect_policy_data.go
simonbaird: Done.

cmd/inspect/inspect_policy_data.go
simonbaird: Done.

cmd/inspect/inspect_policy_data.go
simonbaird: I decided against it for this PR, but we could consider if for the future maybe. There's also .xml, .ini, and probably a bunch of other formats that conftest would probably parse.

cmd/inspect/inspect_policy.go
```diff
@@ -82,27 +81,32 @@ func inspectPolicyCmd() *cobra.Command {
 				return fmt.Errorf("invalid value for --output '%s'. accepted values: %s", outputFormat, strings.Join(validFormats, ", "))
 			}
 
+			ctx := cmd.Context()
+			afs := utils.FS(ctx)
```
zregvart, line 85: I've added this on https://github.com/hacbs-contract/ec-cli/pull/514, noting as there will be a conflict.

internal/utils/helpers.go
```diff
@@ -73,6 +75,16 @@ func CreateWorkDir(fs afero.Fs) (string, error) {
...
 
+// CleanupWorkDir removes all files in a directory
+// Return nothing so we can use it with defer more conveniently
+func CleanupWorkDir(fs afero.Fs, path string) {
+	err := fs.RemoveAll(path)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1) // I guess..?
+	}
+}
```
zregvart, line 86: Why not:

```suggestion
func CleanupWorkDir(fs afero.Fs, path string) error {
	return fs.RemoveAll(path)
}
```

Or even:
```suggestion
func CleanupWorkDir(fs afero.Fs, path string) {
	_ := fs.RemoveAll(path)
}
```

zregvart: Any returns from defer functions are [ignored](https://go.dev/ref/spec#Defer_statements)

simonbaird: But I think maybe I get a lint-fix error for ignoring an error.

zregvart (approved):
Merge already

PR #514, opened 2023-03-15
Add `--policy` parameter to `ec inspect policy`

zregvart:
The parameter abides with the same semantics as does the same-named parameter in `ec validate image`. The sources found in the policy configuration are appended to the sources already provided via `--source` parameter(s).

Ref. https://issues.redhat.com/browse/HACBS-1731

### DIFF
--- a/cmd/inspect/inspect_policy.go
+++ b/cmd/inspect/inspect_policy.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/hacbs-contract/ec-cli/internal/opa"
 	opaRule "github.com/hacbs-contract/ec-cli/internal/opa/rule"
+	"github.com/hacbs-contract/ec-cli/internal/policy"
 	"github.com/hacbs-contract/ec-cli/internal/policy/source"
 	"github.com/hacbs-contract/ec-cli/internal/utils"
 )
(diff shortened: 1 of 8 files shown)


cmd/inspect/inspect_policy.go
```diff
@@ -126,6 +148,7 @@ func inspectPolicyCmd() *cobra.Command {
 		},
 	}
 
+	cmd.Flags().StringVarP(&policyRef, "policy", "p", "", "reference to the policy configuration, either EnterpriseContractPolicy Kubernetes custom resource reference [<namespace>/]<name>, or inline JSON or YAML of the `spec` part")
 	cmd.Flags().StringArrayVarP(&sourceUrls, "source", "s", []string{}, "policy source url. multiple values are allowed")
```
simonbaird, line 152: Should it say, "Likely to be ignored when using the --policy flag"?

zregvart: It's not, one can use `--policy` and `--source` if they so desire. Thought it would be useful in "what if I add this source" use case.

lcarva (approved):

zregvart:
Made `--policy` and `--source` parameters mutually exclusive.

PR #520, opened 2023-03-16
Add `--snapshot` parameter to `ec validate image`

zregvart:
Includes #515, review/merge it first.

Includes:

[Refactor to pass single parameter: Input](https://github.com/hacbs-contract/ec-cli/commit/ef6d123b5be98e5dcf9dea62ecafe18b3a4d0063)

Easy to make a mistake and confuse the order of parameters of the same type; best introduce a type to hold the parameters.

[Merge all provided inputs](https://github.com/hacbs-contract/ec-cli/commit/24d49edc3bedae1ac60d3b5e8d473ea2f29c0bd6)

Might be a nice feature.

[Add --snapshot parameter to ec validate image](https://github.com/hacbs-contract/ec-cli/commit/cc6883c2d3c8730df79d0409c14d574861268ca4)

This allows specifying a cluster reference to an AppStudio Snapshot via `--snapshot` parameter to `ec validate image`.

[Add snapshot to output](https://github.com/hacbs-contract/ec-cli/commit/b581d7f48e6ab48f19498eee7b4a0cf807c32a2c)

If the Application Snapshot reference was given via `--snapshot` it is now included in the output.

Ref https://issues.redhat.com/browse/HACBS-1882


internal/applicationsnapshot/input.go
```diff
@@ -17,61 +17,140 @@
...
+const unnamed = "Unnamed"
+
+type Input struct {
+	File     string
+	JSON     string
+	Image    string
+	Snapshot string
+}
+
+type snapshots struct {
```
lcarva, line 43: Use the singular form `snapshot`?

internal/applicationsnapshot/input.go
```diff
@@ -17,61 +17,140 @@
...
+				s.Components[i].Name = c.Name
+			} else {
+				images[c.ContainerImage] = c.Name
+				s.Components = append(s.Components, c)
+			}
+		}
+	}
+}
+
+func DetermineInputSpec(ctx context.Context, input Input) (*app.SnapshotSpec, error) {
```
lcarva, line 80: IMO, the merging functionality is confusing from a user's perspective. The user must proactively know about it, and understand the precedence logic. I'd be in favor of just throwing an error if the user tries to use more than one source.

This comment is highly subjective, so feel free to ignore it :)

But if we do keep the merge logic, let's make sure it's clear how it behaves from the ec-cli help.

lcarva (approved):

internal/applicationsnapshot/input.go
zregvart: @hacbs-contract/devs thoughts?

lcarva: Let's keep it if no one else has an opinion.

PR #521, opened 2023-03-16
Add effective_on to success results and skip if not effective

mbestavros:
This code _should_ work based on the format of the `custom` field in `Annotations` as described in `policy/lib/time_test.rego` within `ec-policies`. However, it seems like `golden-image` doesn't expose the correct metadata, so another test case might be needed. (Or I'm just wrong about how to grab `effective_on` data.)

Ref: https://issues.redhat.com/browse/HACBS-1985

### DIFF
--- a/internal/opa/rule/rule.go
+++ b/internal/opa/rule/rule.go
@@ -39,6 +39,16 @@ func description(a *ast.AnnotationsRef) string {
 	return a.Annotations.Description
 }
 
+func effectiveOn(a *ast.AnnotationsRef) string {
+	if a == nil || a.Annotations == nil || a.Annotations.Custom == nil {
+		return ""
+	}
+	if value, ok := a.Annotations.Custom["effective_on"].(string); ok {
+		return value
+	}
+	return ""
+}
+
 func lastTerm(a *ast.AnnotationsRef) string {
 	if a == nil || len(a.Path) == 0 {
 		return ""
@@ -213,6 +223,7 @@ type Info struct {
 	Collections      []string
 	Description      string
 	DocumentationUrl string
+	EffectiveOn      string
 	Kind             RuleKind
 	Package          string
 	ShortName        string
@@ -226,6 +237,7 @@ func RuleInfo(a *ast.AnnotationsRef) Info {
 		Collections:      collections(a),
 		Description:      description(a),
 		DocumentationUrl: documentationUrl(a),
+		EffectiveOn:      effectiveOn(a),
 		Kind:             kind(a),
 		Package:          packageName(a),
 		ShortName:        shortName(a),
--- a/internal/opa/rule/rule_test.go
+++ b/internal/opa/rule/rule_test.go
@@ -233,6 +233,62 @@ func TestShortName(t *testing.T) {
 	}
 }
 
+func TestEffectiveOn(t *testing.T) {
+	cases := []struct {
+		name       string
+		annotation *ast.AnnotationsRef
+		expected   string
+	}{
+		{
+			name:       "empty",
+			annotation: nil,
+			expected:   "",
+		},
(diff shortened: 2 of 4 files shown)


lcarva (approved):
Unit tests would be nice so we don't accidentally break this in the future :grin:

simonbaird:
Yeah, I think test coverage is needed.

It's probably easier to piggyback on the acceptance test for #498.

(For #498 I decided a unit test was going to be too hard, (I was looking in TestConftestEvaluatorIncludeExclude in internal/evaluator/conftest_evaluator_test.go wondering how to adjust the mocking so that successes were included in the test).

simonbaird:
Possible approach for testing:

- Create a new rego file called something like `acceptance/examples/future_pass.rego`
- This rego file should have a single rule that always passes with annotations like the rules `filtering.rego`.
-  It should have effective_on date from the distant future.
- Then in `features/validate_image.feature`, under `Scenario: policy rule filtering for successes` include that new rego file in the dummy git repo. Because it's a pass in the future, the new rule should not appear in the output, so the output under "output should contain" would remain unchanged. (You could confirm that for yourself by tweaking the effective_on date to the past, which would make that acceptance test fail.)

simonbaird:
Bonus tip for running acceptance tests locally: I like to do this to save time:

```
sed -i '1i@focus' features/validate_image.feature # put @focus at top of feature you're working on (or just add it manually)
make focus-acceptance
```

Remember not to accidentally commit the `@focus` diff.

mbestavros:
I added some basic unit tests to `rule_test.go` and `conftest_evaluator_test.go`. I can look more into Rego tests if that would be better.

internal/opa/rule/rule_test.go
```diff
@@ -233,6 +233,62 @@ func TestShortName(t *testing.T) {
...
+			annotation: annotationRef(heredoc.Doc(`
+				package a
+				# METADATA
+				# custom:
+				#   hmm: 14
+				deny() { false }`)),
+			expected: "",
+		},
+		{
+			name: "with short_name annotation",
```
lcarva, line 274: Should this say "with effective_on annotation" ?

internal/opa/rule/rule_test.go
```diff
@@ -233,6 +233,62 @@ func TestShortName(t *testing.T) {
...
+				# METADATA
+				# custom:
+				#   effective_on: '2022-01-01T00:00:00Z'
+				deny() { true }`)),
+			expected: "2022-01-01T00:00:00Z",
+		},
+	}
+
+	for i, c := range cases {
+		t.Run(fmt.Sprintf("[%d] - %s", i, c.name), func(t *testing.T) {
```
lcarva, line 286: FYI, usually `t.Run(c.name, func...)` is enough.

lcarva (commented):
Tests are failing due to a nil-pointer exception. Might wanna check that out :wink:

internal/opa/rule/rule.go
```diff
@@ -39,6 +39,13 @@ func description(a *ast.AnnotationsRef) string {
 	return a.Annotations.Description
 }
 
+func effectiveOn(a *ast.AnnotationsRef) string {
+	if value, ok := a.Annotations.Custom["effective_on"].(string); ok {
```
zregvart, line 43: This needs something like, to prevent the nil reference error:
```suggestion
	if a == nil {
		return ""
	}
	if value, ok := a.Annotations.Custom["effective_on"].(string); ok {
```

mbestavros: Thanks! I'll try that - the `shortName` function has a similar condition.

mbestavros:
@lcarva I don't seem to have merge permission:

image

PR #522, opened 2023-03-16
Remove support for tracking required tasks

lcarva:


features/track_bundle.feature
```diff
@@ -48,65 +37,46 @@ Feature: track bundles
...
     task-bundles:
       ${REGISTRY}/acceptance/bundle:
-        - digest: sha256:486981f90bd8cca5586b7d73d5c5ce7e1f29d174f0b5fe027b80730612f37155
+        - digest: sha256:21040e5abd0e077b7344574473468beff02cd6cc66dc464acb3c6b4be5bb82af
           effective_on: "${TODAY_PLUS_30_DAYS}"
           tag: tag
 
     """
 
   Scenario: Replace data bundle in OCI registry
```
lcarva, line 51: I made a fix to this scenario. It now uses two bundles, each containing a different set of tasks and pipelines. This way, we can be sure that `--replace` is working by adding to the existing data.

joejstuart (approved):

PR #523, opened 2023-03-16
A convenient way to test a single feature or scenario

simonbaird:


Makefile
```diff
@@ -85,10 +85,21 @@ acceptance: ## Run acceptance tests
 
 # Add @focus above the feature you're hacking on to use this
 .PHONY: focus-acceptance
-focus-acceptance: ## Run acceptance tests with @focus tag
```
lcarva, line 88: Intentional removal?

Makefile
```diff
@@ -85,10 +85,21 @@ acceptance: ## Run acceptance tests
...
+focus-acceptance:
 	@$(MAKE) build
 	@cd acceptance && go test -tags=acceptance . -args -tags=@focus
 
+feature-%: ## Test a single feature, e.g. make feature-validate_image
```
lcarva, line 92: Missing `.PHONY`?

lcarva (approved):
:tada: neat!

Makefile
zregvart, line 92: Should this depend on `build` target?

zregvart (commented):
Just to note, this should work:

```shell
cd acceptance
go test -test.run 'TestFeatures/Scenario_name'
```

For example:
```shell
cd acceptance
go test -test.run 'TestFeatures/JSON_output'
```
will run [this scenario](https://github.com/hacbs-contract/ec-cli/blob/c8b2b2694d5e249b298a0164c8769f6be13f8ecc/features/version.feature#L31)

I'd rather see something going from `git status` to test selection for acceptance tests. Also we could be doing much nicer things in `acceptance/acceptance.go` in terms of selection...

Makefile
simonbaird: Yeah. The reasoning is that a dev would use one of the new targets and have little need to run this one directly any more.

Makefile
simonbaird: I wasn't sure if it would work with %, but perhaps it does.

simonbaird: Yeah, thanks.

Makefile
```diff
@@ -113,7 +124,7 @@ lint-fix: ## Fix linting issues automagically
 	@go run -modfile tools/go.mod github.com/daixiang0/gci write -s standard -s default -s "prefix(github.com/hacbs-contract/ec-cli)" .
 
 .PHONY: tekton-lint
-tekton-lint: ## Run tekton-lint for 'tasks' subdirectory.
+tekton-lint: ## Run tekton-lint for 'tasks' subdirectory
```
simonbaird, line 127: Drive-by formatting consistency fix (for `make help`)

Makefile
lcarva: Hmm ok. It seems odd to remove the docs for it instead of just removing the command.

simonbaird: I'll put the docs back since maybe it's still useful by itself.

Makefile
simonbaird: Afact `.PHONY: feature-%` doesn't do anything useful.

simonbaird:
I might scrap the whole @focus hackery if the `-test.run` flag works like @zregvart mentioned. Moving to draft.

zregvart (approved):

PR #524, opened 2023-03-16
Support passing a json or yaml string as file param to validate definition

joejstuart:
- Support json or yaml string as `--file` param
- Remove extra afero mock

### DIFF
--- a/internal/definition/validate.go
+++ b/internal/definition/validate.go
@@ -35,7 +35,7 @@ var definitionFile = definition.NewDefinition
 // ValidatePipeline calls NewPipelineEvaluator to obtain an PipelineEvaluator. It then executes the associated TestRunner
 // which tests the associated pipeline file(s) against the associated policies, and displays the output.
 func ValidateDefinition(ctx context.Context, fpath string, sources []source.PolicySource, namespace []string) (*output.Output, error) {
-	defFiles, err := detectFiles(ctx, fpath)
+	defFiles, err := detectInput(ctx, fpath)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


internal/definition/validate.go
```diff
@@ -93,3 +98,31 @@ func detectFiles(ctx context.Context, fpath string) ([]string, error) {
 
 	return defFiles, nil
 }
+
+// detect if the string is json
+func detectJson(msg string) bool {
```
zregvart, line 103: We have similar logic in a couple of places, perhaps a good candidate to place in the `internal/utils`?

```suggest
func IsJson(text string) bool
```

internal/definition/validate.go
```diff
@@ -93,3 +98,31 @@ func detectFiles(ctx context.Context, fpath string) ([]string, error) {
...
+		return nil, err
+	}
+
+	filePath := filepath.Join(tmpDir, fileName)
+	err = afero.WriteFile(fs, filePath, []byte(msg), 0644)
+	if err != nil {
+		return nil, err
+	}
+	return []string{filePath}, nil
+}
```
zregvart, line 128: Perhaps something like:
```suggestion
func writeTempFile(ctx context.Context, data string) (string, error) {
	fs := utils.FS(ctx)
	file, err := fs.TempFile("", "*")
	if err != nil {
		return "", err
	}
	path := file.Name()
	if _, err := file.WriteString(data); err != nil {
		_ := fs.Remove(path)
		return "", err
	}
	return path, nil
}
```

internal/utils/helpers.go
```diff
@@ -98,3 +100,16 @@ func FS(ctx context.Context) afero.Fs {
 func WithFS(ctx context.Context, fs afero.Fs) context.Context {
 	return context.WithValue(ctx, fsKey, fs)
 }
+
+func GenerateRandomString(length int) (string, error) {
```
zregvart, line 104: Guess we might not need this?

zregvart:
Would we also want to support YAML?

lcarva:
> Would we also want to support YAML?

I think at this point, we should do this by default. We'd need a reason **not** to do it.

internal/definition/validate.go
joejstuart: Good point. I'm sure that can be accommodated.

internal/definition/validate.go
```diff
@@ -55,44 +55,73 @@ func ValidateDefinition(ctx context.Context, fpath string, sources []source.Poli
 }
 
 // detect if a file or directory was passed. if a directory, gather all files in it
-func detectFiles(ctx context.Context, fpath string) ([]string, error) {
+// the order is file lookup, json lookup then yaml
+func detectInput(ctx context.Context, fpath string) ([]string, error) {
+	log.Error("trying to detect input as JSON")
```
zregvart, line 60: Not sure about logging at this level, perhaps the only error we should print is when we can't detect/load from a path?

zregvart (approved):
Straightforward enough, not sure about the error logging level

internal/definition/validate.go
joejstuart: I wasn't in love with it either. Was just trying to find a way to communicate to the user what was going on. I moved the logging to after the validation attempt and to debug. wdyt?

```golang
if utils.IsJson(fpath) {
    log.Debug("valid JSON found for definition file")
    return definitionFromString(ctx, fpath)
}
log.Debug("unable to detect input as JSON")
```

lcarva (approved):

internal/definition/validate.go
```diff
@@ -55,41 +55,81 @@ func ValidateDefinition(ctx context.Context, fpath string, sources []source.Poli
...
+	log.Debug("unable to detect input as YAML")
+
+	files, err := fileLookup(ctx, fpath)
+	if len(files) > 0 {
+		log.Debug("valid file path found for definition file")
+		return files, nil
+	}
+	// just log this error. it could be an actual os error or "file not found".
+	// either way, move on
+	if err != nil {
```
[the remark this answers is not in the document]
joejstuart: Good points. I think this is the best I can do with it. https://github.com/joejstuart/ec-cli/commit/d9fb3330ad2ec583b99f0cc04fd85da34aa15852

PR #540, opened 2023-03-20
Fix error when fetching oci bundle with scheme

lcarva:
User reported error:

error validating image ... of component ...: client get:
error downloading 'https://quay.io/hacbs-contract/ec-release-policy:latest': reference:
invalid reference: invalid repository

The issue is due to a recent change in the conftest downloader. Since a fix is not yet merged, this commit picks up the fix from my fork.


lcarva:
Another approach could be to just remove the scheme on the ec-cli code before sending it to conftest.

joejstuart (approved):

PR #544, opened 2023-03-21
Remove `fetch commit` command

lcarva:
https://issues.redhat.com/browse/HACBS-2019

My argument for removing this command is as follow:

1. The command is currently broken. This is a clear indicator that this is dead functionality. Here's the error I get:
```
🐚 ec fetch commit  --image-ref quay.io/redhat-appstudio/ec-golden-image:latest --public-key k8s://tekton-chains/public-key
Error: 1 error occurred:
        * component: {Name:Unnamed ContainerImage:quay.io/redhat-appstudio/ec-golden-image:latest} - no matching signatures:
rekor client not provided for online verification
```

2. The code dilutes the meaning of authorization. When a user adds `Signed-off-by:` to the commit message, they are not stating an authorization of any sort. In fact, the [meaning is relevant to the repository](https://git-scm.com/docs/git-commit):
> The meaning of a signoff depends on the project to which you’re committing. For example, it may certify that the committer has the rights to submit the work under the project’s license or agrees to some contributor representation, such as a Developer Certificate of Origin. (See http://developercertificate.org/ for the one used by the Linux kernel and Git projects.) Consult the documentation or leadership of the project to which you’re contributing to understand how the signoffs are used in that project.

3. We have moved away from authorization as a git commit towards the idea that an authorization is something dynamic. It should be possible to authorize, or revoke authorization, without having to make any changes to the container image. An authorization embedded in a git commit prevents this.


lcarva:
@hacbs-contract/devs, let me know what you think of this. Let's get consensus before merging, or dropping, this.

mbestavros (commented):
The reasoning makes sense to me.

Without knowing the full `ec-cli` picture, do we anticipate needing signed commit verification at any point in the future, even if it's not for this specific command?

joejstuart:
The concept of authorization or signoff has changed since this was done and it's not needed. Is there any reason to keep the code that extracts the commit sha from an attestation and gets the commit? If not, I would vote to remove it all.

lcarva:
> do we anticipate needing signed commit verification at any point in the future, even if it's not for this specific command?

I think if we do, the existing code does little towards that.

> Is there any reason to keep the code that extracts the commit sha from an attestation and gets the commit (I don't think so)? If not, I would vote to remove it all.

I'm hesitant to keep code just in case we need it later one. If it comes to that, we can always retrieve it from the git history.

internal/image/authorization.go
```diff
@@ -20,14 +20,7 @@ import (
...
 	"fmt"
-	"net/mail"
-	"os"
-	"regexp"
-	"strings"
-
-	"github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/go-git/go-git/v5/plumbing/object"
+
```
zregvart, line 23: This ruined otherwise a perfectly lovely PR

zregvart (approved):

PR #552, opened 2023-03-23
Enable gosec linter and fix linting issues

zregvart:
Ref. https://issues.redhat.com/browse/HACBS-1940

### DIFF
--- a/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
+++ b/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
@@ -88,7 +88,7 @@ func NewApplicationSnapshotImage(ctx context.Context, url string, p policy.Polic
 	for _, sourceGroup := range p.Spec().Sources {
 		// Todo: Make each fetch run concurrently
 		log.Debugf("Fetching policy source group '%s'", sourceGroup.Name)
-		policySources, err := fetchPolicySources(&sourceGroup)
+		policySources, err := fetchPolicySources(sourceGroup)
 		if err != nil {
 			log.Debugf("Failed to fetch policy source group '%s'!", sourceGroup.Name)
 			return nil, err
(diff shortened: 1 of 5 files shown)


internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -111,7 +111,7 @@ func NewApplicationSnapshotImage(ctx context.Context, url string, p policy.Polic
 }
 
 // fetchPolicySources returns an array of policy sources
-func fetchPolicySources(s *ecc.Source) ([]source.PolicySource, error) {
+func fetchPolicySources(s ecc.Source) ([]source.PolicySource, error) {
```
lcarva, line 114: What's the reason here? Don't pass by reference if not needed?

internal/evaluator/conftest_evaluator.go
```diff
@@ -219,7 +219,8 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) (Check
 		exceptions := []output.Result{}
 		skipped := []output.Result{}
 
-		for _, warning := range result.Warnings {
+		for i := range result.Warnings {
```
lcarva, line 222: Why?

lcarva (approved):

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
zregvart: Yeah, the method doesn't mutate, it was used in the loop (see below)

internal/evaluator/conftest_evaluator.go
zregvart: Marked as a security issue, the `warning` variable is mutated for each iteration of the loop, it's address stays the same, so for each iteration a method receiving the address of such variable receives the same address. If that method was to keep the pointer and modify it, that would lead to data corruption. We don't keep the pointer in these methods so it is fairly safe, but it was still flagged by the linter.

Try [this](https://go.dev/play/p/d244Z8lyP-d) go playground code

lcarva: Oh! It's because we do this `addRuleMetadata(&warning, rules)`. That makes sense. Tricky tricky, but glad this got caught!

I suppose it doesn't affect us because `addRuleMetadata` only pull data from the warning result. We'd see a bug if we actually stored it in full.

PR #553, opened 2023-03-23
Use sigs.k8s.io/yaml instead of gh.com/ghodss/yaml

zregvart:
github.com/ghodss/yaml doesn't seem to be maintained, sigs.k8s.io/yaml is a permanent fork that is maintained.

Ref. https://issues.redhat.com/browse/HACBS-1936


lcarva (approved):
Yay!

PR #554, opened 2023-03-23
Introduce allow-list of secure network protocols

zregvart:
This makes sure that download doesn't proceed for insecure network protocols.

Ref. https://issues.redhat.com/browse/HACBS-1937

### DIFF
--- a/internal/downloader/downloader.go
+++ b/internal/downloader/downloader.go
@@ -21,9 +21,17 @@ package downloader
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/open-policy-agent/conftest/downloader"
 	log "github.com/sirupsen/logrus"
+
+	e "github.com/hacbs-contract/ec-cli/pkg/error"
+)
+
+var (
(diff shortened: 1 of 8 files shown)


zregvart:
Need to address the insecure git server used in acceptance tests

internal/downloader/downloader.go
```diff
@@ -21,9 +21,17 @@ package downloader
...
+	"strings"
 
 	"github.com/open-policy-agent/conftest/downloader"
 	log "github.com/sirupsen/logrus"
+
+	e "github.com/hacbs-contract/ec-cli/pkg/error"
+)
+
+var (
+	DL001 = e.NewError("DL001", "Attempting to download from unsecure source", e.ErrorExitStatus)
```
simonbaird, line 34: I was wondering about unsecure vs insecure. I think maybe insecure is correct, but there are people arguing either way in https://english.stackexchange.com/questions/19653/insecure-or-unsecure-when-dealing-with-security so :shrug: .

```suggestion
	DL001 = e.NewError("DL001", "Attempting to download from insecure source", e.ErrorExitStatus)
```

simonbaird (approved):
Great stuff.

PR #555, opened 2023-03-23
Use specificity score in result filtering

lcarva:
https://issues.redhat.com/browse/HACBS-1986

### DIFF
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -81,6 +81,22 @@ partial names, so you can't specify `"s*"` to match every package starting with
 You can also specify `"<packagename>.*"` and it works the same as just
 `"<packagename>"` to represent every rule in a package.
 
+A certain rule may match one or more items in the list of includes or excludes. In order
+to determine the precedence between these two lists, and, ultimately, whether a rule should
+be included, a specificity score is calculated for every match on each list. A score over the
+list of includes that is higher than the score over the list of excludes means the rule is
+included. Otherwise, the rule is excluded.
+
+For each name in each of the lists, the guidelines for computing the score are as such:
+
+. Add 1 if the name covers everything, i.e. "*"
+. Add 10 if the name specifies a package name, e.g. "package name", "package name.", "package name.*", or "package name.rule name"
+. Add 100 if a term is used, e.g. "*:term", "package name:term" or "package name.rule name:term"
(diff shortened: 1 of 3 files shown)


lcarva:
This approach was inspired by [CSS Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity). 

I was happily surprised at how well the algorithm behaves with the existing tests. No change in behavior other than fixing the issue reported in [HACBS-1986](https://issues.redhat.com/browse/HACBS-1986) which is encouraging.

I can see an argument that this approach makes it harder for a user to understand the filtering rules. However, it does feel natural to leverage specificity as a mechanism to determine precedence.

docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -81,6 +81,22 @@ partial names, so you can't specify `"s*"` to match every package starting with
...
+
+For each name in each of the lists, the guidelines for computing the score are as such:
+
+. Add 1 if the name covers everything, i.e. "*"
+. Add 10 if the name specifies a package name, e.g. "package name", "package name.", "package name.*", or "package name.rule name"
+. Add 100 if a term is used, e.g. "*:term", "package name:term" or "package name.rule name:term"
+. Add 100 if a rule is used, e.g. "package name.rule name", "package name.rule name:term"
+
+The score is cumulative. If a name is covered by multiple items in the guidelines, they
+are added together. For example, "package name.rule name:term" scores at 210.
```
zregvart, line 98: Not sure about repeating "package name.rule name:term" twice. For me it is easier if I see a concrete examples.

```suggestion
. Add 10 if the name specifies a package name, e.g. "release.test", "release.test.", "release.test.*", or "release.test.test_result_failures"
. Add 100 if a term is used, e.g. "*:term", "release.test:clamav-scan" or "release.test.test_result_failures:clamav-scan"
. Add 100 if a rule is used, e.g. "release.test.test_result_failures"

The score is cumulative. If a name is covered by multiple items in the guidelines, they
are added together. For example, "release.test.test_result_failures:clamav-scan" scores at 210.
```

zregvart (approved):
Neat

simonbaird:
Does this change mean the existing examples now do work as described?

lcarva:
> Does this change mean the existing examples now do work as described?
> 
> Edit: I think it does from your comment above.

Yup!

PR #568, opened 2023-03-29
Use a single Kind cluster in acceptance tests

zregvart:
Makes for much faster acceptance tests. At some point we _might_ need distinct clusters but I don't see the need right now.


joejstuart:
Awesome!

joejstuart (approved):

lcarva (approved):

PR #572, opened 2023-03-29
Display HACBS_TEST_OUTPUT in logs

mbestavros:
Implementing the result of discussion in Slack.

Ref:

- https://issues.redhat.com/browse/HACBS-2032
- https://issues.redhat.com/browse/STONE-574


tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -135,3 +135,19 @@ spec:
           # value expected by Tekton Pipelines. NOTE: If params.SSL_CERT_DIR is empty, the value
           # will contain a trailing ":" - this is ok.
           value: "/tekton-custom-certs:/etc/ssl/certs:/etc/pki/tls/certs:/system/etc/security/cacerts:$(params.SSL_CERT_DIR)"
+    - name: display
```
lcarva, line 138: How about `summary`?

This is somewhat user visible so worth thinking of a good descriptive name.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -135,3 +135,19 @@ spec:
...
+      image: quay.io/hacbs-contract/ec-cli:snapshot
+      command: [jq]
```
lcarva, line 140: We may have to install this in the ec-cli image?

lcarva (commented):
Seems like the right track

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
zregvart: [Yeah](https://github.com/enterprise-contract/ec-cli/actions/runs/4558491080/jobs/8041441182?pr=572#step:5:552):
```
Error executing command: exec: "jq": executable file not found in $PATH
```

mbestavros: I don't know if the top-level Dockerfile is the right place to install `jq`, but that's where I put it. I can move if needed.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
mbestavros: Done!

zregvart (approved):

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -135,3 +135,19 @@ spec:
...
+      args:
+        - "."
+        - "$(results.HACBS_TEST_OUTPUT.path)"
+    - name: assert
+      image: quay.io/hacbs-contract/ec-cli:snapshot
+      command: [jq]
+      args:
+        - "--argjson"
```
zregvart, line 148: Do we need this?
```suggestion
	- "--exit-status"
        - "--argjson"
```

zregvart (changes requested):
Seems that the `assert` step outputs [`false`](https://github.com/enterprise-contract/ec-cli/actions/runs/4568914276/jobs/8064551533?pr=572#step:5:567), which is odd given that we can see [`success: true`](https://github.com/enterprise-contract/ec-cli/actions/runs/4568914276/jobs/8064551533?pr=572#step:5:546) in the YAML output.
I think we need to rely on exit code from `jq`, and to do that we need to add `--exit-status` parameter to `jq`.

zregvart:
Hint, you can run `make scenario_Golden_container_image` to run just that single acceptance test that failed

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
zregvart: Nevermind I see `-e` below

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -135,3 +135,19 @@ spec:
...
+        - "strict"
+        - "$(params.STRICT)"
+        - "-e"
+        - ".result == "SUCCESS" or ($strict | not)"
```
zregvart, line 152: I think we need this to make the YAML valid:

```suggestion
        - >
          ".result == "SUCCESS" or ($strict | not)"
```

lcarva: alternatively these also work: 

`- '.result == "SUCCESS" or ($strict | not)'`
`- .result == "SUCCESS" or ($strict | not)`

pick your poison!

lcarva (approved):

PR #574, opened 2023-03-30
Rework JSON diff logic in acceptance tests

zregvart:
We had logic in place to check if the expected JSON in acceptance tests matched the output JSON by checking if any of the detected changes matched via regular expressions.
This was used to not fail the acceptance tests for JSON outputs that had few bits that either changed with time (e.g. time/data/timestamps), or had a random element to them (e.g. digests of random images). The output in case of differences still presented all changes, even those that were matched via regular expressions.
This change filters out the differences that have been matched via regular expressions, so that the output contains only the relevant difference, i.e. the cause of the failure.
Due to several issues with the existing logic the output was matched loosely, and for those the acceptance tests were corrected.


acceptance/image/image.go
```diff
@@ -437,12 +446,25 @@ func storeAttestationSignatures(ctx context.Context, rawAttestation []byte) (con
 	if err := json.Unmarshal(rawAttestation, &attestationPayload); err != nil {
 		return nil, err
 	}
-	signaturesJson, err := json.Marshal(attestationPayload.Signatures)
+
+	signatures := make([]entitySignature, 0, len(attestationPayload.Signatures))
+	for _, signature := range attestationPayload.Signatures {
+		signatures = append(signatures, entitySignature{
```
lcarva, line 452: Is this basically stripping off some fields? It would be good to document it.

lcarva (approved):

PR #576, opened 2023-03-30
Add support for keyless workflow

lcarva:
https://issues.redhat.com/browse/HACBS-2045

This is still a work in progress, but wanted to share the progress so far.

If you want to try it out:

```
EC_EXPERIMENTAL=1 go run . validate image \
  --policy "" \
  --image quay.io/lucarval/festoji:latest \
  --output yaml \
  --certificate-identity-regexp '.*' \
  --certificate-oidc-issuer 'https://token.actions.githubusercontent.com' \
  --rekor-url 'https://rekor.sigstore.dev'
```

I'm using an empty policy because I don't really care about the contents of the SLSA Provenance attestation. But it should be possible to create some rego that does evaluate the attestation for the image in the example above (which is built on GitHub!).

~~We need to figure out how to properly support the `--certificate-oidc-issuer ` and `--certificate-identity` flags. I *think* we want support for multiple identities, but just one OIDC issuer. Not sure. Also, the image I picked to test this on is signed by a certain `--certificate-identity`, but its attestation uses a different `--certificate-identity`. The code currently works because it accepts a sig or att from either certificate identity. This is probably not the best solution. Maybe we want to be explicit and say something like the image signature must come from this particular certificate identity, and the slsa provenance from this other one, and the sbom from yet another one. That might give us the flexibility we need.~~
UPDATE: After some consideration, let's instead allow a Regular Expression for those value, like cosign does. Then, we can pipe this data to rego which gives us a lot of flexibility in expressing what is expected for those values. Also updated the example command above.

We need to figure out how to properly handle a custom rekor and fulcio instances. But maybe it can be postponed for now.

When keyless is used, there's no public key to display in the report. That's expected. We should add something to the report though. Maybe the identities used? We may get this for free if we move the identities to ECP (which I think we should!).

Anyways, looking for any feedback here. I'll continue to hack away and may push changes. I'll do so as new commits to make it easier to review.

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -26,6 +26,7 @@ import (
 	hd "github.com/MakeNowJust/heredoc"
 	"github.com/hashicorp/go-multierror"
 	app "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	"github.com/sigstore/cosign/v2/pkg/cosign"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 
@@ -40,20 +41,24 @@ type imageValidationFunc func(context.Context, string, policy.Policy, bool) (*ou
 
 func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 	var data = struct {
-		effectiveTime       string
-		filePath            string
-		imageRef            string
-		info                bool
-		input               string
-		output              []string
-		outputFile          string
-		policy              policy.Policy
-		policyConfiguration string
-		publicKey           string
-		rekorURL            string
-		snapshot            string
-		spec                *app.SnapshotSpec
-		strict              bool
+		certificateIdentity         string
+		certificateIdentityRegExp   string
+		certificateOIDCIssuer       string
+		certificateOIDCIssuerRegExp string
+		effectiveTime               string
+		filePath                    string
+		imageRef                    string
+		info                        bool
+		input                       string
+		output                      []string
+		outputFile                  string
+		policy                      policy.Policy
+		policyConfiguration         string
+		publicKey                   string
+		rekorURL                    string
+		snapshot                    string
+		spec                        *app.SnapshotSpec
+		strict                      bool
 	}{
 
 		policyConfiguration: "enterprise-contract-service/default",
@@ -127,6 +132,22 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			Write output in YAML format to stdout and in HACBS format to a file
 			  ec validate image --image registry/name:tag --output yaml --output hacbs=<path>
+
+			Validate a single image with keyless workflow. This is an experimental feature
+			that requires setting the EC_EXPERIMENTAL environment variable to "1".
+
+			  EC_EXPERIMENTAL="1" ec validate image --image registry/name:tag --policy my-policy \
+			    --certificate-identity 'https://github.com/user/repo/.github/workflows/push.yaml@refs/heads/main' \
+			    --certificate-oidc-issuer 'https://token.actions.githubusercontent.com' \
+			    --rekor-url 'https://rekor.sigstore.dev'
+
+			Use a regular expression to match certificate attributes. This is an experimental
+			feature that requires setting the EC_EXPERIMENTAL environment variable to "1".
+
+			  EC_EXPERIMENTAL="1" ec validate image --image registry/name:tag --policy my-policy \
+			    --certificate-identity-regexp '^https://github\.com' \
+			    --certificate-oidc-issuer-regexp 'githubusercontent' \
+			    --rekor-url 'https://rekor.sigstore.dev'
 		`),
 
 		PreRunE: func(cmd *cobra.Command, args []string) (allErrors error) {
@@ -140,6 +161,13 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 				allErrors = multierror.Append(allErrors, err)
 			} else {
 				data.spec = s
+			}
+
+			identity := cosign.Identity{
+				Issuer:        data.certificateOIDCIssuer,
+				IssuerRegExp:  data.certificateOIDCIssuerRegExp,
+				Subject:       data.certificateIdentity,
+				SubjectRegExp: data.certificateIdentityRegExp,
 			}
 
 			// Check if policyConfiguration is a file path, if so, we read it into the var data.policyConfiguration
@@ -161,7 +189,8 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			}
 
 			if p, err := policy.NewPolicy(
-				cmd.Context(), data.policyConfiguration, data.rekorURL, data.publicKey, data.effectiveTime,
+				cmd.Context(), data.policyConfiguration, data.rekorURL, data.publicKey,
+				data.effectiveTime, identity,
 			); err != nil {
 				allErrors = multierror.Append(allErrors, err)
 			} else {
@@ -263,6 +292,18 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 	cmd.Flags().StringVarP(&data.rekorURL, "rekor-url", "r", data.rekorURL,
 		"Rekor URL. Overrides rekorURL from EnterpriseContractPolicy")
+
+	cmd.Flags().StringVar(&data.certificateIdentity, "certificate-identity", data.certificateIdentity,
+		"EXPERIMENTAL. URL of the certificate identity for keyless verification")
+
+	cmd.Flags().StringVar(&data.certificateIdentityRegExp, "certificate-identity-regexp", data.certificateIdentityRegExp,
+		"EXPERIMENTAL. Regular expression for the URL of the certificate identity for keyless verification")
+
+	cmd.Flags().StringVar(&data.certificateOIDCIssuer, "certificate-oidc-issuer", data.certificateOIDCIssuer,
+		"EXPERIMENTAL. URL of the certificate OIDC issuer for keyless verification")
+
+	cmd.Flags().StringVar(&data.certificateOIDCIssuerRegExp, "certificate-oidc-issuer-regexp", data.certificateOIDCIssuerRegExp,
+		"EXPERIMENTAL. Regular expresssion for the URL of the certificate OIDC issuer for keyless verification")
 
 	cmd.Flags().StringVarP(&data.filePath, "file-path", "f", data.filePath,
 		"path to ApplicationSnapshot Spec JSON file")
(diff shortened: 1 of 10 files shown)


mbestavros:
Very cool! Excited to see where this goes :eyes:

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -169,6 +169,9 @@ func (a *ApplicationSnapshotImage) ValidateImageSignature(ctx context.Context) e
 
 // ValidateAttestationSignature executes the cosign.VerifyImageAttestations method
 func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Context) error {
+	// TODO: This is a hack of course. We need maybe not set ClaimVerifier, and make a copy
+	// of checkOpts so we can set the correct one for signature vs attestation.
+	a.checkOpts.ClaimVerifier = cosign.IntotoSubjectClaimVerifier
```
zregvart, line 174: Derived from the policy? So that the CheckOpts created in `policy.CheckOpts()` is already preconfigured?

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -180,7 +183,9 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
 	for _, att := range layers {
 		sp, err := attestation.SLSAProvenanceFromLayer(att)
 		if err != nil {
-			return err
+			// TODO: This is probably the right call.
+			log.Debugf("Ignoring non SLSA Provenance attestation: %s", err)
+			continue
```
zregvart, line 188: Yeah, policy will catch if it's missing...

internal/policy/policy.go
```diff
@@ -278,49 +288,57 @@ func parseEffectiveTime(choosenTime string) (*time.Time, error) {
...
-					logId: {
-						PubKey: rekorPublicKey,
-						Status: tuf.Active,
-					},
-				},
-			}
+		if opts.RekorPubKeys, err = cosign.GetRekorPubs(ctx); err != nil {
+			return nil, err
+		}
+		log.Debug("Retrieved Rekor public keys")
```
zregvart, line 308: We could pull this change as a separate PR and have it merged in main... Does it get the key from TUF or from Rekor?

internal/policy/policy.go
```diff
@@ -278,49 +288,57 @@ func parseEffectiveTime(choosenTime string) (*time.Time, error) {
...
+		for _, identity := range p.certificateIdentity {
+			identities = append(identities, cosign.Identity{
+				Issuer:  p.certificateOIDCIssuer,
+				Subject: identity,
+			})
+		}
+		opts.Identities = identities
+
+		// Get some Fulcio stuff
+		if opts.RootCerts, err = fulcio.GetRoots(); err != nil {
```
zregvart, line 328: This is from TUF?

internal/policy/policy.go
```diff
@@ -278,49 +288,57 @@ func parseEffectiveTime(choosenTime string) (*time.Time, error) {
...
+			return nil, err
 		}
+		if opts.IntermediateCerts, err = fulcio.GetIntermediates(); err != nil {
+			return nil, err
+		}
+
+		// Now get something from the ctlog
+		// TODO: what is the ctlog?!
```
zregvart, line 336: [Certificate Transparency](https://en.wikipedia.org/wiki/Certificate_Transparency) log I guess

zregvart (commented):
Neat, not a lot of changes after all.

On the verification side, do we care about identities or do we care that those identities were issued a certificate from a known CA?

internal/policy/policy.go
lcarva: Yes. It seems to be the case.

internal/policy/policy.go
lcarva: +1, and yes it does get it from TUF.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
lcarva: Yeah. Updated with the copy approach described in the comment.

lcarva:
Updated the description after our conversation about handling identities.

cmd/validate/image_test.go
```diff
@@ -271,6 +272,44 @@ func Test_ValidateImageCommand(t *testing.T) {
 	  }`, mockPublicKey, mockPublicKey), out.String())
 }
 
+func Test_ValidateImageCommandKeyless(t *testing.T) {
```
lcarva, line 275: This test is timing out because the call to `NewPolicy` is triggering resolving certs and keys.

zregvart:
> When keyless is used, there's no public key to display in the report. That's expected. We should add something to the report though. Maybe the identities used? We may get this for free if we move the identities to ECP (which I think we should!).

For sure there are keys involved that we can display, but they might not be as important as displaying identity (certificate) information. The non-keyless signatures can also utilize certificates for which we can display some information, e.g: subject, subjectAlternativeName extension, issuer and serial.

lcarva:
This is ready for review. I'd like to have it merged soon so it's easier to collaborate.

internal/policy/policy.go
```diff
@@ -344,3 +386,23 @@ func signatureVerifier(ctx context.Context, p *policy) (sigstoreSig.Verifier, er
...
+		errs = multierror.Append(errs, errors.New(
+			"certificate identity must be provided for keyless workflow"))
+	}
+
+	return errs
+}
+
+func experimental() bool {
+	return os.Getenv("EC_EXPERIMENTAL") == "1"
+}
```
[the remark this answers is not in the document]
simonbaird: (Others might suggest a fancier feature flag mechanism, but I think that would be overkill.)

simonbaird (approved):

cmd/validate/image.go
```diff
@@ -127,6 +132,14 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			Write output in YAML format to stdout and in HACBS format to a file
 			  ec validate image --image registry/name:tag --output yaml --output hacbs=<path>
+
+			Validate a single image with keyless workflow. This is an experimental feature
+			that requires setting the EC_EXPERIMENTAL environment variable to "1".
+
+			  EC_EXPERIMENTAL="1" ec validate image --image registry/name:tag --policy my-policy \
+			  --certificate-identity-regexp '.*' \
+			  --certificate-oidc-issuer-regexp '.*' \
```
simonbaird, line 141: Suggestion: One more example that doesn't use `'.*'`, just so reader can see how a tighter regex might look. E.g. should there be a '^' at the start or is it implicit?

simonbaird (commented):
It feels nice and clean. Good stuff!

cmd/validate/image.go
```diff
@@ -40,20 +41,24 @@ type imageValidationFunc func(context.Context, string, policy.Policy, bool) (*ou
...
-		policyConfiguration string
-		publicKey           string
-		rekorURL            string
-		snapshot            string
-		spec                *app.SnapshotSpec
-		strict              bool
+		certificateIdentity         string
+		certificateIdentityRegExp   string
+		certificateOIDCIssuer       string
+		certificateOIDCIssuerRegExp string
```
simonbaird, line 47: Could be "certFoo" instead of "certificateFoo" here too perhaps. (Ps, I have nothing against the longer names if you like them, just something to consider.)

zregvart, line 47: Should we consider providing this (also?) as inline JSON, worse for UX but better for flexibility if `cosign.Identity` changes?

cmd/validate/image_test.go
```diff
@@ -759,3 +802,57 @@ func Test_WarningOutput(t *testing.T) {
...
 	  }`, mockPublicKey, mockPublicKey), out.String())
 }
+
+func setupFulcioRoots(t *testing.T) {
+	// Do not use afero.NewMemMapFs() here because the file is read by cosign
+	// which does not understand the filesystem-from-context pattern
+	f, err := os.Create(path.Join(t.TempDir(), "fulcio.pem"))
+	assert.NoError(t, err)
+	defer f.Close()
+	_, err = f.Write([]byte(hd.Doc(`
```
zregvart, line 812: Could use [embed](https://pkg.go.dev/embed) here, then these files could be descriptive by file name

internal/policy/policy.go
```diff
@@ -66,11 +69,18 @@ type policy struct {
 	choosenTime     string
 	effectiveTime   *time.Time
 	attestationTime *time.Time
+	// TODO: Move these to ecc.EnterpriseContractPolicySpec
+	identity cosign.Identity
```
zregvart, line 73: We should have a story for this

internal/policy/policy_test.go
```diff
@@ -396,10 +594,64 @@ func toYAML(policy *ecc.EnterpriseContractPolicySpec) string {
...
 }
+
+func setupTestFulcioRoots(t *testing.T) {
+	// Do not use afero.NewMemMapFs() here because the file is read by cosign
+	// which does not understand the filesystem-from-context pattern
+	f, err := os.Create(path.Join(t.TempDir(), "fulcio.pem"))
+	assert.NoError(t, err)
+	defer f.Close()
+	_, err = f.Write([]byte(hd.Doc(`
+		-----BEGIN CERTIFICATE-----
```
zregvart, line 612: Could we reuse the ones from above so we don't need to update twice?

zregvart (approved):
Neat, we should file more stories for docs and acceptance tests

lcarva:
> we should file more stories for docs and acceptance tests

We have whole epic: https://issues.redhat.com/browse/HACBS-2069
Let me know if any stories are missing there!

cmd/validate/image.go
```diff
@@ -264,6 +285,18 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 	cmd.Flags().StringVarP(&data.rekorURL, "rekor-url", "r", data.rekorURL,
 		"Rekor URL. Overrides rekorURL from EnterpriseContractPolicy")
 
+	cmd.Flags().StringVar(&data.certificateIdentity, "certificate-identity", data.certificateIdentity,
```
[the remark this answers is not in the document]
lcarva: Cosign uses the long form. I think I'd prefer to match that to give users some sense of familiarity.

internal/policy/policy.go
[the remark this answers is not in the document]
lcarva: It makes the code using this function a bit easier to read. I like it.

cmd/validate/image.go
lcarva: Let's keep the long form for now. These are internal attributes that we could change later on without breaking anything.

lcarva: https://issues.redhat.com/browse/HACBS-2073 should take care of that.

PR #577, opened 2023-03-30
UPDATE: show value of ec version --short in output

robnester-rh:
This commit addresses HACBS-2018 and adds the same output as executing `ec version --short` in the report output for `ec validate image` and `ec validate definition`.


zregvart (approved):
Sweet

PR #579, opened 2023-03-31
Don't hang in acceptance tests when Kind won't ...

zregvart:
...start

We were counting a cluster consumer(s) even though Kind cluster could not be started. Similarly we should not try to stop a non-running cluster. Both could lead to deadlocks or errors.


robnester-rh (approved):
lgtm

PR #580, opened 2023-03-31
Verbose Kind logs in acceptance tests

zregvart:
Adds (very) verbose Kind logs in acceptance tests. These will show up only on failures, so they will not pollute the successful execution.

### DIFF
--- a/acceptance/log/log.go
+++ b/acceptance/log/log.go
@@ -21,13 +21,23 @@ import (
 	"context"
 	"testing"
 
+	"sigs.k8s.io/kind/pkg/log"
+
 	"github.com/enterprise-contract/ec-cli/acceptance/testenv"
 )
 
 type Logger interface {
-	Log(args ...interface{})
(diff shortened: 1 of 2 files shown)


acceptance/log/log.go
```diff
@@ -21,34 +21,76 @@ import (
...
-func (l logger) Printf(format string, args ...interface{}) {
+func (l logger) Printf(format string, args ...any) {
 	l.Logf(format, args...)
 }
 
+func (l logger) Warn(message string) {
+	l.Logf("[WARN ] %s", message)
+}
+
+func (l logger) Warnf(format string, args ...interface{}) {
```
robnester-rh, line 66: Is there a reason to use `interface{}` here instead of being consistent with the above code and using `any`? I understand that `any` is an alias for `interface{}` so I'm just wondering if I'm missing some subtle reason?

zregvart: Copy pasta

PR #582, opened 2023-03-31
Use cosign's claim verifiers

lcarva:
Cosign has built-in support to ensure the fetched signature and attestations match the image being verified. This commit uses that feature for both signatures and attestations.

This is what the removed code intended to do for attestations. However, it did nothing for signatures :O


internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -163,13 +163,19 @@ func (a *ApplicationSnapshotImage) SetImageURL(url string) error {
 
 // ValidateImageSignature executes the cosign.VerifyImageSignature method on the ApplicationSnapshotImage image ref.
 func (a *ApplicationSnapshotImage) ValidateImageSignature(ctx context.Context) error {
-	_, _, err := NewClient(ctx).VerifyImageSignatures(ctx, a.reference, &a.checkOpts)
+	// Set the expected ClaimVerifier on a shallow *copy* of CheckOpts
+	opts := a.checkOpts
```
zregvart, line 167: Do we need to do this? If so perhaps the comment could elaborate as to why

lcarva: I think we do because otherwise, this method is causing a non-obvious side effect. I'll comment that.

lcarva:
> I guess we trust cosign to do this correctly, should we have a test that verifies that?

Yes. Filed https://issues.redhat.com/browse/HACBS-2068 for that.

mbestavros (approved):
Hooray for upstreaming!

PR #586, opened 2023-04-03
ADD ability to pass policy file via --policy flag

robnester-rh:
This commit addresses HACBS-1975 and allows the user to specify a YAML or JSON file containing policy configuration to `ec` in lieu of inline policy configuration YAML or JSON text.

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -20,11 +20,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	hd "github.com/MakeNowJust/heredoc"
 	"github.com/hashicorp/go-multierror"
 	app "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 
 	"github.com/enterprise-contract/ec-cli/internal/applicationsnapshot"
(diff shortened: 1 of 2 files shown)


cmd/validate/image.go
```diff
@@ -140,6 +142,23 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 				data.spec = s
 			}
 
+			// Check if policyConfiguration is a file path, if so, we read it into the var data.policyConfiguration
+			if strings.Contains(data.policyConfiguration, ".yaml") || strings.Contains(data.policyConfiguration, ".yml") || strings.Contains(data.policyConfiguration, ".json") {
```
zregvart, line 146: Perhaps `strings.HasSuffix` instead of `strings.Contains`?

cmd/validate/image.go
```diff
@@ -140,6 +142,23 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
...
+				policyBytes, err := os.ReadFile(data.policyConfiguration)
```
zregvart, line 147: Go through afero so we can write a test for this?

zregvart (commented):
Tests?

cmd/validate/image.go
robnester-rh: ack

robnester-rh:
> Tests?

I'll add them, standby.

zregvart (approved):

PR #587, opened 2023-04-03
Task should not fail on warnings in strict mode

simonbaird:
The new jq test from #572 fails the task when running with the strict param set, unless the result is "SUCCESS", which means that a result of "WARNING" is now failing unexpectedly. Noticed by @shebert.

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -150,5 +150,5 @@ spec:
         - "$(params.STRICT)"
         - "-e"
         - >
-          .result == "SUCCESS" or ($strict | not)
+          .result == "SUCCESS" or .result == "WARNING" or ($strict | not)
         - "$(results.HACBS_TEST_OUTPUT.path)"


simonbaird:
I think "FAILURE" is the right string to check for because of https://github.com/enterprise-contract/ec-cli/blob/main/internal/applicationsnapshot/report.go#L231 .

simonbaird:
To test the jq command I'm doing things like this:

```
echo '{"result": "SKIPPED"}' |  jq --argjson strict true -e '.result != "FAILURE" or ($strict | not)'; echo $?

echo '{"result": "WARNING"}' |  jq --argjson strict true -e '.result != "FAILURE" or ($strict | not)'; echo $?

echo '{"result": "FAILURE"}' |  jq --argjson strict true -e '.result != "FAILURE" or ($strict | not)'; echo $?

```

Afaict it's working as expected.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -150,5 +150,5 @@ spec:
         - "$(params.STRICT)"
         - "-e"
         - >
-          .result == "SUCCESS" or ($strict | not)
+          .result != "FAILURE" or ($strict | not)
```
lcarva, line 153: It's also possible to have an `ERROR` state, which we never set but mentioning it here just in case.

simonbaird: We could check for all the known passing statuses:
```suggestion
          .result == "SUCCESS" or or .result == "WARNING" or .result == "SKIPPED" or ($strict | not)
```

simonbaird:
There was some discussion on slack about the behavior of `ec validate image --strict` and what we expected for result "SKIPPED". Conclusion was to merge this asap and consider those questions in later PRs.

PR #589, opened 2023-04-03
Isolate report output

lcarva:
This ensures the big report from the verify-ec Task is not accidentally disrupted by unrelated stderr text, such as reported in https://issues.redhat.com/browse/HACBS-2064

It also allows the task to produce runtime logs that are helpful for debugging purposes.


lcarva:
There's an impact to the UI in the sense that they must look at a different step name. But we can rename the steps to make that not the case if that's a concern.

mbestavros (approved):
Good change! LGTM.

zregvart (approved):

PR #590, opened 2023-04-04
Don't show effective_on in output if expired

mbestavros:
Quick patch for https://issues.redhat.com/browse/HACBS-2085.

The second requirement in https://issues.redhat.com/browse/HACBS-1985 wasn't met by https://github.com/enterprise-contract/ec-cli/pull/521.

### DIFF
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -65,6 +65,7 @@ type testRunner interface {
 
 const (
 	effectiveOnFormat   = "2006-01-02T15:04:05Z"
+	effectiveOnTimeout  = -90 * 24 * time.Hour // keep effective_on metadata up to 90 days
 	metadataCode        = "code"
 	metadataCollections = "collections"
 	metadataDescription = "description"
@@ -390,6 +391,18 @@ func addMetadataToResults(r *output.Result, rule rule.Info) {
 
 	if rule.Title != "" {
 		r.Metadata[metadataTitle] = rule.Title
+	}
+	if rule.EffectiveOn != "" {
+		effectiveOnTime, err := time.Parse(effectiveOnFormat, rule.EffectiveOn)
+		if err == nil {
+			// If the rule has been effective for a long time, we'll consider
+			// the effective_on date not relevant and not bother including it
+			if effectiveOnTime.After(time.Now().Add(effectiveOnTimeout)) {
+				r.Metadata[metadataEffectiveOn] = rule.EffectiveOn
+			}
+		} else {
+			log.Warnf("Invalid %q value %q", metadataEffectiveOn, rule.EffectiveOn)
+		}
 	}
 	if rule.Description != "" {
 		r.Metadata[metadataDescription] = rule.Description
--- a/internal/evaluator/conftest_evaluator_test.go
+++ b/internal/evaluator/conftest_evaluator_test.go
@@ -21,6 +21,7 @@ package evaluator
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/MakeNowJust/heredoc"
 	ecc "github.com/enterprise-contract/enterprise-contract-controller/api/v1alpha1"
@@ -1182,6 +1183,8 @@ func TestCollectAnnotationData(t *testing.T) {
 }
 
 func TestRuleMetadata(t *testing.T) {
+	effectiveOnTest := time.Now().Format(effectiveOnFormat)
+
 	rules := policyRules{
 		"warning1": rule.Info{
 			Title: "Warning1",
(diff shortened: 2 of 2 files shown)


internal/evaluator/conftest_evaluator.go
```diff
@@ -71,6 +71,7 @@ const (
 	metadataEffectiveOn = "effective_on"
 	metadataTerm        = "term"
 	metadataTitle       = "title"
+	EffectiveOnStale    = 90
```
simonbaird, line 74: I think it should be lowercase unless you want it visible in other modules.
```suggestion
	effectiveOnStale    = 90
```

mbestavros: Good point, fixed.

simonbaird: Maybe a comment:

```suggestion
	effectiveOnStale    = 90 # Days
```

Or make it more clear like this:

```suggestion
	effectiveOnStaleDays  = 90
```

simonbaird:
I think some test coverage would be a good idea.

internal/evaluator/conftest_evaluator.go
mbestavros: I like the second option better, but it does push the spacing for the entire `const` a little bit. Still okay?

simonbaird:
I mightbe wrong, but I think this only works for the successes. Does it also remove stale effective_on values from the warnings and violations?

simonbaird (changes requested):

internal/evaluator/conftest_evaluator.go
```diff
@@ -64,13 +64,14 @@ type testRunner interface {
...
-	metadataTerm        = "term"
-	metadataTitle       = "title"
+	effectiveOnFormat    = "2006-01-02T15:04:05Z"
+	metadataCode         = "code"
+	metadataCollections  = "collections"
+	metadataDescription  = "description"
+	metadataEffectiveOn  = "effective_on"
+	metadataTerm         = "term"
+	metadataTitle        = "title"
+	effectiveOnStaleDays = 90
```
zregvart, line 74: Why not
```suggestion
	effectiveOnTimeout = -90 * 24 * time.Hour // keep effective_on metadata up to 90 days
```

internal/evaluator/conftest_evaluator.go
lcarva: > I like the second option better, but it does push the spacing for the entire `const` a little bit. Still okay?

That shouldn't be a concern.

mbestavros:
For those following, I've added tests for the new functionality in `conftest_evaluator_test.go`.

internal/evaluator/conftest_evaluator_test.go
```diff
@@ -1235,6 +1248,45 @@ func TestRuleMetadata(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "drop stale effectiveOn",
+			result: output.Result{
+				Metadata: map[string]any{
+					"code":        "warning2",
+					"collections": []any{"A"},
```
simonbaird, line 1256: Maybe this to be consistent with line 1264..?

```suggestion
					"collections": []string{"A"},
```

internal/evaluator/conftest_evaluator_test.go
```diff
@@ -1235,6 +1248,45 @@ func TestRuleMetadata(t *testing.T) {
...
+			want: output.Result{
+				Metadata: map[string]any{
+					"code":         "warning3",
+					"collections":  []string{"A"},
+					"description":  "Warning 3 description",
+					"title":        "Warning3",
+					"effective_on": effectiveOnTest,
+				},
+			},
+		},
```
simonbaird, line 1289: Looks good.

internal/evaluator/conftest_evaluator_test.go
```diff
@@ -1235,6 +1248,45 @@ func TestRuleMetadata(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "drop stale effectiveOn",
+			result: output.Result{
+				Metadata: map[string]any{
+					"code":        "warning2",
+					"collections": []string{"A"},
```
simonbaird, line 1256: You're too fast! :D After reading it some more I wasn't sure if the `[]any{"A"}` was more realistic compare the real output.

simonbaird: Since we have `"collections": []any{"A"}` on the existing test, I think it's probably better.

Sorry for the bad advice.

mbestavros: _ninja editor_ :wink: 

I can change it back, if that's better!

simonbaird: Yeah, let's be consistent with the other tests.

internal/evaluator/conftest_evaluator_test.go
simonbaird: For the record I deleted a comment about using `[]string{"A"},` instead of `[]any{"A"},` after looking a bit more carefully.

internal/evaluator/conftest_evaluator_test.go
mbestavros: done!

simonbaird (approved):
Lgtm. I'd suggest it's ready to merge once the CI goes green.

PR #605, opened 2023-04-07
Isolate policies from accessing external resources

lcarva:
https://issues.redhat.com/browse/HACBS-1938

### DIFF
--- a/acceptance/cli/cli.go
+++ b/acceptance/cli/cli.go
@@ -420,6 +420,36 @@ func theStandardOutputShouldContain(ctx context.Context, expected *godog.DocStri
 	return fmt.Errorf("expected and actual output differ:\n%s", b.String())
 }
 
+// theStandardErrorShouldContain looks at the standard error (stderr) of the last invoked ec
+// command and compares the expected error with the resulted error.
+func theStandardErrorShouldContain(ctx context.Context, expected *godog.DocString) error {
+	status, err := ecStatusFrom(ctx)
+	if err != nil {
+		return err
+	}
+
+	if expected == nil {
+		return errors.New("must provide expected error")
+	}
+
+	expectedStdErr := os.Expand(expected.Content, func(key string) string {
+		return status.vars[key]
+	})
+
+	stderr := status.stderr.String()
+
+	// shortcut, if the output is exactly as expected
+	if stderr == expectedStdErr {
+		return nil
+	}
+
+	if matched, err := regexp.MatchString(expectedStdErr, stderr); matched && err == nil {
+		return nil
+	}
+
+	return fmt.Errorf("expected error:\n%s\nnot found in standard error:\n%s", expected, stderr)
+}
+
 func filterMatchedByRegexp(obj any, diff gojsondiff.Diff) diffy {
 	deltas := diff.Deltas()
 	if v, ok := obj.(map[string]any); ok {
@@ -605,6 +635,7 @@ func AddStepsTo(sc *godog.ScenarioContext) {
 	sc.Step(`^ec command is run with "(.+)"$`, ecCommandIsRunWith)
 	sc.Step(`^the exit status should be (\d+)$`, theExitStatusIs)
 	sc.Step(`^the standard output should contain$`, theStandardOutputShouldContain)
+	sc.Step(`^the standard error should contain$`, theStandardErrorShouldContain)
 	sc.After(func(ctx context.Context, sc *godog.Scenario, err error) (context.Context, error) {
 		logExecution(ctx)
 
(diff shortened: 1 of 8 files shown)


features/validate_image.feature
```diff
@@ -952,3 +952,35 @@ Feature: evaluate enterprise contract
...
+          "policy": [
+            "git::https://${GITHOST}/git/happy-day-policy.git"
+          ]
+        }
+      ]
+    }
+    """
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/ec-happy-day --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --rekor-url ${REKOR} --strict"
+    Then the exit status should be 1
+    Then the standard error should contain
```
joejstuart, line 980: Would it make sense to add another line defining what the standard output will look like?

acceptance/cli/cli.go
```diff
@@ -420,6 +420,36 @@ func theStandardOutputShouldContain(ctx context.Context, expected *godog.DocStri
...
+		return nil
+	}
+
+	if matched, err := regexp.MatchString(expectedStdErr, stderr); matched && err == nil {
+		return nil
+	}
+
+	return fmt.Errorf("expected error:\n%s\nnot found in standard error:\n%s", expected, stderr)
+}
+
```
simonbaird, line 452: Awesome, I've been looking for this one for a while.

features/validate_image.feature
lcarva: It doesn't hurt! Added.

internal/evaluator/conftest_evaluator.go
```diff
@@ -622,3 +657,46 @@ func ExtractStringFromMetadata(result output.Result, key string) string {
...
 	return ""
 }
+
+func withCapabilities(ctx context.Context, capabilities string) context.Context {
+	return context.WithValue(ctx, capabilitiesKey, capabilities)
+}
+
+// capabilities returns a JSON serialized OPA Capability meant to isolate rego
+// policies from accessing external information, such as hosts or environment
+// variables. If the context already contains the capability, then that is
```
simonbaird, line 667: I was going to suggest some commentary to indicate "we do this for security reasons", but it's probably clear enough now.

simonbaird (approved):

internal/evaluator/conftest_evaluator.go
```diff
@@ -622,3 +657,46 @@ func ExtractStringFromMetadata(result output.Result, key string) string {
...
+	// drops the "allow_net" attribute if it's an empty list. So when it's loaded
+	// by OPA, it's seen as a nil value. As a workaround, we add an empty string
+	// to the list which shouldn't match any host but preserves the list after the
+	// JSON dance.
+	capabilities.AllowNet = []string{""}
+
+	builtins := make([]*ast.Builtin, 0, len(capabilities.Builtins))
+	for _, b := range capabilities.Builtins {
+		switch b.Name {
+		case "opa.runtime":
```
joejstuart, line 686: Do we want to log skipped capabilities? I'm just thinking if a user sees that the functions are undefined in stderr, maybe having it logged that it's not allowed may help.

internal/evaluator/conftest_evaluator.go
```diff
@@ -622,3 +657,46 @@ func ExtractStringFromMetadata(result output.Result, key string) string {
...
+
+func withCapabilities(ctx context.Context, capabilities string) context.Context {
+	return context.WithValue(ctx, capabilitiesKey, capabilities)
+}
+
+// capabilities returns a JSON serialized OPA Capability meant to isolate rego
+// policies from accessing external information, such as hosts or environment
+// variables. If the context already contains the capability, then that is
+// returned as is. Use withCapabilities to pre-populate the context if needed.
+func capabilities(ctx context.Context) (string, error) {
```
joejstuart, line 669: Would it make sense to rename this to `defaultCapabilities` or maybe `noExternalCapabilities` or even `strictCapabilities`?

joejstuart (approved):

internal/evaluator/conftest_evaluator.go
lcarva: No, that's a good idea.

I renamed `capabilities` to `strictCapabilities` as per Joe's suggestion then added the following to the comment:
> The strict capabilities aim to provide a safe environment to execute arbitrary rego policies.

internal/evaluator/conftest_evaluator.go
lcarva: Good call. I had to restructure these lines to avoid having to repeat the function names, but it does log what is being skipped now. Thanks!

internal/evaluator/conftest_evaluator.go
lcarva: Good call. Renamed to `strictCapabilities`.

zregvart (approved):

PR #617, opened 2023-04-13
Support custom rule data

zregvart:
Given custom `ruleData` in a source of an EnterpriseContractPolicy is passed to Conftest. Each `ruleData` is wrapped in a JSON within a top level key `rule_data__configuration__` and stored in a separate file `ruledata<N>.json` within `$dataDir/rule_data` which makes the data available as `data.rule_data__configuration__` in Rego.

Ref. https://issues.redhat.com/browse/HACBS-2042


acceptance/examples/rule_data.rego
```diff
@@ -0,0 +1,14 @@
+package main
+
+deny[result] {
+    not data.rule_data__configuration__
+
+    result := "No custom rule data"
+}
+
+deny[result] {
+    object.get(data.rule_data__configuration__, "custom", "") != "data1"
```
lcarva, line 10: Just to clarify, given the scenario that uses this file, it'll never be the case that both `custom` and `other` are both set, right?

internal/evaluator/conftest_evaluator.go
```diff
@@ -483,6 +483,37 @@ func createConfigJSON(ctx context.Context, dataDir string, p policy.Policy) erro
 	return nil
 }
 
+// addRuleData creates the $dataDir/rule_data__configuration__/ruledata.json
+// files with the provided custom rule data for each source in the data
+// directory
+func addRuleData(ctx context.Context, dataDir string, p policy.Policy) error {
+	if p == nil {
```
lcarva, line 490: What would call this function with a nil policy?

internal/evaluator/conftest_evaluator.go
```diff
@@ -483,6 +483,37 @@ func createConfigJSON(ctx context.Context, dataDir string, p policy.Policy) erro
...
+	}
+
+	ruleDir := path.Join(dataDir, "rule_data")
+
+	fs := utils.FS(ctx)
+	if err := fs.MkdirAll(ruleDir, 0755); err != nil {
+		return err
+	}
+
+	for i, source := range p.Spec().Sources {
```
lcarva, line 501: I don't think this is right. We don't want to mix ruleData across different sources. This should probably be done from `fetchPolicySources` so each evaluator only sees the sources relevant to them.

acceptance/examples/rule_data.rego
zregvart: Well… I realized later that we only invoke conftest once with all sources, I was almost certain that we would invoke it once per source

internal/evaluator/conftest_evaluator.go
zregvart: No idea, copied from `createConfigJSON`

internal/evaluator/conftest_evaluator.go
zregvart: That was my thought exactly, but `Evaluate` runs with all sources, or did I miss this as well?

lcarva: The only place that actually processes `Sources` is `NewApplicationSnapshotImage`. That iterates of the source groups, and for each one, it calls `fetchPolicySources` and `newConftestEvaluator` (alias to `NewConftestEvaluator`). `ValidateImage` then iterates through each evaluator, and calls `Evaluate` on it.

So I think that piece is working as expected since each call to `NewConftestEvaluator` creates its own isolated work dir.

internal/evaluator/conftest_evaluator.go
```diff
@@ -483,6 +483,37 @@ func createConfigJSON(ctx context.Context, dataDir string, p policy.Policy) erro
 	return nil
 }
 
+// addRuleData creates the $dataDir/rule_data__configuration__/ruledata.json
```
simonbaird, line 486: Not sure, but is the file actually just `$dataDir/rule_data/ruledata.json` ?

internal/evaluator/conftest_evaluator.go
simonbaird: It's confusing I think because `c.PolicySources` as used in Evaluate is a list of policy and data source urls for a single "sourceGroup", but the `spec.configuration.sources` is a list of "sourceGroups".

zregvart:
Thanks for the review @lcarva and @simonbaird, yeah, we do end up calling Conftest per source, not as I though with all sources. I've reworked this so that the raw data occurs as another data source via new `source.inlineData`. Fancy having another look?

internal/policy/source/source_test.go
```diff
@@ -93,3 +95,23 @@ func TestGetPolicy(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	file := path.Join(dest, "rule_data.json")
+	assert.FileExists(t, file)
+
+	data, err := os.ReadFile(file)
+	assert.NoError(t, err)
+	assert.Equal(t, []byte("some data"), data)
+
+	assert.Equal(t, "data:application/json;base64,c29tZSBkYXRh", s.PolicyUrl())
```
lcarva, line 116: Cool!

PR #619, opened 2023-04-14
UPDATE npm org to new value of enterprise-contract

robnester-rh:
This is part of renaming our npm org from hacbs-contract to enterprise-contract.


package-lock.json
```diff
@@ -469,312 +472,5 @@
         "node": ">= 6"
       }
     }
-  },
-  "dependencies": {
```
lcarva, line 473: dependencies are missing. Expected/intentional?

lcarva (approved):

package-lock.json
robnester-rh: This was automatically done by npm, so I'm assuming expected?

lcarva: Probably! We'll find out if something breaks :D

zregvart (approved):

PR #626, opened 2023-04-19
Deploy Sigstore Fulcio in the dev environment

zregvart:
This adds Helm to tools/go.mod so we can get updates as needed and configures Sigstore Fulcio deployment against Kubernetes OIDC provider.

To support Helm via tools/go.mod Docker dependencies needed to be downgraded.

This is first stab at getting Fulcio operational, while it does start and seems to be responsive there is more work in getting it fully usable.

Ref. https://issues.redhat.com/browse/HACBS-2082


hack/helm.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+# Copyright 2022 Red Hat, Inc.
```
lcarva, line 2: Update copyright to this year?

hack/development/kustomization.yaml
```diff
@@ -21,3 +21,4 @@ kind: Kustomization
 resources:
   - ../test
   - ../chains
+  - ../sigstore/fulcio
```
lcarva, line 24: Eventually, we want rekor and friends here as well, right?

zregvart: Is Rekor necessary for keyless?

tools/go.mod
```diff
@@ -10,12 +10,19 @@ require (
...
 )
 
 // until https://github.com/grafeas/grafeas/pull/565 is merged/released
 replace github.com/grafeas/grafeas => github.com/zregvart/grafeas v0.0.0-20230102130750-341817cf5037
 
+// Helm v3 is using ORAS 1.2.2 which needs an older version of docker, this
+// helps with:
+//   packages/pkg/mod/oras.land/oras-go@v1.2.2/pkg/auth/docker/login_tls.go:181:30:
+//   assignment mismatch: 3 variables but registry.PingV2Registry returns 2 values
+replace github.com/docker/docker => github.com/docker/docker v20.10.24+incompatible
```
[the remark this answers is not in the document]
zregvart: I have no way of knowing, but I think most likely the code paths required for those will not be invoked

hack/development/kustomization.yaml
lcarva: I _think_ it's possible to skip rekor, but it's definitely not something we want to do in a real environment. For instance, to disable rekor in cosign v2 you have to use the `--insecure-ignore-tlog` parameter. We can iterate here.

simonbaird (commented):
Appreciate the multiple commits and explanations.

zregvart:
I might have a look at disabling certain components, but this should deploy Sigstore services and the demo script shows getting certificate from Fulcio with signing and verifying of a blob.

hack/development/kustomization.yaml
zregvart: Ended up using Rekor, need to see about ctlog and timestamp, perhaps in a subsequent PR...

hack/cosign-sign-blob-fulcio.sh
```diff
@@ -0,0 +1,116 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Installs Tekton Pipeline, the Enterprise Contract Policy custom resource and
```
lcarva, line 18: Outdated description?

hack/ingress/kustomization.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+nameSuffix: ""
+namespace: ingress-nginx
+
+resources:
+  - https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
```
lcarva, line 25: Pin to a release/commit?

hack/setup-dev-environment.sh
```diff
@@ -50,11 +50,22 @@ nodes:
...
+    nodeRegistration:
+      kubeletExtraArgs:
+        node-labels: "ingress-ready=true"
   extraPortMappings:
   - containerPort: ${REGISTRY_PORT:-5000}
     hostPort: ${REGISTRY_PORT:-5000}
     listenAddress: 127.0.0.1
     protocol: TCP
+  - containerPort: 80
+    hostPort: 80
```
lcarva, line 64: It would be nice to not have to use a low port number here.

lcarva (approved):
I think this is a good starting point. My main concern is that there's a lot going on here that will be quite hard for anyone else in the team to debug. A bit of documentation on how to verify things would probably go a long way.

hack/setup-dev-environment.sh
zregvart: Yeah, perhaps in a followup, this would involve plugins, not sure if these are configurable via Helm charts...

lcarva (approved):

PR #627, opened 2023-04-19
Clean up asciidoc links

lcarva:
https://issues.redhat.com/browse/HACBS-2102

### DIFF
--- a/acceptance/examples/reject.rego
+++ b/acceptance/examples/reject.rego
@@ -3,7 +3,9 @@ package main
 
 # METADATA
 # title: Reject rule
-# description: This rule will always fail
+# description: >-
+#   This rule xref:sith.adoc#commandments[will always]
+#   xref:attachment$failing_is_the_new_success.yml[fail]
 # custom:
 #   short_name: rejector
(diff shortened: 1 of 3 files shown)


zregvart (commented):
This only affects the documentation, the AppStudio UI gets the description from the Task YAML logs. Not sure what are we gaining with this?

lcarva:
> This only affects the documentation, the AppStudio UI gets the description from the Task YAML logs. Not sure what are we gaining with this?

Oh? I see this affecting the output of `ec validate image --info` and `ec inspect policy`. The change in `acceptance/examples/reject.rego` shows that's the case. Am I missing something?

zregvart:
> > This only affects the documentation, the AppStudio UI gets the description from the Task YAML logs. Not sure what are we gaining with this?
> 
> Oh? I see this affecting the output of `ec validate image --info` and `ec inspect policy`. The change in `acceptance/examples/reject.rego` shows that's the case. Am I missing something?

My head was in ec-policies repository not in ec-cli... Yeah this works

zregvart (approved):

PR #629, opened 2023-04-20
Support snapshot versions of Chains

zregvart:
The `hack/setup-dev-environment.sh` started failing with Chains upgrade to `v0.15.1-0.20230406140737-50a40c019838` because the `v0.15.1-0.20230406140737-50a40c019838` release YAML doesn't exist. With this the `hack/chains/chains.sh` script tries to fetch different versions, in order and for this example: `v0.15.1` and `v0.15.0`.


lcarva (approved):
Thank you for fixing this!

PR #630, opened 2023-04-20
Revert "Image signature & attestation checks in Successes output"

simonbaird:
The lack of metadata.title and probably also metadata.description in
the result output is breaking the security tab UI.

The revert is intended to get that unbroken quickly. Once the patch
for HACBS-2106 is done, this can be un-reverted.

The change being reverted is from HACBS-2057.

This reverts commit e47b6024fa3cb89f0e437ccd630b6e9bfca2c53a.


lcarva (approved):

mbestavros (approved):
LGTM! Let's fix it :rocket:

PR #631, opened 2023-04-20
Include complete metadata for image signature and attestation checks

mbestavros:
This PR creates and returns full Result objects for image signature and attestation checks, including both Message and Metadata. This should allow those checks to appear in UI alongside other fully-fledged checks.


internal/output/output.go
```diff
@@ -92,81 +92,96 @@ type Output struct {
 
 // SetImageAccessibleCheck sets the passed and result.message fields of the ImageAccessibleCheck to the given values.
 func (o *Output) SetImageAccessibleCheckFromError(err error) {
+	metadata := map[string]interface{}{
+		"code":        "image.accessible",
+		"collections": []string{"minimal"},
```
simonbaird, line 97: Thinking about collection minimal here...

These rules are a bit special since they cannot be skipped, so we maybe have to think about this a little.

E.g. consider if a user does `exclude: '@minimal'` then (correctly) expects this to be skipped.

lcarva: I don't think we should add "collections" here.  Maybe we can set the code to indicate this is some sort of built-in thing?

mbestavros: Maybe something like `builtin.image_accessible`?

simonbaird: I like `builtin`.

features/validate_image.feature
```diff
@@ -41,6 +41,15 @@ Feature: evaluate enterprise contract
           "name": "Unnamed",
           "containerImage": "localhost:(\\d+)/acceptance/ec-happy-day",
           "successes": [
+            {
+              "msg": "Image attestation check passed"
```
simonbaird, line 45: You can try some TDD. All these need metadata key with something in it, right?

zregvart (approved):
LGTM

features/validate_image.feature
```diff
@@ -890,7 +1188,7 @@ Feature: evaluate enterprise contract
...
... value="https:\/\/in-toto.io\/Statement\/v0.1"><\/property>|<property name="metadata.predicateBuildType" value="https:\/\/tekton.dev\/attestations\/chains\/pipelinerun@v2"><\/property>)+<testcase name="builtin.attestation.signature_check: Pass \[title=Attestation signature check passed\]" classname="builtin.attestation.signature_check: Pass \[title=Attestation signature check passed\]" time="0"><\/testcase><testcase name="builtin.attestation.syntax_check: Pass \[title=Attestation syntax check passed\]" classname="builtin.attestation.syntax_check: Pass \[title=Attestation syntax check passed\]" time="0"><\/testcase><testcase name="builtin.image.signature_check: Pass \[title=Image signature check passed\]" classname="builtin.image.signature_check: Pass \[title=Image signature check passed\]" time="0"><\/testcase><testcase name="main.acceptor: Pass" classname="main.acceptor: Pass" time="0"><\/testcase><testcase name="main.rejector: Fails always \[effective_on=2022-01-01T00:00:00Z\]" classname="main.rejector: Fails always \[effective_on=2022-01-01T00:00:00Z\]" time="0"><failure message="Fails always" type=""><!\[CDATA\[Fails always\]\]><\/failure><\/testcase><\/testsuite><\/testsuites>
```
simonbaird, line 1191: oof.

simonbaird (approved):
Lgtm, nice PR.

simonbaird:
Notes:
* We could maybe add a description as well
* Is the `--info` flag required to show the title? For consistency I think the behavior should be that without the `--info` no titles or descriptions are shown, iiuc. (Not a blocker for merging this one either way IMO.)

simonbaird (approved):

PR #641, opened 2023-04-26
Send summary output to both old & new task results

simonbaird:
I'm not sure what consumes this, but to be conservative let's write both result fields for a while. The plan is to transition to TEST_OUTPUT and eventually remove the deprecated HACBS_TEST_OUTPUT field.

Note: I didn't change anything in
pkg/schema/examples/hacbs2_valid.json in this commit. I don't think it will matter, but we might do that later I guess.

JIRA: [HACBS-2160](https://issues.redhat.com/browse/HACBS-1595)

### DIFF
--- a/internal/applicationsnapshot/report.go
+++ b/internal/applicationsnapshot/report.go
@@ -213,6 +213,8 @@ func condensedMsg(results []conftestOutput.Result) map[string][]string {
 
 // toHACBS returns a version of the report that conforms to the
 // HACBS_TEST_OUTPUT format.
+// (Note: the name of the Tekton task result where this generally
+// gets written is now TEST_OUTPUT instead of HACBS_TEST_OUTPUT)
 func (r *Report) toHACBS() hacbsReport {
 	result := hacbsReport{Timestamp: r.created.UTC().Unix()}
 
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -96,6 +96,9 @@ spec:
 
   results:
     - name: HACBS_TEST_OUTPUT
+      description: Short summary of the policy evaluation for each image (Deprecated)
+
+    - name: TEST_OUTPUT
       description: Short summary of the policy evaluation for each image
 
   stepTemplate:
@@ -130,6 +133,10 @@ spec:
         - "--effective-time=$(params.EFFECTIVE_TIME)"
         - "--output"
         - "yaml=$(params.HOMEDIR)/report.yaml"
+        # Write to both the new and the old deprecated result
+        # HACBS_TEST_OUTPUT will be removed in the future
+        - "--output"
+        - "hacbs=$(results.TEST_OUTPUT.path)"
         - "--output"
         - "hacbs=$(results.HACBS_TEST_OUTPUT.path)"
       env:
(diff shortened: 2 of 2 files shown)


simonbaird:
Not sure about the test coverage for this one..

internal/applicationsnapshot/report.go
```diff
@@ -213,6 +213,8 @@ func condensedMsg(results []conftestOutput.Result) map[string][]string {
 
 // toHACBS returns a version of the report that conforms to the
 // HACBS_TEST_OUTPUT format.
+// (Note: the name of the Tekton task result where this generally
+// gets written is now TEST_OUTPUT instead of HACBS_TEST_OUTPUT)
```
lcarva, line 217: It would be nice to update the `hacbs` references in the structs and functions here, e.g. `toTestOutput`. I'd rather the docs were written as if assuming the current state, e.g.
```
// toTestOutput returns a version of the report that conforms to the
// TEST_OUTPUT format. NOTE: This used to be called the 
// HACBS_TEST_OUPUT format.
```

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -130,6 +133,10 @@ spec:
         - "--effective-time=$(params.EFFECTIVE_TIME)"
         - "--output"
         - "yaml=$(params.HOMEDIR)/report.yaml"
+        # Write to both the new and the old deprecated result
+        # HACBS_TEST_OUTPUT will be removed in the future
+        - "--output"
+        - "hacbs=$(results.TEST_OUTPUT.path)"
```
lcarva, line 139: Oh we should probably change the name of the format to `test`? instead of it being `hacbs` ? It's visible to the user and it's going to be confusing.

simonbaird: I did consider that, but a) I feel like "test" is a bad name for a format and b) it's not needed before the deadline, so we could postpone that and figure something out later.

internal/applicationsnapshot/report.go
simonbaird: It's a good point, but I don't mind if we have vestigial traces of hacbs in the code for a while longer. Also we should think of a better name for this format.

zregvart: AppStudio Test results seems appropriate (see https://github.com/redhat-appstudio/book/blob/main/ADR/0013-integration-service-api-contracts.md)

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -95,9 +95,12 @@ spec:
       default: "now"
 
   results:
-    - name: HACBS_TEST_OUTPUT
+    - name: TEST_OUTPUT
       description: Short summary of the policy evaluation for each image
 
+    - name: HACBS_TEST_OUTPUT
+      description: Short summary of the policy evaluation for each image (Deprecated)
+
```
zregvart, line 103: Perhaps we don't reorder in case someone consumes this as `results[0]` and then expects the `name` to be `HACBS_TEST_OUTPUT`

simonbaird: Good idea. Done.

simonbaird:
I filed https://issues.redhat.com/browse/HACBS-2170 re the name change.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
simonbaird: Split off to https://issues.redhat.com/browse/HACBS-2170 .

lcarva (approved):

PR #642, opened 2023-04-26
Strip old effective_on dates from conftest results

mbestavros:
ref: https://issues.redhat.com/browse/HACBS-2143

It turns out that some conftest results already have `effective_on` data present before EC touches it. These `effective_on` dates don't currently get stripped if they fall before the timeout implemented in https://github.com/enterprise-contract/ec-cli/pull/590.

This patch fixes that oversight.

### DIFF
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -43,8 +43,9 @@ import (
 type contextKey string
 
 const (
-	runnerKey       contextKey = "ec.evaluator.runner"
-	capabilitiesKey contextKey = "ec.evaluator.capabilities"
+	runnerKey        contextKey = "ec.evaluator.runner"
+	capabilitiesKey  contextKey = "ec.evaluator.capabilities"
+	effectiveTimeKey contextKey = "ec.evaluator.effective_time"
 )
 
 type CheckResult struct {
@@ -223,6 +224,7 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) (Check
 	}
 
 	effectiveTime := c.policy.EffectiveTime()
+	ctx = context.WithValue(ctx, effectiveTimeKey, effectiveTime)
 
 	// loop over each policy (namespace) evaluation
 	// effectively replacing the results returned from conftest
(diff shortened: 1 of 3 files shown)


mbestavros:
A few thoughts on this patch:

Not sure if this is the cleanest way to do it. My biggest concern organizationally is that `addMetadataToResults`' stated purpose is to add rule metadata, not to touch existing metadata. Maybe we either:

- split the `effective_on` cleanup into its own function (eg. `cleanupConftestResults`) which also gets run alongside `addMetadataToResults` 
- reword the description of `addMetadataToResults` to reflect its expanded scope

internal/evaluator/conftest_evaluator.go
```diff
@@ -427,6 +427,19 @@ func addMetadataToResults(r *output.Result, rule rule.Info) {
 	if len(rule.Collections) > 0 {
 		r.Metadata[metadataCollections] = rule.Collections
 	}
+
+	if effectiveOnString, ok := r.Metadata["effective_on"]; ok {
+		effectiveOnTime, err := time.Parse(effectiveOnFormat, effectiveOnString.(string))
+		if err == nil {
+			// If the rule has been effective for a long time, we'll consider
+			// the effective_on date not relevant and not bother including it
+			if effectiveOnTime.Before(time.Now().Add(effectiveOnTimeout)) {
```
simonbaird, line 436: I think `time.Now()` is wrong here FYI, you have to use `c.policy.EffectiveTime()` somehow.

simonbaird:
How about if line 409 would just always add the effective_on value no matter what. Then the logic from this patch (pretty much as is) would remove it again if needed.

simonbaird:
FYI, I think suggestion about using `c.policy.EffectiveTime()` could be considered out of scope here. It's a separate thing, so we could file a bug to fix it later.

mbestavros:
@simonbaird Good suggestion re. line 409. I've updated with the suggestion.

internal/evaluator/conftest_evaluator.go
```diff
@@ -427,6 +418,20 @@ func addMetadataToResults(r *output.Result, rule rule.Info) {
 	if len(rule.Collections) > 0 {
 		r.Metadata[metadataCollections] = rule.Collections
 	}
+
+	// If the rule has been effective for a long time, we'll consider
+	// the effective_on date not relevant and not bother including it
+	if effectiveOnString, ok := r.Metadata[metadataEffectiveOn]; ok {
```
zregvart, line 424: This makes it a bit safer, `ok` now tells if the key exists and if the value is a string, which makes the cast below unnecessary

```suggestion
	if effectiveOnString, ok := r.Metadata[metadataEffectiveOn].(string); ok {
```

simonbaird (approved):
Lgtm.

PR #647, opened 2023-04-28
Enables TUF in development environment

zregvart:
The TUF HTTP service is available at `http://tuf-server.tuf-system.svc.cluster.local` in-cluster and at `http://tuf.localhost` on the local machine.

The `hack/cosign-sign-blob-fulcio.sh` was modified to perform `cosign initialize` before running `cosign sign-blob` to take advantage of TUF.

Also includes:

[Run cosign locally](https://github.com/enterprise-contract/ec-cli/commit/1bb4c0d3442d9043927a90962b94be00f0d3ed0c)

With TUF, we can now run cosign outside of the cluster, given proper `cosign initialize`.

### DIFF
--- a/hack/cosign-sign-blob-fulcio.sh
+++ b/hack/cosign-sign-blob-fulcio.sh
@@ -22,55 +22,79 @@ set -o errexit
 set -o pipefail
 set -o nounset
 set -o errtrace
+# Enable debugging
+# set -o xtrace
 
 if (( $# != 1 )); then
     echo "usage $0 <file to sign>"
     exit 1
 fi
 
+# Setup for the Cosign image from Sigstore
+IMAGE=gcr.io/projectsigstore/cosign
+SIGSTORE_CONF=/home/nonroot/.sigstore
+
+# For custom container this might look like
+# IMAGE=quay.io/zregvart_redhat/cosign-debug
+# SIGSTORE_CONF=/root/.sigstore
+
 LOGS="$(mktemp --tmpdir)"
 SIGNATURE="$(mktemp --tmpdir)"
 CERTIFICATE="$(mktemp --tmpdir)"
 TLOG_ENTRY="$(mktemp --tmpdir)"
 cleanup() {
+    [ -o xtrace ] && echo -e "📜 \033[1mCosign logs\033[0m" && kubectl logs sign-blob --all-containers
+
     rm -f "${LOGS}" "${SIGNATURE}" "${CERTIFICATE}" "${TLOG_ENTRY}"
+    kubectl delete pod sign-blob > /dev/null 2>&1 || true
 }
 trap cleanup exit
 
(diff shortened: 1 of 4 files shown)


hack/cosign-sign-blob-fulcio.sh
```diff
@@ -22,94 +22,68 @@ set -o errexit
...
+#set -o xtrace
 
 if (( $# != 1 )); then
     echo "usage $0 <file to sign>"
     exit 1
 fi
 
+# We call `cosign initialize` below which will modify the local
+# $HOME/.sigstore/root, so we keep a backup of it and restore it afterwards
+SIGSTORE_ROOT="${HOME}"/.sigstore/root
```
lcarva, line 35: FYI, the environment variable `TUF_ROOT` can be used to instrument cosign to use a different location other than `~/.sigstore/root`. Something like `export TUF_ROOT="$(mktemp -d)"` wouldn't require a backup/restore strategy.

hack/cosign-sign-blob-fulcio.sh
```diff
@@ -22,94 +22,68 @@ set -o errexit
...
-                        "secretName": "ctlog-public-key"
-                    },
-                    "name": "ctlog-public-key"
-                }
-            ]
-        }
-    }' \
-    -- \
-    sign-blob \
+cosign initialize --mirror http://tuf.localhost --root http://tuf.localhost/root.json
```
lcarva, line 58: :heart:

hack/cosign-sign-blob-fulcio.sh
```diff
@@ -22,94 +22,68 @@ set -o errexit
...
+
+cosign sign-blob \
     --oidc-issuer https://kubernetes.default.svc.cluster.local \
-    --fulcio-url=http://fulcio-server.fulcio-system.svc.cluster.local \
-    --identity-token /var/run/secrets/kubernetes.io/serviceaccount/token \
-    --rekor-url http://rekor-server.rekor-system.svc.cluster.local \
+    --fulcio-url http://fulcio.localhost \
```
lcarva, line 62: How is `fulcio.localhost` resolved?

lcarva (approved):

hack/cosign-sign-blob-fulcio.sh
zregvart: You can resolve anything.localhost to the loopback interface (see https://datatracker.ietf.org/doc/html/rfc6761#section-6.3)

hack/cosign-sign-blob-fulcio.sh
zregvart: I thought I tried that, let me doublecheck

zregvart: Yeah, missed to `export`... works like a charm with `TUF_ROOT`

PR #649, opened 2023-05-02
TODO: revert when Tekton Chains bug is fixed

zregvart:
Allows `predicate.materials.uri` not to be in URI syntax.

Ref. https://github.com/tektoncd/chains/pull/792
Ref. https://issues.redhat.com/browse/HACBS-2115


zregvart:
Merging this and spinning up infra-deployments update

PR #650, opened 2023-05-02
Script to update bundle in build-definitions

zregvart:
This script updates the Task bundle reference in the pipeline definition within build-definitions repository.

Ref. https://issues.redhat.com/browse/HACBS-2176

### DIFF
--- a/hack/update-build-definitions.sh
+++ b/hack/update-build-definitions.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+# Copyright 2023 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


hack/update-build-definitions.sh
```diff
@@ -0,0 +1,63 @@
...
+skopeo inspect "docker://quay.io/hacbs-contract/ec-task-bundle:${TASK_BUNDLE_TAG}" --raw > "${MANIFEST}"
+TASK_BUNDLE_DIGEST="$(skopeo manifest-digest "${MANIFEST}")"
+REVISION="$(jq -r '.annotations["org.opencontainers.image.revision"]' "${MANIFEST}")"
+if [[ -n "${REVISION}" && "${REVISION}" != null ]]; then
+    TASK_BUNDLE_TAG="${REVISION}"
+fi
+# Sanity check
+diff \
+    <(skopeo inspect --raw "docker://quay.io/hacbs-contract/ec-task-bundle:${TASK_BUNDLE_TAG}") \
+    <(skopeo inspect --raw "docker://quay.io/hacbs-contract/ec-task-bundle@${TASK_BUNDLE_DIGEST}")
```
simonbaird, line 48: Does it return non zero if there is a diff?

simonbaird (approved):

hack/update-build-definitions.sh
zregvart: Yes, quick example:

```
diff <(echo A) <(echo B) && echo not happening
```

PR #661, opened 2023-05-05
Document the builtin policies via Rego files

zregvart:
See https://github.com/enterprise-contract/ec-policies/pull/464
Ref. https://issues.redhat.com/browse/HACBS-2162

### DIFF
--- a/docs/policy/release/attestation.rego
+++ b/docs/policy/release/attestation.rego
@@ -1,0 +1,45 @@
+#
+# METADATA
+# title: Builtin attestation policies
+# description: >-
+#   This package is responsible for validating attestation signature and syntax.
+#   Note that builtin policies are always included and cannot be skipped
+#   regardless of your policy.
+#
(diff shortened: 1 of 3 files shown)


docs/policy/release/attestation.rego
```diff
@@ -0,0 +1,43 @@
+#
+# METADATA
+# title: Builtin attestation policies
+# description: >-
+#   This package is responsible for validating attestation signature and syntax.
```
simonbaird, line 5: Perhaps something like this:

```suggestion
#   This package is responsible for validating attestation signature and syntax.
#   Note that all builtin policies are always included and cannot be skipped regardless of your policy configuration.
```

simonbaird (approved):

docs/policy/release/attestation.rego
zregvart: I've added the note to the builtin collection and the two packages.

PR #662, opened 2023-05-05
Add policy as yaml file example to command documentation

ralphbean:
I didn't realize this was possible until I read the code.

When working locally, policies in files feel simpler to work with than policies in environment variables.


zregvart:
Thanks!

PR #665, opened 2023-05-08
Ensure dist dir exists in make reference docs

simonbaird:
This fixes a error when running the antora build via `make preview` in the `enterprise-contract.github.io` repo. IIUC it's because the dist directory doesn't exist when `make reference-docs` is run. Should be low risk, so I'm going to insta-merge.


robnester-rh (approved):
Lgtm

simonbaird:
Actually I'm no longer confident this fixes my problem...

PR #679, opened 2023-05-10
update acceptance tests to work with MacOS and podman machine

joejstuart:


acceptance/kubernetes/kind/image.go
```diff
@@ -133,3 +138,14 @@ func (k *kindCluster) buildTaskBundleImage(ctx context.Context) error {
 
 	return nil
 }
+
+func getTag(ctx context.Context) (string, error) {
+	archCmd := exec.CommandContext(ctx, "podman", "version", "-f", "{{.Server.OsArch}}")
+	archOut, archErr := archCmd.CombinedOutput()
+	if archErr != nil {
+		fmt.Print(string(archOut))
```
zregvart, line 146: Probably left over from debugging
```suggestion
```

Makefile
```diff
@@ -85,9 +91,12 @@ acceptance: ## Run all acceptance tests
 	@cd "$${ACCEPTANCE_WORKDIR}"
 	@go run acceptance/coverage/coverage.go .
 	@$(MAKE) build
+ifneq ($(BUILD_IMG_ARCH),$(BUILD_LOCAL_ARCH))
+	@$(MAKE) build-for-test
```
zregvart, line 95: Not sure we need this, test will target `push-image` -> `push_image_$(BUILD_IMG_ARCH)` -> `image_$(BUILD_IMG_ARCH)` -> `$(ALL_SUPPORTED_OS_ARCH)`, or when that is expanded: `push-image` -> `push_image_linux_amd64` -> `image_linux_amd64`. I think...

I think the `kindCluster.buildCliImage` will overwrite the image anyhow. I think...

Makefile
```diff
@@ -85,9 +91,12 @@ acceptance: ## Run all acceptance tests
...
+endif
 	@export COVERAGE_FILEPATH="$${ACCEPTANCE_WORKDIR}"
 	@export COVERAGE_FILENAME="-acceptance"
-	@cd acceptance && go test ./...
+	@cd acceptance && go test -timeout $(ACCEPTANCE_TIMEOUT) ./...
```
zregvart, line 99: Did it block forever? We should check how long this takes on GitHub, might be longer than 10min...

zregvart (approved):
Looks good, couple of nitpicks but LGTM otherwise

Makefile
joejstuart: yeah, just over 10. I changed it to 20m. It didn't block forever, but I found it takes much longer on my Mac so I made it configurable.

zregvart (approved):
Awesome, thanks!

PR #691, opened 2023-05-12
Don't try to build images on darwin (MacOS)

zregvart:
Makes sense, our base image is Red Hat UBI, so we need to exclude darwin (MacOS).


joejstuart (approved):
Works on the Mac. Thanks!

PR #695, opened 2023-05-16
Image relocation acceptance test

zregvart:
Adds an acceptance tests to verify that a signed and attested image can be relocated to a new repository and remain valid.

Ref. https://issues.redhat.com/browse/HACBS-2157


lcarva (approved):
Nice!

PR #697, opened 2023-05-16
Implement an ec init-policies command

mbestavros:
Still needs work, but feedback on what the created directory should look like would be appreciated.

### DIFF
--- a/cmd/initialize/init.go
+++ b/cmd/initialize/init.go
@@ -1,0 +1,31 @@
+// Copyright The Enterprise Contract Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package initialize
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var InitCmd *cobra.Command
+
+func init() {
+	InitCmd = &cobra.Command{
+		Use:   "init",
+		Short: "Initialize a directory for use",
+	}
+	InitCmd.AddCommand(initPoliciesCmd())
+}
--- a/cmd/initialize/init_policies.go
+++ b/cmd/initialize/init_policies.go
@@ -1,0 +1,99 @@
+// Copyright The Enterprise Contract Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Define the `ec init-policies` command
+package initialize
+
+import (
+	"fmt"
+	"path/filepath"
+
+	hd "github.com/MakeNowJust/heredoc"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/enterprise-contract/ec-cli/internal/utils"
+)
+
+func initPoliciesCmd() *cobra.Command {
+	var destDir string
+
+	cmd := &cobra.Command{
+		Use:   "policies --dest-dir <directory-url>",
+		Short: "Initialize a directory with minimal EC scaffolding",
+
+		Long: hd.Doc(`
+			This command creates the necessary files for a minimal EC policy setup in the
+			specified destination directory.
+
+			More information about authoring policies is available in the EC documentation:
+			https://enterprisecontract.dev/docs/ec-policies/authoring.html
+		`),
+
+		Example: hd.Doc(`
+			Initialize the "my-policy" directory with minimal EC policy scaffolding:
+
+			  ec init policies --dest-dir my-policy
+		`),
+
+		Args: cobra.NoArgs,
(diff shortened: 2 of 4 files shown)


zregvart:
If we must implement this in ec-cli lets try to do it in a manner that would make the maintenance easier, quick search of the term ["scaffold" on pkg.go.dev](https://pkg.go.dev/search?q=scaffold), shows at least two promising libraries that we could consider:
 - https://gitlab.com/metakeule/scaffold
 - https://github.com/leaanthony/gosod

joejstuart:
How much of the initialization do we want in `ec`? Right now the `Makefile` and maybe `hack/update-bundles.sh` contain most of the magic we'd want to share. If we want to keep it that way, I'd suggest having `ec` just download the Makefile and run a subcommand of it , like `make setup` or something.

simonbaird:
FWIW I'm okay with merging a very basic implementation as a starting point, then improving it later with https://github.com/leaanthony/gosod perhaps. (Of the two links @zregvart shared, that one's readme seems to make more sense to me.)

That said, https://github.com/leaanthony/gosod does seem like it would be easy to get started with.

Re using the Makefile to bootstrap, it could work maybe, but it feels a little clunky, and there's a versioning problem that goes away it we build the templates into the ec command being run.

cmd/initialize/init_policies.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+			policyPath := filepath.Join(workDir, "sample.rego")
+			file, err := fs.Create(policyPath)
+			if err != nil {
+				log.Debug("Failed to create sample policy!")
+				return err
+			}
+			defer file.Close()
+			fmt.Fprintf(file, "%s", `# Simplest never-failing policy
+			package main
```
simonbaird, line 74: Rather than `main` here, we might need something like `policy.release.my_package` to fit the conventions that ec expects.

cmd/initialize/init_policies.go
```diff
@@ -0,0 +1,100 @@
...
+
+			# METADATA
+			# title: Allow rule
+			# description: This rule will never fail
+			# custom:
+			#   short_name: acceptor
+			#   failure_msg: Always succeeds
+			#   solution: Easy
+			#   collections:
+			#   - A
```
simonbaird, line 84: We'll probably make a more realistic example, but it's not a blocker for this PR. Merging early and playing around with it is going to be useful.

cmd/initialize/init_policies.go
```diff
@@ -0,0 +1,100 @@
...
+				return err
+			}
+			policyPath := filepath.Join(workDir, "sample.rego")
+			file, err := fs.Create(policyPath)
+			if err != nil {
+				log.Debug("Failed to create sample policy!")
+				return err
+			}
+			defer file.Close()
+			fmt.Fprintf(file, "%s", hd.Doc(`# Simplest never-failing policy
```
simonbaird, line 73:
```suggestion
			fmt.Fprintf(file, "%s", hd.Doc(`
			    # Simplest never-failing policy
```

cmd/initialize/init_policies.go
```diff
@@ -0,0 +1,100 @@
...
+			# custom:
+			#   short_name: acceptor
+			#   failure_msg: Always succeeds
+			#   solution: Easy
+			#   collections:
+			#   - A
+			deny[result] {
+			    false
+			    result := "Never denies"
+			}`))
```
simonbaird, line 88:
```suggestion
			    }
			`))
```

simonbaird: We want the trailing line break I think.

cmd/initialize/init_policies_test.go
```diff
@@ -0,0 +1,46 @@
...
+	cmd.SetOut(&buffy)
+
+	cmd.SetArgs([]string{
+		"--dest-dir",
+		"todo",
+	})
+
+	err := cmd.Execute()
+	assert.NoError(t, err)
+}
```
simonbaird, line 46: Are you planning to add more tests here, or is this a case of merge early and often? Ps, I did enjoy writing acceptance tests for the `ec test` command, maybe that's a nice way to test this one.

simonbaird:
One nitpick: Maybe a short directory name consistent with the command name is better, i.e. `cmd/init/init.go` instead of `cmd/initialize/init`.

cmd/initialize/init_policies_test.go
mbestavros: Not for this PR. But I do want to write better tests, so I'll have a look at `ec test` for sure - thanks!

mbestavros:
@simonbaird I tried making the new command as `init/init.go` and I ran into a bunch of issues. It appears that `init` is a special keyword and Go gets confused if we try and create a new package with that name. Hence, `initialize`.

Otherwise, I updated the policy with better indentation.

simonbaird:
> ..init is a special keyword..

Ah, okay. That makes sense then.

simonbaird (approved):

cmd/initialize/init.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2022 Red Hat, Inc.
```
lcarva, line 1: `// Copyright The Enterprise Contract Contributors`

Check other files as well.

cmd/initialize/init_policies.go
```diff
@@ -0,0 +1,102 @@
...
+	"path/filepath"
+
+	hd "github.com/MakeNowJust/heredoc"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/enterprise-contract/ec-cli/internal/utils"
+)
+
+func initCmd() *cobra.Command {
```
lcarva, line 32: NIT: this is a bit odd because `init` is a command on its own. Following the convention we have used in the other commands here, consider naming this `initPoliciesCmd`. It would make this line in `init.go` a little bit easier to read: `InitCmd.AddCommand(initCmd())` -> `InitCmd.AddCommand(initPoliciesCmd())`

cmd/initialize/init_policies.go
```diff
@@ -0,0 +1,102 @@
...
+		`),
+
+		Args: cobra.NoArgs,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			fs := utils.FS(ctx)
+			workDir := filepath.Join(".", destDir)
```
lcarva, line 60: Why the `Join` here? This turns absolute paths into relative ones, `/tmp/foo` -> `./tmp/foo`. This is a surprising behavior.

lcarva (approved):

PR #709, opened 2023-05-19
Use go-snaps in acceptance tests

zregvart:
Now instead of inlining the expected output and our custom diff implementation we can use go-snaps which makes it easier to maintain the expected stdout/stderr output of the CLI.

[Use specific hostnames for stubbed services](https://github.com/enterprise-contract/ec-cli/commit/e5a282bf0a7b5e614e271ac9ecb03da14ed89edb)

Uses `rekor.localhost`, `tuf.localhost` and `apiserver.localhost` instead of `localhost`, this helps with snapshots and creating a distinction.


joejstuart (approved):
Very neat!

PR #713, opened 2023-05-24
Rule dependencies

zregvart:
A rule can specify a custom annotation `depends_on` with a string or a list of strings containing the code(s) of the rule(s) that a particular rule depends on. When a rule with dependents is reported as a failure, warning or it is skipped, dependent rules are filtered out and not reported.

Ref. https://issues.redhat.com/browse/HACBS-2208

Also includes:

[Refactor successes computation](https://github.com/enterprise-contract/ec-cli/commit/4250c7ad9e656c9dcfee056c6e85746aaca4bfab)

Pulls in the successes computation into a separate method.

[Refactor and fix consistency](https://github.com/enterprise-contract/ec-cli/commit/0c58b214c75cdce6a58d9d56cc93b5ef9d51327e)

Two rules with the same code, possibly by adding the same policy source twice with different versions, may interfere with each other. Do not allow that. Collect successes when looping over the `output.CheckResult` rather than assign to the first result.

### DIFF
--- a/features/__snapshots__/validate_image.snap
+++ b/features/__snapshots__/validate_image.snap
@@ -228,3 +228,12 @@ ${TEMP}/ec-work-${RANDOM}/policy/${RANDOM}/main.rego:29: rego_type_error: undefi
 [happy day:stderr - 1]
 
 ---
+
+[rule dependencies:stdout - 1]
+{"success":false,"components":[{"name":"Unnamed","containerImage":"${REGISTRY}/acceptance/image@${REGISTRY_acceptance/image:latest_HASH}","violations":[{"msg":"Failure","metadata":{"code":"pkg.fails"}}],"warnings":[{"msg":"Warning","metadata":{"code":"pkg.warns"}}],"successes":[{"msg":"Pass","metadata":{"code":"builtin.attestation.signature_check"}},{"msg":"Pass","metadata":{"code":"builtin.attestation.syntax_check"}},{"msg":"Pass","metadata":{"code":"builtin.image.signature_check"}}],"success":false,"signatures":${ATTESTATION_SIGNATURES_JSON}}],"key":${known_PUBLIC_KEY_JSON},"policy":{"sources":[{"policy":["git::https://${GITHOST}/git/with-dependencies.git"]}],"publicKey":"${known_PUBLIC_KEY}"},"ec-version":"${EC_VERSION}"}
+---
+
+[rule dependencies:stderr - 1]
+Error: success criteria not met
+
+---
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -228,7 +228,6 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, inputs []string) (Check
 
 	// loop over each policy (namespace) evaluation
 	// effectively replacing the results returned from conftest
-	ruleCollection := make(map[string]bool)
 	for i, result := range runResults {
 		log.Debugf("Evaluation result at %d: %#v", i, result)
 		warnings := []output.Result{}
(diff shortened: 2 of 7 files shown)


features/__snapshots__/validate_image.snap
```diff
@@ -228,3 +228,12 @@ ${TEMP}/ec-work-${RANDOM}/policy/${RANDOM}/main.rego:29: rego_type_error: undefi
 [happy day:stderr - 1]
 
 ---
+
+[rule dependencies:stdout - 1]
+{"success":false,"components":[{"name":"Unnamed","containerImage":"${REGISTRY}/acceptance/image@${REGISTRY_acceptance/image:latest_HASH}","violations":[{"msg":"Failure","metadata":{"code":"pkg.fails"}}],"warnings":[{"msg":"Warning","metadata":{"code":"pkg.warns"}}],"successes":[{"msg":"Pass","metadata":{"code":"builtin.attestation.signature_check"}},{"msg":"Pass","metadata":{"code":"builtin.attestation.syntax_check"}},{"msg":"Pass","metadata":{"code":"builtin.image.signature_check"}}],"success":false,"signatures":${ATTESTATION_SIGNATURES_JSON}}],"key":${known_PUBLIC_KEY_JSON},"policy":{"sources":[{"policy":["git::https://${GITHOST}/git/with-dependencies.git"]}],"publicKey":"${known_PUBLIC_KEY}"},"ec-version":"${EC_VERSION}"}
```
lcarva, line 233: tangent: since we're using snapshots (which are awesome btw!), maybe it makes sense to switch most scenarios to run ec with yaml output?

internal/evaluator/conftest_evaluator.go
```diff
@@ -65,6 +65,61 @@ func (c CheckResults) ToConftestResults() []output.CheckResult {
...
+// a result reported as failure, warning or skipped. Dependencies are declared
+// by setting the metadata via metadataDependsOn.
+func (c *CheckResults) trim() {
+	// holds codes for all failures, warnings or skipped rules, as a map to ease
+	// the lookup, any rule that depends on a reported code will be removed from
+	// the results
+	reported := map[string]bool{}
+
+	for _, checks := range *c {
+		for _, results := range [][]output.Result{checks.Failures, checks.Warnings, checks.Skipped} {
```
lcarva, line 78: I forget why this is a slice of slices. Is the first dimension meant to represent source groups? I wonder if we should take care to ensure one source group doesn't affect another one. I _think_ that would be the right behavior.

lcarva (approved):
Very nice!

internal/evaluator/conftest_evaluator.go
```diff
@@ -65,6 +65,61 @@ func (c CheckResults) ToConftestResults() []output.CheckResult {
...
+		}
+
+		return trimmed
+	}
+
+	for i, checks := range *c {
+		(*c)[i].Failures = trimOutput(checks.Failures)
+		(*c)[i].Warnings = trimOutput(checks.Warnings)
+		(*c)[i].Skipped = trimOutput(checks.Skipped)
+		(*c)[i].Successes = trimOutput(checks.Successes)
```
simonbaird, line 119: Is it obvious to everyone why we can't do `checks.Failures = trimOutput(checks.Failures)` and so on? Maybe add a comment if not..?

internal/evaluator/conftest_evaluator.go
zregvart: This is just a shorter way of writing the loop, so I don't need to loop over checks.Failures, checks.Warnings and checks.Skipped in three loops...

IIRC we have an Evaluator per source group, but we should test (created [HACBS-2257](https://issues.redhat.com/browse/HACBS-2257))

internal/evaluator/conftest_evaluator.go
zregvart: `c[i]` would be a compile error, see [an example](https://go.dev/play/p/Q5xIBKIxqTh)

zregvart:
I think we need to talk about [Refactor and fix consistency](https://github.com/enterprise-contract/ec-cli/commit/0c58b214c75cdce6a58d9d56cc93b5ef9d51327e) bit, so I won't merge this immediately...

simonbaird:
> I think we need to talk about [Refactor and fix consistency](https://github.com/enterprise-contract/ec-cli/commit/0c58b214c75cdce6a58d9d56cc93b5ef9d51327e) bit, so I won't merge this immediately...

Discussed at standup, I think we agreed it was a sensible behavior change.

internal/evaluator/conftest_evaluator.go
simonbaird: Ok.

simonbaird (approved):

PR #727, opened 2023-05-30
remove deprecated HACBS_TEST_OUTPUT

joejstuart:


cmd/validate/image.go
```diff
@@ -147,9 +147,9 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			  ec validate image --image registry/name:tag --output json=<path>
 
-			Write output in YAML format to stdout and in HACBS format to a file
+			Write output in YAML format to stdout and in test format to a file
 
-			  ec validate image --image registry/name:tag --output yaml --output hacbs=<path>
+			  ec validate image --image registry/name:tag --output yaml --output test=<path>
```
lcarva, line 152: I'm assuming this PR is to address https://issues.redhat.com/browse/HACBS-2198 (It would be nice to mention this in the commit message!)

I think changing the output name is outside of its scope. But it is part of https://issues.redhat.com/browse/HACBS-2170

There are some special considerations we must take to make that change. Let's make sure we're addressing them. We can also not worry about renaming this right now.

lcarva (approved):

zregvart (approved):

PR #731, opened 2023-05-31
Update the rhtap tekton pipelines

lcarva:


joejstuart (approved):

lcarva:
Requires https://github.com/redhat-appstudio/build-definitions/pull/416

zregvart (approved):

zregvart:
Should we enable renovate bot here? Or would it collide with dependabot?

lcarva:
> 

Yes, we should. I think there's a way to run both. May require some tweaks.

PR #733, opened 2023-06-01
Add test coverage in conftest_evaluator

zregvart:
Adds a unit test that exercises Conftest and the surrounding logic on augmenting the CheckResults with successes.

### DIFF
--- a/internal/evaluator/__testdir__/a.rego
+++ b/internal/evaluator/__testdir__/a.rego
@@ -1,0 +1,28 @@
+# A set of policies
+package a
+
+# METADATA
+# custom:
+#   short_name: failure
+deny[result] {
+	result := {
+		"code": "a.failure",
+		"msg": "Failure!",
+	}
+}
+# METADATA
+# custom:
+#   short_name: warning
+warn[result] {
+	result := {
+		"code": "a.warning",
+		"msg": "Warning!",
+	}
+}
+# METADATA
+# custom:
+#   short_name: success
+deny[result] {
+	false
+	result := "Success!"
+}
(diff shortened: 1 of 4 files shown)


internal/evaluator/__testdir__/a.rego
```diff
@@ -0,0 +1,28 @@
+# A set of policies
+package a
+
+# METADATA
+# custom:
+#   short_name: failure
+deny[result] {
+	result := {
+		"code": "main.failure",
```
lcarva, line 9: Should the code use the package name instead of `main` to better simulate our policies?

internal/evaluator/__testdir__/b.rego
```diff
@@ -0,0 +1,28 @@
+# B set of policies
+package b
+
+# METADATA
+# custom:
+#   short_name: failure
+deny[result] {
+	result := {
+		"code": "main.failure",
```
lcarva, line 9: I think this is highlighting a feature gap. Notice how if this rule fails, the `code` is `main.failure`, but if it succeeds, it would be `b.failure`. It would be confusing for a user to experience this behavior.

It's less of a problem when using policies from ec-policies because we ensure those things match there, but a potential pitfall for anyone writing their own policies.

lcarva (approved):
Very nice

internal/evaluator/__testdir__/b.rego
zregvart: Right, copy-pasta error, I used to have just the `main` package, then I created the `a` and `b` packages. Good catch!

PR #736, opened 2023-06-01
Initialize TUF root in ec task

lcarva:
https://issues.redhat.com/browse/HACBS-2072

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -84,6 +84,12 @@ func initializeScenario(sc *godog.ScenarioContext) {
 	})
 }
 
+func initializeSuite(ctx context.Context) func(*godog.TestSuiteContext) {
+	return func(tsc *godog.TestSuiteContext) {
+		kubernetes.InitializeSuite(ctx, tsc)
+	}
+}
+
 // setupContext creates a Context prepopulated with the *testing.T and *persist
 // values
 func setupContext(t *testing.T) context.Context {
@@ -109,20 +115,23 @@ func TestFeatures(t *testing.T) {
 		t.Error(err)
 	}
 
+	ctx := setupContext(t)
+
 	opts := godog.Options{
 		Format:         "pretty",
 		Paths:          []string{featuresDir},
 		Randomize:      -1,
 		Concurrency:    runtime.NumCPU(),
 		TestingT:       t,
-		DefaultContext: setupContext(t),
+		DefaultContext: ctx,
 		Tags:           *tags,
 		NoColors:       *noColors,
 	}
 
 	suite := godog.TestSuite{
-		ScenarioInitializer: initializeScenario,
-		Options:             &opts,
+		ScenarioInitializer:  initializeScenario,
+		TestSuiteInitializer: initializeSuite(ctx),
+		Options:              &opts,
 	}
 
 	if suite.Run() != 0 {
--- a/acceptance/kubernetes/kind/kind.go
+++ b/acceptance/kubernetes/kind/kind.go
@@ -386,22 +386,34 @@ func (k *kindCluster) Stop(ctx context.Context) (context.Context, error) {
 	clusterGroup.Done()
 	logger.Log("[Stop] Released cluster to group")
 
+	return ctx, nil
+}
+
+func Destroy(ctx context.Context) {
+	logger, _ := log.LoggerFor(ctx)
 	destroy.Do(func() {
-		logger.Log("[Stop] Stopping cluster")
+		if globalCluster == nil {
+			logger.Log("[Destroy] Skipping global cluster destruction")
+			return
+		}
+		logger.Log("[Destroy] Destroying global cluster")
 
 		// wait for other cluster consumers to finish
+		logger.Log("[Destroy] Waiting for all consumers to finish")
 		clusterGroup.Wait()
-		logger.Log("[Stop] Last cluster consumer finished")
+		logger.Log("[Destroy] Last global cluster consumer finished")
 
 		defer func() {
-			kindDir := path.Join(k.kubeconfigPath, "..")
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -14,11 +14,29 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+FROM registry.access.redhat.com/ubi9/ubi-minimal as downloads
```
lcarva, line 17: At some point, I'm hoping that Red Hat starts packaging cosign as its own image. This will allow us to use that image directly (hopefully) in the task instead of having to install cosign here.

joejstuart: We could pin to an appstudio-utils version, but this seems like the safest plan.

lcarva: I definitely don't want to use an appstudio image here.

zregvart:
Scenarios run concurrently so [this](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:409) is probably from another scenario.

The odd thing is that [kindCluster.Up](https://github.com/enterprise-contract/ec-cli/blob/f82ba3c39dbfa1ffb127b4b72aa5b477926a55eb/acceptance/kubernetes/kind/kind.go#L111) is returning `false`, most likely because the cluster has [been stopped](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2865).

So what happened was that scenario with thread id=8 [started](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2853) and [stopped](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2862) the cluster before other consumers of the cluster thread ids=[10](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2867),[11](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2869) requested a cluster, so the wait group was empty when the cluster stop was performed.

Not sure what the best solution would be here, possibly re-initializing the `create` & `destroy` `sync.Once`s.

A quick and dirty alternative could be to sleep for a couple of seconds [here](https://github.com/enterprise-contract/ec-cli/blob/f82ba3c39dbfa1ffb127b4b72aa5b477926a55eb/acceptance/kubernetes/kind/kind.go#L384) and wait if there are any other takers of the cluster.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -112,6 +117,22 @@ spec:
       command: [ec]
       args:
         - version
+    - name: initialize-tuf
```
zregvart, line 120: I was wondering if `when` was available on Steps, seems it is available only on Tasks, would be nice not to have shell code here. Also I wondered if this should be in a sidecar instead.

zregvart (approved):

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
lcarva: Yeah, `when` on Steps would've been nice. I don't see anything in the code that would support that unfortunately. I think the bash is pretty minimal and there's sufficient coverage in the acceptance tests though. Let's try that out first.

I don't think a sidecar is the right pattern here. My understanding is that a sidecar is meant when you need something to be running throughout the lifecycle of a Task, e.g. setup a proxy that is used in the steps. Here, we just need to initialize TUF root once and early. The Steps interact with it via the filesystem which does already persist throughout the Task lifecycle.

lcarva:
> Scenarios run concurrently so [this](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:409) is probably from another scenario.
> 
> The odd thing is that [kindCluster.Up](https://github.com/enterprise-contract/ec-cli/blob/f82ba3c39dbfa1ffb127b4b72aa5b477926a55eb/acceptance/kubernetes/kind/kind.go#L111) is returning `false`, most likely because the cluster has [been stopped](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2865).
> 
> So what happened was that scenario with thread id=8 [started](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2853) and [stopped](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2862) the cluster before other consumers of the cluster thread ids=[10](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2867),[11](https://github.com/enterprise-contract/ec-cli/actions/runs/5149408977/jobs/9273470961#step:5:2869) requested a cluster, so the wait group was empty when the cluster stop was performed.
> 
> Not sure what the best solution would be here, possibly re-initializing the `create` & `destroy` `sync.Once`s.
> 
> A quick and dirty alternative could be to sleep for a couple of seconds [here](https://github.com/enterprise-contract/ec-cli/blob/f82ba3c39dbfa1ffb127b4b72aa5b477926a55eb/acceptance/kubernetes/kind/kind.go#L384) and wait if there are any other takers of the cluster.

I'll play with it a bit. The odd thing is that I cannot reproduce this locally :disappointed:

zregvart:
Looks like the same issue occurs #739

lcarva:
@zregvart, check out the second commit on the PR. It's one way to fix the timing issue with the kind cluster.

zregvart:
@lcarva thanks, this is awesome, I could not figure out how to configure an `After` for a suite, really neat solution this is

PR #739, opened 2023-06-02
Test coverage around the `STRICT` Task parameter

zregvart:
In addition to adding coverage the Task acceptance tests were converted to use snapshots, with variable substitutions not to succumb to flakiness.

Snapshots support per-step logs, and currently the `report` step is asserted, others don't offer much value to test (e.g.`assert` logs are evident from the Task status) or change frequently (e.g. `version`).

Ref. https://issues.redhat.com/browse/HACBS-2261

### DIFF
--- a/acceptance/kubernetes/kind/kubernetes.go
+++ b/acceptance/kubernetes/kind/kubernetes.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"strings"
 	"time"
 
 	ecc "github.com/enterprise-contract/enterprise-contract-controller/api/v1alpha1"
@@ -35,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 
+	"github.com/enterprise-contract/ec-cli/acceptance/crypto"
 	"github.com/enterprise-contract/ec-cli/acceptance/kubernetes/types"
 	"github.com/enterprise-contract/ec-cli/acceptance/kustomize"
 	"github.com/enterprise-contract/ec-cli/acceptance/testenv"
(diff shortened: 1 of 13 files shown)


acceptance/kubernetes/kind/kubernetes.go
```diff
@@ -195,6 +216,8 @@ func (k *kindCluster) CreateNamespace(ctx context.Context) (context.Context, err
 
 	t.namespace = namespace.GetName()
 
+	t.registry = fmt.Sprintf("registry.image-registry.svc.cluster.local:%d", k.registryPort)
```
lcarva, line 219: Use `k.Registry()` here?

acceptance/kubernetes/kind/kubernetes.go
```diff
@@ -45,6 +47,25 @@ import (
 func (k *kindCluster) createPolicyObject(ctx context.Context, specification string) (*ecc.EnterpriseContractPolicy, error) {
 	t := testenv.FetchState[testState](ctx)
 
+	specification = os.Expand(specification, func(key string) string {
+		if strings.HasSuffix(key, "_PUBLIC_KEY") {
+			publicKeys := crypto.PublicKeysFrom(ctx)
+			keyName := strings.TrimSuffix(key, "_PUBLIC_KEY")
+			publicKey, ok := publicKeys[keyName]
+			if !ok {
+				panic(fmt.Sprintf("Unexpected or no value found for public key named: `%s`", keyName))
```
lcarva, line 56: Should we just return an error here? Same below.

acceptance/kubernetes/kubernetes.go
```diff
@@ -233,6 +238,69 @@ func logTaskOutput(ctx context.Context, c ClusterState) error {
...
+	}
+
+	vars := map[string]string{
+		"REGISTRY": r,
+	}
+
+	publicKeys := crypto.PublicKeysFrom(ctx)
+	for name, key := range publicKeys {
+		// account for various indentations
+		vars[fmt.Sprintf("__%s_PUBLIC_KEY", name)] = indent(key, 2)
```
lcarva, line 265: Clever :smile:

lcarva (approved):

acceptance/kubernetes/kind/kubernetes.go
zregvart: Tots :+1:

acceptance/kubernetes/kind/kubernetes.go
zregvart: this is a `func(string) string`, so we can't...

lcarva: oh I missed the inner function. Sounds good.

PR #741, opened 2023-06-05
`data` output format

zregvart:
The `ec validate image` command can now output the data used in policy evaluation with the `--output data=<path>` parameter. The data is written to the provided path as a YAML file containing an array of data passed to each evaluation, equivalent of a source group.

A step has been added to the v-e-c to output the data as a result.

Ref. https://issues.redhat.com/browse/HACBS-2132


cmd/validate/image.go
```diff
@@ -151,6 +152,11 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			  ec validate image --image registry/name:tag --output yaml --output appstudio=<path>
 
+			Write output the data used in policy evaluation as YAML file
```
lcarva, line 155: This sentence reads oddly to me. Maybe pick either "write" or "output" but not both?

internal/evaluator/conftest_evaluator.go
```diff
@@ -120,8 +122,10 @@ func (c *CheckResults) trim() {
 	}
 }
 
+type Data map[string]any
```
lcarva, line 125: Does it make sense for this to be `map[string]json.RawMessage` ?

https://pkg.go.dev/encoding/json#RawMessage

internal/evaluator/conftest_evaluator.go
```diff
@@ -149,6 +153,46 @@ type conftestEvaluator struct {
...
+
+	ids := []string{} // everything
+
+	var d any
+	d, err = store.Read(ctx, txn, ids)
+	if err != nil {
+		return
+	}
+
+	data, _ = d.(map[string]any) // if we fail here we don't have the data to return, and mostly we won't
```
lcarva, line 191: If an error happens, don't we want things to fail?

lcarva (approved):
Have you had a chance to try this out with ruleData and a "complex" data source like our default one (data in multiple files)?

cmd/validate/image.go
zregvart: Yeah, how about: `Write the data used in the policy evaluation to a file in YAML format`

internal/evaluator/conftest_evaluator.go
zregvart: This could only fail if we don't pass any data to Conftest or if the data passed
is not a `map[string]any`, with this in those cases `data` is `nil`. So this
prevents a panic in those cases.
Not passing any data to Conftest would mean that we didn't even pass effective
time.
We could err here, but I don't think that error could happen

internal/evaluator/conftest_evaluator.go
zregvart: What we get out of Conftest is something like
`map[string]map[string]map[string]int64`, so that would not fit...

internal/evaluator/conftest_evaluator.go
lcarva: Personally, I'd rather propagate an error even if I don't think it's possible to happen. If I'm right, the error never shows up and it doesn't matter, but if I'm wrong, I'll have a chance of understanding what went wrong. Just my 2 cents. Not a blocker for this PR.

zregvart: I did end up propagating the error...

zregvart:
> Have you had a chance to try this out with ruleData and a "complex" data source like our default one (data in multiple files)?

Yeah, the output is a bit funky -- bits repeat, I was hesitant to merge the data...

lcarva:
> 

ack. We can sort this out later. Right now this is just really helpful information.

zregvart:
Codecov is lying

PR #745, opened 2023-06-05
Remove dependency on `k8s.io/kubernetes`

zregvart:
We should not depend on `k8s.io/kubernetes` as a library, so this replaces it with `github.com/jstemmer/go-junit-report/v2`. This changes the JUnit output slightly, but otherwise should not have an adverse effect.


acceptance/go.mod
```diff
@@ -36,13 +36,14 @@ require (
...
 	k8s.io/apimachinery v0.27.2
 	k8s.io/client-go v0.26.3
-	k8s.io/kubernetes v1.27.2
 	sigs.k8s.io/kind v0.19.0
 	sigs.k8s.io/kustomize/api v0.13.4
 	sigs.k8s.io/kustomize/kyaml v0.14.2
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/jstemmer/go-junit-report/v2 v2.0.0
```
lcarva, line 45: Should this go on the `require` block above?

lcarva (approved):

PR #748, opened 2023-06-06
Honor -persist flag in kind cluster destruction

lcarva:


acceptance/kubernetes/kubernetes.go
```diff
@@ -269,6 +272,8 @@ func AddStepsTo(sc *godog.ScenarioContext) {
 
 func InitializeSuite(ctx context.Context, tsc *godog.TestSuiteContext) {
 	tsc.AfterSuite(func() {
-		kind.Destroy(ctx)
+		if !persisted {
```
zregvart, line 275: So we don't need the global variable

```suggestion
		if !testenv.Persisted(ctx) {
```

zregvart (commented):
I don't think we need the global variable

acceptance/kubernetes/kubernetes.go
lcarva: I don't think this instance of context will have the right info. This is the "default" context. Let me try it.

lcarva: It worked! Thank you.

zregvart (approved):

PR #750, opened 2023-06-07
ADD feature to hide successes in output by default

robnester-rh:
This commit addresse HACBS-2262 and disables the display of the multitude of "successes" in the output by default. This display can be enabled by adding the "--show-successes" flag if need.


mbestavros (approved):
LGTM!

cmd/validate/image.go
```diff
@@ -277,8 +277,9 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 					// Skip on err to not panic. Error is return on routine completion.
 					if err == nil {
 						res.component.Violations = out.Violations()
+						showSuccesses, _ := cmd.Flags().GetBool("show-successes")
 						res.component.Warnings = out.Warnings()
-						res.component.Successes = out.Successes()
+						res.component.Successes = out.Successes(showSuccesses)
```
zregvart, line 282: I think this is better suited for the reporting bit, if we want to keep it here why not do something like:
```suggestion
						if showSuccesses {
							res.component.Successes = out.Successes()
						}
```

robnester-rh: ack. Updated.

zregvart (approved):

PR #751, opened 2023-06-09
Don't write data output as a Task result

lcarva:
The data can easily reach the 4k limit imposed by Tekton on TaskRun results. If this limit is reached, the TaskRun fails.

Let's instead just dump this value into the logs which was the initial intent anyways.


joejstuart (approved):

zregvart:
I’m expecting a snapshot to fail

lcarva:
No snapshot updates needed. I'll file a follow up PR to make sure Task results do get snapshoted as well.

PR #752, opened 2023-06-09
Remove sig/att not found error wrapper for keyless

lcarva:
https://issues.redhat.com/browse/HACBS-2161


zregvart (approved):
LGTM with the stated intent, though I think we should do this for all cases, would make it easier to troubleshoot having the whole error message from Cosign

lcarva:
> LGTM with the stated intent, though I think we should do this for all cases, would make it easier to troubleshoot having the whole error message from Cosign

The issues was that the message from cosign is so cryptic that it makes it hard for a regular user to understand what's happening.

PR #761, opened 2023-06-14
Snapshot TaskRun results in acceptance tests

zregvart:
Adds the ability to snapshot TaskRun results in acceptance tests and adds it to the current Task acceptance tests.

Ref. https://issues.redhat.com/browse/HACBS-2295

### DIFF
--- a/features/__snapshots__/task_validate_image.snap
+++ b/features/__snapshots__/task_validate_image.snap
@@ -64,13 +64,18 @@ components:
       code: builtin.image.signature_check
       title: Image signature check passed
     msg: Pass
+  warnings:
   - metadata:
       code: test.no_skipped_tests
       depends_on:
       - test.test_data_found
       description: Reports any test that has its result set to "SKIPPED".
+      solution: There is a test that was skipped. Make sure that each task with a
+        result named 'TEST_OUTPUT' was not skipped. You can find which test was skipped
+        by examining the 'result' key in the 'TEST_OUTPUT'.
+      term: skipped
       title: No tests were skipped
-    msg: Pass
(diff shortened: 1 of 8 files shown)


features/__snapshots__/task_validate_image.snap
```diff
@@ -259,3 +259,31 @@ success: true
...
+[Strict with failures:results - 1]
+{}
+---
+
+[Non strict with failures:results - 1]
+{"TEST_OUTPUT":"{\"timestamp\":\"${TIMESTAMP}\",\"namespace\":\"\",\"successes\":0,\"failures\":1,\"warnings\":0,\"result\":\"FAILURE\"}"}
+---
+
+[Strict with warnings:results - 1]
+{"TEST_OUTPUT":"{\"timestamp\":\"${TIMESTAMP}\",\"namespace\":\"\",\"successes\":1,\"failures\":0,\"warnings\":0,\"result\":\"SUCCESS\"}"}
```
lcarva, line 276: Is this a bug on this test and the one below? The name implies there should be at least one warning.

lcarva (approved):

features/__snapshots__/task_validate_image.snap
zregvart: Most likely a bug in the test, will investigate

zregvart:
Will investigate and undraft when I figure out why we don't show warnings in a test that tests for warnings

zregvart:
Should be okay now, fixed the test cases with warnings

lcarva (approved):
Nice! Thanks for addressing the test case :pray:

PR #774, opened 2023-06-16
Introduce the concept of validators

lcarva:
The idea is to move the various checks performed by the `ec validate image` command into their own package under validators. By doing so we enabled a couple of things:

1. We can eventually tie validators to source groups. This would allow a policy configuration to determine exactly which validators should be executed.

2. We can also add many more validators that rely on specific data without having to worry about wasted resources when policies don't leverage such data. A validator is reponsible for fetching the data it needs. If it's not configured, the data is not fetched.

There's still a lot to go. This commit introduces a single validator responsible for verifying image signatures. The beginning of a slsa provenance validator is added for inspiration.

There are some subtle changes here that made sense to do at this time. For example, some of the logic was moved away from the output package and into the validator. This was intentional as I believe that is a better home.

The abstraction is still not perfect. For instance, we should avoid exposing conftest and consign structs everywhere. I decided to avoid tackling this existing problem to avoid a larger patch. But it's something we should take care of at some point.


lcarva:
Created as a draft because the unit tests need some work. However! The acceptance tests do pass! (at least on my machine :tm: )

internal/sigstore/client.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2022 Red Hat, Inc.
```
lcarva, line 1: This is a direct copy from `internal/evaluation_target/application_snapshot_image/client.go`. This was needed to avoid circular imports. This PR needs to be improved to delete that one in favor of this. I consider it a blocker.

internal/sigstore/errors.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2022 Red Hat, Inc.
```
lcarva, line 1: Mostly copied from output. Needs to be consolidated as part of this PR.

lcarva:
@enterprise-contract/devs any feedback here? :pray:

internal/validator/validator.go
```diff
@@ -0,0 +1,70 @@
...
+	Successes  []confOutput.Result
+	Signatures []output.EntitySignature
+	// TODO: Add support for Attestations
+}
+
+type imageInitializer func(Options) ImageValidator
+
+var imageValidators = map[string]imageInitializer{}
+
+func RegisterImageValidator(name string, v imageInitializer) error {
```
joejstuart, line 56: I'm trying to understand registering instead of just allowing the calling code create the validators it needs. Is this so we can enable/disable validators at some point?

simonbaird:
One thought I had: I was imagining the signature check as more transparent aspect of the "fetching", rather than being its own separate validator. Compare to curl, when we use https we expect the fetch to fail on TLS errors without that being a separate check done explicitly. We might design it so anything being fetched with cosign has an implicit signature check. Perhaps with a --insecure flag to disable that.

internal/validator/validator.go
lcarva: Yeah. Exactly! It's about decoupling the implementation from its usage.

simonbaird:
Overall, looks good.

lcarva:
> One thought I had: I was imagining the signature check as more transparent aspect of the "fetching", rather than being its own separate validator. Compare to curl, when we use https we expect the fetch to fail on TLS errors without that being a separate check done explicitly. We might design it so anything being fetched with cosign has an implicit signature check. Perhaps with a --insecure flag to disable that.

Consider keyless where we want to pass the certificate used to rego. By making it its own validator, we can treat it just like any other check that is performed. If you don't care about the signature, then omit the signature validator.

internal/validator/validator.go
```diff
@@ -0,0 +1,70 @@
...
+	confOutput "github.com/open-policy-agent/conftest/output"
+
+	"github.com/enterprise-contract/ec-cli/internal/output"
+	"github.com/enterprise-contract/ec-cli/internal/policy"
+	"github.com/enterprise-contract/ec-cli/internal/policy/source"
+)
+
+type ImageValidator interface {
+	Validate(context.Context, name.Reference) *ImageResult
+}
```
zregvart, line 33: Should this be generic?

```suggestion
type Validator interface {
	Validate(context.Context, Artifact) *Result
}
```

internal/validator/validator.go
```diff
@@ -0,0 +1,70 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package validator
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	confOutput "github.com/open-policy-agent/conftest/output"
```
zregvart, line 24: We probably don't want to depend directly from the validator on these

internal/validator/validator.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/enterprise-contract/ec-cli/internal/policy/source"
+)
+
+type ImageValidator interface {
+	Validate(context.Context, name.Reference) *ImageResult
+}
+
+type Options struct {
+	Policy     policy.Policy
+	PolicyURLs []source.PolicyUrl
```
zregvart, line 37: Do we need this, it is already contained or derived from `policy.Policy`?

```suggestion
```

internal/validator/validator.go
```diff
@@ -0,0 +1,70 @@
...
+	// TODO: Add support for ruleData
+	// RuleData map[string]string
+}
+
+type ImageResult struct {
+	// TODO: Remove dependency on conftest/output.Result
+	Violations []confOutput.Result
+	Warnings   []confOutput.Result
+	Successes  []confOutput.Result
+	Signatures []output.EntitySignature
```
zregvart, line 48: Most likely we'd have a `Info` struct here, that might or might not contain signatures

internal/validator/validator.go
```diff
@@ -0,0 +1,70 @@
...
+	// TODO: Add support for Attestations
+}
+
+type imageInitializer func(Options) ImageValidator
+
+var imageValidators = map[string]imageInitializer{}
+
+func RegisterImageValidator(name string, v imageInitializer) error {
+	if _, ok := imageValidators[name]; ok {
```
zregvart, line 57: Not sure we need to distinguish between validators, we could extend the `ImageValidator` interface with something like:

```
IsApplicableTo(Artifact) bool
```

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -86,6 +92,18 @@ func NewApplicationSnapshotImage(ctx context.Context, url string, p policy.Polic
 		return nil, err
 	}
 
+	// TODO: At some point, the SignatureValidator is no longer a special case, and it can
+	// be enabled or disabled via the source groups in the PolicyConfiguration.
+	signatureValidator, err := validator.NewImageValidator(
+		signature.SignatureValidatorName, validator.Options{Policy: p})
```
zregvart, line 98: Illustrates the point of having named validators, if they did not have a name we could not tie this bit of code to the specific validator, in the end what is the difference than calling a validation method instead? Goes through a map to get to the method?

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -86,6 +92,18 @@ func NewApplicationSnapshotImage(ctx context.Context, url string, p policy.Polic
...
+	if err != nil {
+		return nil, err
+	}
+	a.Validators = append(a.Validators, signatureValidator)
+
+	// TODO: Populate a.Validators from sourceGroup. Initially, assume that they all
+	// want ProvenanceV0_2Validator.
```
zregvart, line 105: We should use something like `validator.Validate(Context, Artifact)` which in turn would loop over and find all applicable validators. Bit of code here should not care about specific validators at all.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -70,6 +71,11 @@ type ApplicationSnapshotImage struct {
 	signatures   []output.EntitySignature
 	attestations []attestation.Attestation[in_toto.ProvenanceStatementSLSA02]
 	Evaluators   []evaluator.Evaluator
+	Validators   []validator.ImageValidator
+}
+
+func (a *ApplicationSnapshotImage) ImageReference() name.Reference {
```
zregvart, line 77: We should drop `ApplicationSnapshotImage`, doesn't really serve any purpose, instead we should have something like `Artifact`, and an `OCIImage` as an implementation or composition of `Artifact`

zregvart (commented):
I guess I'm missing the main point of this: we introduce a Validator interface but use somewhat directly by naming a Validator to use. For me there should not be any distinction between Validator implementations, e.g. by name or otherwise.
If we're to model a Contract it would not be tied to one specific kind of artifact, i.e. applied only to OCI images. We would have a set of rules that would need to be fulfilled by any kind of artifact, by not attempting to be that abstract, we're bound to refactor again for different kinds of artifacts.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
lcarva: I agree that this piece of code shouldn't care about which validator is being executed. The long-term idea here is that the set of validators that are included is completely configuration driven. For example, I'm picturing the policy config (contract) will be expressed as something like this:

```yaml
validators:  # This bit instead of `sources`
  - name: image-signature
    policy: ...
    data: ...
 
  # Apply policies to the config resource associated with the image. For example, to validate
  # that the image contains all the required labels an enterprise expects.
  - name: image-config
    policy: ...
    data: ...

    # This should eventually evolve into something like `in-toto-slsa-provenance` which can
    # handle any supported version of SLSA Provenance. We may also want to introduce a
    # a new validator `in-toto-slsa-provenance-v1.0` for those who care about that particular
    # version
  - name: in-toto-slsa-provenance-v0.2
    policy: ...
    data: ...

  - name: in-toto-cyclonedx-sbom
    policy: ...
    data: ...
```

Once we get to some version of that, we can loop over the validators, get each one by name (without having to care about its implementation), and simply execute them. (I can see this even leveraging a plugin system in the future if the need arises.) We're not there yet, so this is now hard-coded to add the validators that match the current state of what "validate image" does.

The reason I like this approach is that it gives policy authors a lot of flexibility while making it obvious what is actually being validated. Long-term, I don't think the ec-cli should have the logic in determining which validators to be executed because that will be difficult to get it right for multiple use cases without cluttering the policy config.

@zregvart, wdyt of this reasoning?

zregvart: I see that this could work, though I don't think it is ambitious enough.

I'd go to the other extreme, I don't want to do it in configuration I want to do it via policy rules. Given the configuration it only applies to images and certain version of SLSA provenance, I would like to be in a place where I can say my contract is:

- artifact needs to be signed
- provenance needs to comply to SLSA vX

Where the internal details of the rule is what it means that an artifact is signed and by whome and that it conforms to a SLSA level. Basically having the configuration (contract) just mention what policy rules need to apply and if needed parameters for those.

lcarva:
This POC has served its purpose. Closing it now.

PR #796, opened 2023-06-28
Pretty format snaps.go

zregvart:


lcarva (approved):
Do we need a CI step to catch this?

PR #801, opened 2023-06-29
Update Conftest to v0.43.1

zregvart:
We no longer need the workaround and can use the upstream version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/in-toto/in-toto-golang v0.9.0
 	github.com/jstemmer/go-junit-report/v2 v2.0.0
 	github.com/leanovate/gopter v0.2.9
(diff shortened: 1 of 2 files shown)


zregvart:
Seems we were impacted by [this change](https://github.com/open-policy-agent/opa/issues/2331) in OPA. [This bit of code](https://github.com/enterprise-contract/ec-policies/blob/e3ce73a8f288de2d1eccf4569c2aeee7bc7fa8d1/policy/release/lib/attestations.rego#L78) is trying to parse any task result in the pipeline, some of which will not be valid JSON...

lcarva:
What should we do here? Close it for now? Move it to draft? Rebase?

zregvart:
I'll rebase this, the feedback as to what is broken helps

zregvart:
https://github.com/enterprise-contract/ec-policies/pull/546 fixed the last issue here, this is now ready to be merged

lcarva (approved):

PR #805, opened 2023-06-30
Update OPA to 0.54.X snapshot, Cosign to 2.1.1

zregvart:
Due to controller-runtime bump, we need a patched version of application-api.

Made changes so that the Cosign error wrapping compiles, for which the tests now fail.


zregvart:
Not sure where the `source` property comes from, just wanted to show what is needed to upgrade to a newer version of Cosign.

zregvart:
Will fix the test failures next...

zregvart:
Compiles and passes the checks, good for review now.

acceptance/go.mod
```diff
@@ -230,13 +230,16 @@ require (
...
 	k8s.io/utils v0.0.0-20230406110748-d93618cff8a2 // indirect
 	knative.dev/pkg v0.0.0-20230412013349-d3d7625d8e1e // indirect
-	sigs.k8s.io/controller-runtime v0.14.5 // indirect
+	sigs.k8s.io/controller-runtime v0.15.0 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+// We need to include the controller-runtime update
+replace github.com/redhat-appstudio/application-api => github.com/zregvart/application-api v0.0.0-20230630104611-185d03306f1d
```
lcarva, line 245: Can we link to the PR where this change will be incorporated?

cmd/validate/image_test.go
```diff
@@ -261,6 +261,7 @@ func Test_ValidateImageCommand(t *testing.T) {
 		  {
 			"name": "Unnamed",
 			"containerImage": "registry/image:tag",
+			"source": {},
```
lcarva, line 264: Let's not leak this out to the user's output unless it's actually something we want.

acceptance/go.mod
zregvart: Oh no that should have been removed, probably a bad conflict resolution in rebase

cmd/validate/image_test.go
lcarva: Wait, this includes the git source information which is indeed useful. It's empty in our tests because we don't provide any data to it. Presumably, this will be filled in if the input contains it. +1 to keep it as is.

We may also want to make sure this information makes its way to the input for the policy rules. (As another story, of course).

internal/output/output.go
```diff
@@ -338,18 +338,11 @@ func wrapCosignErrorMessage(err error, checkType string, p policy.Policy) string
...
-		case "signature":
-			noMatchingErr = cosign.ErrNoMatchingSignaturesType
-			msg = missingSignatureMessage
-		case "attestation":
-			noMatchingErr = cosign.ErrNoMatchingAttestationsType
-			msg = missingAttestationMessage
-		}
-		if vErr, ok := err.(*cosign.VerificationError); ok && vErr.ErrorType() == noMatchingErr && msg != "" {
-			return msg
+		switch err.(type) {
```
lcarva, line 341: The tests are passing so I'm assuming this is working as expected.

lcarva (approved):

PR #807, opened 2023-06-30
Move certificate logic to its own module

lcarva:
The logic is isolated enough and there is nothing here that explicitly ties to an Application Snapshot Image. By doing this, we can leverage the certificate related functionality more easily in other modules, e.g. `attestation`. This paves the way to exposing attestation certificates to policy rules.


internal/certificate/certificate.go
```diff
@@ -14,20 +14,25 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-package application_snapshot_image
+package certificate
```
zregvart, line 17: Perhaps?
```suggestion
package signature
```

lcarva: Seems reasonable enough. Updated package name

zregvart (approved):

PR #810, opened 2023-07-05
Move EntitySignature from output to signature

zregvart:
Seems more fitting to place it here.

PR #811, opened 2023-07-05
Remove generics for attestation statement

zregvart:
Using generics for this was to invasive on the rest of the code and didn't offer much flexibility.

Includes #810

### DIFF
--- a/internal/attestation/attestation.go
+++ b/internal/attestation/attestation.go
@@ -32,9 +32,9 @@ var (
 // Attestation holds the raw attestation data, usually fetched from the
 // signature envelope's payload; statement of a particular type and any
 // signing information.
-type Attestation[T any] interface {
+type Attestation interface {
 	Data() []byte
-	Statement() T
+	Statement() any
 	Signatures() []signature.EntitySignature
 }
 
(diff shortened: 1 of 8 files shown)


internal/attestation/attestation.go
```diff
@@ -28,11 +30,15 @@ var (
 	AT004 = e.NewError("AT004", "Unsupported attestation predicate type", e.ErrorExitStatus)
 )
 
+type Statement interface {
+	in_toto.ProvenanceStatementSLSA02 | any
```
lcarva, line 34: This is only used as the return type of `Attestation.Statement()`. That method is only called from `ApplicationSnapshotImage.WriteInputFile` when building up the large input JSON. I think `any` is probably the right type for that. It just has to be something that can be encoded in JSON.

zregvart: I need to add a method on the `Statement`, the `| any` is here only so that the tests can be a bit less verbose. In fact I was thinking of removing it and keeping a type safe list of supported attestation formats. We want to be careful of the data we pass to the rules and eventually output.

lcarva: IMO the payload of an attestation that has been properly signed is safe to forward to the policies. I don't think we should care too much about the type unless we are doing additional validation in ec-cli.

zregvart: We have `Data` method on `Attestation` that returns the raw bytes, as found on the layer, the `Statement` method should return something that is usable (type safe). I think I’ll remove the `any` from it.

zregvart: Hmm, doesn't compile without `| any`, I'm going to remove the `Statement` type and return `any` from the `Statement` method...

PR #813, opened 2023-07-05
Expose sigs of attestations in report

lcarva:
This commit adds the `components[].attestations` attribute to the yaml/json output of the `validate image` command. Its value is an array of objects containing information about each attestation, including the signature used. For example:

```yaml
- predicateBuildType: https://github.com/slsa-framework/slsa-github-generator/container@v1
  predicateType: https://slsa.dev/provenance/v0.2
  signatures:
    - certificate: -----BEGIN CERTIFICATE-----...
      chain:
        - -----BEGIN CERTIFICATE-----...
        - -----BEGIN CERTIFICATE-----...
      keyid: 2e72b04f189b4582e6519674626c2dcc20cf2bb4
      metadata:
        Fulcio Build Config Digest: 848edc452ccbc6d42ec56c2807eef2f49e754c5e
        # ...
      sig: MEYCIQCvXFb52u0Ida8AGuZGSNaAKPE6sII44XJoWhp1YjK7rAIhAKQKdXVxDlzVBi34rSGfsFGmbwa/7aZZNAthlys0b9jr
  type: https://in-toto.io/Statement/v0.1
```

https://issues.redhat.com/browse/HACBS-2296


lcarva:
This PR is marked as a draft because the acceptance tests are failing. 

I was able to verify the code is working properly by validating existing images. [Signature](https://pkg.go.dev/github.com/sigstore/cosign/v2/pkg/oci#Signature)`.Base64Signature` returns the signature ID-ish value we want, e.g. `MEUCIQD1QDuipAR2gXOK6hTklLM0wAvY2Ydfr9XatxhVIgy/HwIgDChdM/w5SvYw0shiZCqmXYxXHNq4kWegpJrRtkeJ4KE=`

But in the acceptance tests, it returns the actual attestation encoded in base64. Trying to figure out what would even cause that. From what I can tell, and when comparing with cosign itself, the acceptance tests appear to be creating the attestation correctly.

features/__snapshots__/task_validate_image.snap
```diff
@@ -95,13 +113,39 @@ success: true
...
+    signatures:
+    - keyid: SHA256:RHajkr+wMEtGfT2CRFrQEhg/8MY2bDLXVg3F8IuI5nE
+      sig: MEUCIGS176zN5aoorLQMukjoCkHm7ocu7UhnKXLhzEdsgp4BAiEAviub3Lf4thLmSTU6ZqnEjw02kkrb9LKBBa1t8hVgAM4=
+    type: https://in-toto.io/Statement/v0.1
+  containerImage: quay.io/hacbs-contract-demo/golden-container@sha256:e76a4ae9dd8a52a0d191fd34ca133af5b4f2609536d32200a4a40a09fdc93a0d
   name: ""
   signatures:
   - keyid: ""
-    sig: ${IMAGE_SIGNATURE_acceptance/strict-with-warnings}
+    sig: MEUCIFPod1d9HhGt+TEQPG4j+LINjkifCFFOFrE4jbkvexGGAiEAqSp3ROZUsIOwWro6Tv+lRiR7sdMR0U6Crs1ISuQhHtA=
```
lcarva, line 148: Hmm need to look into why this change happened.

internal/output/output.go
```diff
@@ -73,6 +73,13 @@ func (v VerificationStatus) addToSuccesses(successes []output.Result) []output.R
 	return result
 }
 
+type Attestation struct {
```
zregvart, line 76: Could we use the existing type `internal/attestation.Attestation` here?

internal/attestation/attestation.go
```diff
@@ -35,7 +36,7 @@ var (
 type Attestation[T any] interface {
 	Data() []byte
 	Statement() T
-	Signatures() []signature.EntitySignature
+	Output() output.Attestation
```
zregvart, line 39: This looks odd, attestation containing attestation?

zregvart (commented):
Not sure about the snapshot changes, did you run a selection of tests?

lcarva:
> Not sure about the snapshot changes, did you run a selection of tests?

Nope. This happens whenever I run `UPDATE_SNAPS=true make acceptance` :disappointed:

internal/output/output.go
lcarva: No, for a couple of reasons. 

First, `output.Attestation` is a struct that **represents** an attestation. It is not the attestation itself. The `output` package seems like the right place for this. This could've been called `AttestationOutput` but given that it resides in the package `output`, that seemed redundant. 

Second, `attestation.Attestation` is an interface. It is not possible to add attributes to an interface in go afaik. Instead, I added a method to the interface to produce the struct that represents the attestation. 

I don't really see a problem with this approach. The separation of concerns is clearly defined to me, but I'm always open to suggestions.

internal/attestation/attestation.go
lcarva: That's not what this means. It's a method, not an attribute. An `attestation.Attestation` knows how to represent itself as an `output.Attestation`. This is similar to how you'd use a `String` method to get a string representation of a struct.

features/__snapshots__/task_validate_image.snap
lcarva: Fixed this. At some point, I must have deleted this file, and let it be recreated. The problem is that the test results may be placed in a different order when the file is created from scratch. If the file already exists, then the existing order is retained.

lcarva:
> Not sure about the snapshot changes, did you run a selection of tests?

@zregvart, addressed this. See explanation [here](https://github.com/enterprise-contract/ec-cli/pull/813#discussion_r1254588299).

simonbaird (approved):

PR #821, opened 2023-07-10
Fetch and expose config of image and its parent

lcarva:
A new package, `fetcher`, is added. This contain two classes. `ImageConfigFetcher` retrieves the image config[1] of a given OCI image. `BaseImageConfigFetcher` retrieves the image config of the parent/base image of a given OCI image.

The image config contains all sort of goodies, including all the image labels.

The image config is exposed in the policy input at `input.image.config` and the image config of the parent image at `input.image.parent.config` (was `input.image.parent_config` at an earlier version of this PR).

[1] https://github.com/opencontainers/image-spec/blob/main/config.md

https://issues.redhat.com/browse/HACBS-2372


lcarva:
There are two issues with the acceptance tests that I'm still trying to resolve:

1. For `validate_image`, I need to regenerate the image used so it has the expected annotation. This is an optional check, but it does raise a warning when not found. This may be something we do as a follow up to this PR.
2. For `task_validate_image`, I'm seeing some "connection refused" errors when accessing the registry. I'm out of ideas of what could be causing this. This is something that needs to be fixed before merging this PR. Any help here would be greatly appreciated!

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -63,11 +64,13 @@ var attestationSchemas = map[string]jsonschema.Schema{
...
-	reference    name.Reference
-	checkOpts    cosign.CheckOpts
-	signatures   []signature.EntitySignature
-	attestations []attestation.Attestation
-	Evaluators   []evaluator.Evaluator
+	reference        name.Reference
+	checkOpts        cosign.CheckOpts
+	signatures       []signature.EntitySignature
+	configJSON       json.RawMessage
+	parentConfigJSON json.RawMessage
```
zregvart, line 71: We never care about parent's parent? Could this be:
```suggestion
	configJSON       map[name.Reference]json.RawMessage
	path             map[name.Reference]name.Reference
```

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -167,6 +170,36 @@ func (a *ApplicationSnapshotImage) SetImageURL(url string) error {
...
+		remote.WithContext(ctx),
+		remote.WithAuthFromKeychain(authn.DefaultKeychain),
+	}
+	f := image_config.Fetcher{}
+	var err error
+	a.configJSON, err = f.Fetch(ctx, a.reference, opts...)
+	if err != nil {
+		return err
+	}
+	return nil
```
zregvart, line 185:
```suggestion
	return err
```

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -167,6 +170,36 @@ func (a *ApplicationSnapshotImage) SetImageURL(url string) error {
...
+		remote.WithContext(ctx),
+		remote.WithAuthFromKeychain(authn.DefaultKeychain),
+	}
+	f := image_config.ParentFetcher{}
+	var err error
+	a.parentConfigJSON, err = f.Fetch(ctx, a.reference, opts...)
+	if err != nil {
+		return err
+	}
+	return nil
```
zregvart, line 200:
```suggestion
	return err
```

internal/fetcher/image_config/client.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+type key string
+
+const clientKey key = "ec.fetcher.config.client"
+
+type Client interface {
```
zregvart, line 31: We have so many clients, could we name this to be more specific, does it need to be public?
```suggestion
const imageFetcherKey key = "ec.fetcher.config.client"

type imageFetcher interface {
```

Or even:

```suggestion
const imageFetcherKey key = "ec.fetcher.config.client"

type imageFetcher func(name.Reference, ...remote.Option) (v1.Image, error)
```

internal/fetcher/image_config/image_config.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+// Fetcher retrieves the config for an image from its OCI registry.
+type Fetcher struct{}
```
zregvart, line 29: Doesn't have state or implement an interface, do we need the struct?

zregvart (commented):
Not sure why the checks need to be builtin, or if we care about parent's parents? Should we check that a/the layer(s) correspond(s) to the base image?

acceptance/image/image.go
```diff
@@ -361,7 +393,25 @@ func createAndPushImage(ctx context.Context, imageName string) (context.Context,
...
+	descriptor, err := remote.Head(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	if err != nil {
+		return "", err
+	}
+	digest := descriptor.Digest.String()
+	if digest == "" {
+		return "", fmt.Errorf("digest for image %q is empty", url)
+	}
+
+	return fmt.Sprintf("%s@%s", ref.Name(), digest), nil
```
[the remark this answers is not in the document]
lcarva: No. `digest` is a string with the digest.

internal/fetcher/image_config/client.go
```diff
@@ -0,0 +1,53 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package image_config
```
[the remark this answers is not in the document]
lcarva: Yeah. I had it named `config` at first, but that seem a bit ambiguous. Maybe `image`? Suggestions welcome.

internal/utils/testing_image.go
```diff
@@ -0,0 +1,33 @@
...
+// tag which is not included when building the ec binary.
+package utils
+
+import (
+	"crypto/sha256"
+	"fmt"
+)
+
+// WithDigest appends a digest to the given image reference.
+func WithDigest(ref string) string {
```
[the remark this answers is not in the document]
lcarva: I rather like the ability to just use this function in various places in the code base without having to hard code some really long digest.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
lcarva: I think that makes it much harder to use the input from the policy.

> We never care about parent's parent?

Not that I know of. I can change `input.image.parent_config` to `input.image.parent.config`. That should allow us to do something like this in the future if needed: `input.image.parent.parent.parent.parent.config`

lcarva: This actually works out quite nice because I can now easily add `input.image.parent.ref` as well which could be useful to some policy rules.

internal/fetcher/image_config/client.go
lcarva: It doesn't have to be public actually. It is odd though because it is part of functions that are exported. :shrug: 

> We have so many clients

I don't see the problem with this. They are scoped to different packages. I find it beneficial to have a nice repeatable pattern.

acceptance/image/image.go
```diff
@@ -325,34 +326,65 @@ func createAndPushAttestationWithPatches(ctx context.Context, imageName, keyName
...
+		return ctx, err
+	}
+
+	return ctx, nil
+}
+
 // createAndPushImage creates a small 4K random image with 2 layers and pushes it to
 // the stub image registry
-func createAndPushImage(ctx context.Context, imageName string) (context.Context, error) {
+func createAndPushPlainImage(ctx context.Context, imageName string, patch func(v1.Image) v1.Image) (context.Context, string, error) {
```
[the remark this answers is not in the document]
lcarva: Done.

acceptance/image/image.go
```diff
@@ -361,7 +393,25 @@ func createAndPushImage(ctx context.Context, imageName string) (context.Context,
...
+	return ctx, ref.String(), nil
+}
+
+// resolveDigest returns an image reference that is guaranteed to have a digest.
+func resolveDigest(url string) (string, error) {
+	ref, err := name.ParseReference(url)
+	if err != nil {
+		return "", err
+	}
+	descriptor, err := remote.Head(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
```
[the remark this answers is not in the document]
lcarva: Added, but not sure if that code path will ever be executed by the acceptance tests.

features/__snapshots__/task_validate_image.snap
```diff
@@ -69,13 +69,26 @@ components:
...
       code: builtin.image.signature_check
       title: Image signature check passed
     msg: Pass
   warnings:
+  - metadata:
+      code: builtin.parent_image.config_check
+      title: Parent image config is accessible
+    msg: 'Parent image config is not accessible: unable to fetch parent image: Get
+      "https://localhost:37957/v2/": dial tcp [::1]:37957: connect: connection refused;
+      Get "http://localhost:37957/v2/": dial tcp [::1]:37957: connect: connection
```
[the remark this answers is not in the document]
lcarva: Unclear what was happening here. With `-persist`, I was able to query the registry just fine from the terminal. Not sure if there was a timing issue during the tests. For now, this is no longer an issue because I removed the newly added `builtin` results.

internal/output/output.go
```diff
@@ -73,6 +73,21 @@ func (v VerificationStatus) addToSuccesses(successes []output.Result) []output.R
...
+// addToWarnings appends the failure result to the warnings slice. Use this for
+// checks that are optional.
+func (v VerificationStatus) addToWarnings(warnings []output.Result) []output.Result {
+	if v.Passed {
+		return warnings
+	}
+
+	result := warnings
+	if v.Result != nil {
+		result = append(warnings, *v.Result)
```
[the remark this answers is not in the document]
lcarva: Removed this function altogether in newest update.

internal/fetcher/image_config/image_config.go
lcarva: We do not. At some point, this patch included having a list of fetchers that the caller would just iterate through. It made sense then. Removed for now. We'll likely come up with some interface at some point, but it is premature to do that right now.

lcarva:
@zregvart, thanks for the initial review! This is ready for another round.

features/__snapshots__/validate_image.snap
```diff
@@ -416,3 +416,11 @@ Error: success criteria not met
...
..."}},{"msg":"Pass","metadata":{"code":"image_config.image_distinct_title_set","description":"Check if the image has a different value than its parent image for the org.opencontainers.image.title label.","title":"Image Distinct Title Label"}},{"msg":"Pass","metadata":{"code":"image_config.image_title_set","description":"Check if the image has the org.opencontainers.image.title label set.","title":"Image Title Label"}},{"msg":"Pass","metadata":{"code":"image_config.parent_image_title_set","description":"Check if the parent image has the org.opencontainers.image.title label set.","title":"Parent Image Title Label"}}],"success":true,"signatures":[${IMAGE_SIGNATURES_JSON_acceptance/image-config}],"attestations":[{"type":"https://in-toto.io/Statement/v0.1","predicateType":"https://slsa.dev/provenance/v0.2","predicateBuildType":"https://tekton.dev/attestations/chains/pipelinerun@v2","signatures":[${ATTESTATION_SIGNATURES_JSON_acceptance/image-config}]}]}],"key":${known_PUBLIC_KEY_JSON},"policy":{"sources":[{"policy":["git::https://${GITHOST}/git/image-config-policy.git"]}],"rekorUrl":"${REKOR}","publicKey":"${known_PUBLIC_KEY}"},"ec-version":"${EC_VERSION}","effective-time":"${TIMESTAMP}"}
```
simonbaird, line 421: Can we use that new trick to make this indented and formatted? Not sure why it's not doing that already, the code seems to be towards the end of MatchSnapshot in acceptance/snaps/snaps.go .

simonbaird (approved):

features/__snapshots__/validate_image.snap
zregvart: The JSON format is broken with the variable substitutions; one example search for `${ATTESTATION_SIGNATURES_JSON`. See [HACBS-2337](https://issues.redhat.com/browse/HACBS-2337)

internal/fetcher/image_config/client.go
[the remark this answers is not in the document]
zregvart: Perhaps `oci/config`? We can refactor later...

zregvart (approved):
Let's roll with this and refactor as needed in the future...

features/__snapshots__/validate_image.snap
simonbaird: Oooh, thanks. Now I understand what that [HACBS-2337](https://issues.redhat.com/browse/HACBS-2337) is about..

PR #822, opened 2023-07-11
Add descriptions to builtin rules

mbestavros:
Ref: https://issues.redhat.com/browse/HACBS-2373

The team noticed recently that the builtin rules don't have descriptions. This PR adds them.


internal/output/output.go
```diff
@@ -99,8 +99,9 @@ type Output struct {
 // SetImageAccessibleCheck sets the passed and result.message fields of the ImageAccessibleCheck to the given values.
 func (o *Output) SetImageAccessibleCheckFromError(err error) {
 	metadata := map[string]interface{}{
-		"code":  "builtin.image.accessible",
-		"title": "Image URL is accessible",
+		"code":        "builtin.image.accessible",
+		"title":       "Image URL is accessible",
+		"description": "The image URL is available and accessible on a public registry.",
```
lcarva, line 104: It doesn't have to be a public registry/repository. The underlying system just has to have access to it. For example, if you have a `~/.docker/config.json`, ec-cli will use that.

internal/output/output.go
```diff
@@ -122,8 +123,9 @@ func (o *Output) SetImageAccessibleCheckFromError(err error) {
 // SetImageSignatureCheck sets the passed and result.message fields of the ImageSignatureCheck to the given values.
 func (o *Output) SetImageSignatureCheckFromError(err error) {
 	metadata := map[string]interface{}{
-		"code":  "builtin.image.signature_check",
-		"title": "Image signature check passed",
+		"code":        "builtin.image.signature_check",
+		"title":       "Image signature check passed",
+		"description": "The image signature matches the provided public key.",
```
lcarva, line 128: It would be nice to phrase this in a way that it also makes sense for keyless signatures where a public key is not actually provided.

zregvart:
I'm expecting acceptance tests to pass once rebased

internal/output/output.go
mbestavros: @lcarva What do you think of "The image signature matches available signing materials"?

lcarva: Sounds good!

internal/output/output.go
mbestavros: @lcarva I've updated it to remove "...on a public registry" - open to more detailed suggestions, though.

lcarva (approved):

zregvart (approved):

PR #824, opened 2023-07-12
Update snapshot with redhat collection

zregvart:


zregvart:
Merging this to get the dependabot updates rebased

PR #827, opened 2023-07-13
Fix broken JSON in test snapshots

joejstuart:
This replaces the json object in the signature with the strings for `keyid` and `sig`.
Resolves: https://issues.redhat.com/browse/HACBS-2337


acceptance/cli/cli.go
```diff
@@ -230,26 +230,26 @@ func setupKeys(ctx context.Context, vars map[string]string, environment []string
 }
 
 func setupSigs(ctx context.Context, vars map[string]string, environment []string) ([]string, map[string]string, error) {
-	type valFunc func(context.Context) (map[string]string, error)
+	type valFunc func(string, context.Context) (map[string]string, error)
```
zregvart, line 233: (nitpick) This would be more inline with golang style...

```suggestion
	type valFunc func(context.Context, string) (map[string]string, error)
```

features/__snapshots__/validate_image.snap
```diff
@@ -1,6 +1,6 @@
...
+{"success":true,"components":[{"name":"Unnamed","containerImage":"${REGISTRY}/acceptance/ec-happy-day@${REGISTRY_acceptance/ec-happy-day:latest_HASH}","successes":[{"msg":"Pass","metadata":{"code":"builtin.attestation.signature_check"}},{"msg":"Pass","metadata":{"code":"builtin.attestation.syntax_check"}},{"msg":"Pass","metadata":{"code":"builtin.image.signature_check"}},{"msg":"Pass","metadata":{"code":"main.acceptor"}}],"success":true,"signatures":[{"keyid": "${IMAGE_SIGNATURE_KEY_ID_acceptance/ec-happy-day}","sig": "${IMAGE_SIGNATURE_acceptance/ec-happy-day}"}],"attestations":[{"type":"https://in-toto.io/Statement/v0.1","predicateType":"https://slsa.dev/provenance/v0.2","predicateBuildType":"https://tekton.dev/attestations/chains/pipelinerun@v2","signatures":[${ATTESTATION_SIGNATURES_JSON_acceptance/ec-happy-day}]}]}],"key":${known_PUBLIC_KEY_JSON},"policy":{"sources":[{"policy":["git::https://${GITHOST}/git/happy-day-policy.git"]}],"rekorUrl":"${REKOR}","publicKey":"${known_PUBLIC_KEY}"},"ec-version":"${EC_VERSION}","effective-time":"${TIMESTAMP}"}
```
zregvart, line 3: I would not expect to see `ATTESTATION_SIGNATURES_JSON_acceptance` here

acceptance/image/image.go
```diff
@@ -633,17 +633,14 @@ func JSONAttestationSignaturesFrom(ctx context.Context) (map[string]string, erro
 
 	signatures := map[string]string{}
 	for name, signature := range state.AttestationSignatures {
-		json, err := json.Marshal(signature)
-		if err != nil {
-			return nil, err
-		}
-		signatures[name] = string(json)
+		signatures[fmt.Sprintf("%s_KEY_ID_%s", prefix, name)] = signature.KeyID
+		signatures[fmt.Sprintf("%s_%s", prefix, name)] = signature.Signature
```
zregvart, line 637: Seems that `ATTESTATION_SIGNATURE_...` can be `""`, then in snaps.go we replace everything with `${ATTESTATION_SIGNATURE_...}`, so let's guard here

```suggestion
		if signature.KeyID != "" {
			signatures[fmt.Sprintf("%s_KEY_ID_%s", prefix, name)] = signature.KeyID
		}
		if signature.Signature != "" {
			signatures[fmt.Sprintf("%s_%s", prefix, name)] = signature.Signature
		}
```

acceptance/image/image.go
```diff
@@ -706,17 +703,14 @@ func JSONImageSignaturesFrom(ctx context.Context) (map[string]string, error) {
 
 	ret := map[string]string{}
 	for name, signature := range state.ImageSignatures {
-		json, err := json.Marshal(signature)
-		if err != nil {
-			return nil, err
-		}
-		ret[name] = string(json)
+		ret[fmt.Sprintf("%s_KEY_ID_%s", prefix, name)] = signature.KeyID
+		ret[fmt.Sprintf("%s_%s", prefix, name)] = signature.Signature
```
zregvart, line 707: Same as above, let's guard here for empty strings:
```suggestion
		if signature.KeyID != "" {
			ret[fmt.Sprintf("%s_KEY_ID_%s", prefix, name)] = signature.KeyID
		}
		if signature.Signature != "" {
			ret[fmt.Sprintf("%s_%s", prefix, name)] = signature.Signature
		}
```

zregvart (commented):
I think this is the correct solution, needs a bit of tweaking...

joejstuart:
> I think this is the correct solution, needs a bit of tweaking...

This was the other solution I thought of. I thought this one (827) was less of a change though. https://github.com/enterprise-contract/ec-cli/pull/830/files

zregvart (commented):
Oh, I think we need to do the same for `*_PUBLIC_KEY_JSON`

simonbaird:
I was expecting to see a bunch of nicely intented json in the snapshot file. Will that happen eventually?

joejstuart:
> I was expecting to see a bunch of nicely indented json in the snapshot file once all the non-valid JSON is removed. Will that happen eventually?

Absolutely, will pick this back up tomorrow.

acceptance/snaps/snaps.go
```diff
@@ -95,7 +95,15 @@ func MatchSnapshot(ctx context.Context, qualifier, text string, vars map[string]
 	errs := capture(ctx, qualifier)
 
 	for k, v := range vars {
-		text = strings.ReplaceAll(text, v, "${"+k+"}")
+		// Todo: remove ASAP. Just a hack until a solution is figured out
+		if k == "IMAGE_SIGNATURES_XML_acceptance/image" {
```
simonbaird, line 99: As mentioned on slack, workaround/hack is okay to merge, but I think there should be a note to help explain what it does and why it's needed.

PR #829, opened 2023-07-14
Refactor the attestation model

zregvart:
Given the general nature of attestations we were to tightly coupled to the model of build attestations. With this we can evolve to support different types of attestations. We were also coupled to the output needed for the reports in our model, which doesn't allow us to evolve the model without modifying the output.

I think the `attestation.Attestation` model can be simplified even more, something that I'll look into as well. Most likely the metadata and the typed statement are not needed and could be kept internal to the attestation implementation.

Right now we also equate the statement with attestation, further refinement could introduce a model for the attestation and statement separately.


lcarva:
Looks like the right direction.

zregvart:
Refactored a bit more and added a bit of test coverage, this should be okay for review now.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -298,48 +298,80 @@ func (a *ApplicationSnapshotImage) Attestations() []attestation.Attestation {
...
+		return nil, err
+	}
+
+	extra, err := json.Marshal(a.Extra)
+	if err != nil {
+		return nil, err
+	}
+
+	buffy.Write(extra)
+	buffy.WriteByte('}')
```
lcarva, line 337: I would expect `{` earlier in this function. Is `a.RawMessage.MarshalJSON()` returning a partial JSON?

internal/attestation/attestation.go
```diff
@@ -34,13 +33,8 @@ var (
...
-	Output() output.Attestation
-}
-
-// extra holds the signatures associated with an attestation. It is meant
-// to facilitate embedding signatures in the result of Attestation.Statement().
-type extra struct {
-	Signatures []signature.EntitySignature `json:"signatures"`
+	Type() string
+	PredicateType() string
+	Statement() []byte
```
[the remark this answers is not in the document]
zregvart: Doesn't need to be JSON nor YAML, an attestation could be in XML format for all we know...

internal/attestation/slsa_provenance_02.go
```diff
@@ -128,24 +120,39 @@ func createEntitySignatures(sig oci.Signature, payload cosign.AttestationPayload
...
+	return a.data
 }
 
-func (a slsaProvenance) Output() output.Attestation {
-	return output.Attestation{
+func (a slsaProvenance) Signatures() []signature.EntitySignature {
+	return a.signatures
+}
+
+func (a slsaProvenance) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
zregvart: I would not force it, this influences the output only...

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
zregvart: [`a.RawMessage.MarshalJSON()`](https://github.com/enterprise-contract/ec-cli/pull/829/files/ba2b2404a6d64e3d7856e379319c8d3185be9bd7#diff-c89fbe58faac75c0a251f1b8af58b7789adc10532e0814ad59ef0adc47d91445R316) is returning JSON from which the last `}` [is stripped](https://github.com/enterprise-contract/ec-cli/pull/829/files/ba2b2404a6d64e3d7856e379319c8d3185be9bd7#diff-c89fbe58faac75c0a251f1b8af58b7789adc10532e0814ad59ef0adc47d91445R321).

lcarva: Aha! Missed that.

lcarva (approved):

PR #831, opened 2023-07-14
read snapshot from file

joejstuart:
Update verify-enterprise-contract task to read the snapshot from a file in a workspace

resolves: https://issues.redhat.com/browse/HACBS-2329

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -99,6 +99,11 @@ spec:
       description: Run policy checks with the provided time.
       default: "now"
 
+  workspaces:
+    - name: data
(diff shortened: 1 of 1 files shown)


joejstuart:
Marking this as draft since there's more work to do.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -137,8 +142,8 @@ spec:
         - validate
         - image
         - "--verbose"
-        - "--json-input"
-        - "$(params.IMAGES)"
+        - "--file-input"
```
lcarva, line 145: An alternative is to introduce a new flag, e.g. `--images`, that can take a file path or an inline snapshot, kind of like the `--policy` flag. That will give us a lot of flexibility, and allow the task to remain backwards compatible.

PR #834, opened 2023-07-17
Add attestation output to ec validate image

zregvart:
Now any attestations can be outputted when running `ec validate image`
using the `--output attestation` format. The output is presented as a
multipart MIME document, allowing more than one attestation to be
outputted.

Using the existing `--output` functionality allows placing the output to
a file.

Ref. https://issues.redhat.com/browse/HACBS-2133

Includes #829 so review that first...

### DIFF
--- a/features/validate_image.feature
+++ b/features/validate_image.feature
@@ -715,3 +715,27 @@ Feature: evaluate enterprise contract
     When ec command is run with "validate image --image ${REGISTRY}/acceptance/image-config --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --rekor-url ${REKOR} --show-successes --info"
     Then the exit status should be 0
     Then the output should match the snapshot
+
+  Scenario: Output attestations
+    Given a key pair named "known"
+      And an image named "acceptance/image"
+      And a valid image signature of "acceptance/image" image signed by the "known" key
+      And a valid attestation of "acceptance/image" signed by the "known" key
+      And a git repository named "my-policy" with
+      | happy.rego  | examples/happy_day.rego   |
+      And policy configuration named "ec-policy" with specification
+    """
+    {
+      "sources": [
+        {
+          "policy": [
+            "git::https://${GITHOST}/git/my-policy.git"
+          ]
+        }
+      ]
+    }
+    """
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/image --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --output=json --output attestation=${TMPDIR}/attestation.mime --strict"
+    Then the exit status should be 0
+    And the output should match the snapshot
+    And the "${TMPDIR}/attestation.mime" file should match the snapshot
--- a/features/validate_image.feature
+++ b/features/validate_image.feature
@@ -735,7 +735,7 @@ Feature: evaluate enterprise contract
       ]
     }
     """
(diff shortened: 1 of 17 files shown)


simonbaird:
I'm curious about the decision to use multipart mime.

I could imaging outputting input.json in its entirerly, which would include the attestations, plus everything else, and might be potentially useful for troubleshooting, debugging, auditing etc.

internal/applicationsnapshot/attestation.go
```diff
@@ -0,0 +1,54 @@
...
+func (r *Report) renderAttestations() ([]byte, error) {
+	buffy := bytes.Buffer{}
+
+	w := multipart.NewWriter(&buffy)
+	adapt(w)
+	defer w.Close()
+
+	for _, c := range r.Components {
+		for i, a := range c.Attestations {
+			pw, err := w.CreatePart(textproto.MIMEHeader{
```
lcarva, line 39: This output format is very surprising for a CLI tool.

zregvart:
a31e00af8502b40f168867a7c195feb9cc11854d illustrates the difference between MIME multipart and JSONL

features/validate_image.feature
```diff
@@ -715,3 +715,27 @@ Feature: evaluate enterprise contract
...
+            "git::https://${GITHOST}/git/my-policy.git"
+          ]
+        }
+      ]
+    }
+    """
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/image --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --output=json --output attestation=${TMPDIR}/attestation.mime --strict"
+    Then the exit status should be 0
+    And the output should match the snapshot
+    And the "${TMPDIR}/attestation.mime" file should match the snapshot
```
lcarva, line 741: Maybe remove the `.mime` suffix to avoid confusion?

internal/applicationsnapshot/attestation.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"bytes"
+)
+
+func (r *Report) renderAttestations() ([]byte, error) {
+	buffy := bytes.Buffer{}
+
+	for i, c := range r.Components {
+		if i > 0 && len(c.Attestations) > 0 {
+			if err := buffy.WriteByte('\n'); err != nil {
```
lcarva, line 28: Could consider using [bytes.Join](https://pkg.go.dev/bytes#Join) to avoid the two conditionals.

internal/attestation/attestation.go
```diff
@@ -35,6 +35,7 @@ var (
 type Attestation interface {
 	Type() string
 	PredicateType() string
+	ContentType() string
```
lcarva, line 38: Is this still needed? Seems to not be used.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -161,6 +161,8 @@ spec:
         - "appstudio=$(results.TEST_OUTPUT.path)"
         - "--output"
         - "data=$(params.HOMEDIR)/data.yaml"
+        - "--output"
+        - "attestation=$(params.HOMEDIR)/attestations.mime"
```
lcarva, line 165: Use `.jsonl` extension?

lcarva (approved):

internal/attestation/attestation.go
zregvart: We don't output it, so it can be removed, it goes unstated that attestations can be other than JSON so...

PR #835, opened 2023-07-17
adding parameter snapshot-source to validate image

joejstuart:
This new param supports the snapshot spec as a JSON string or a file path to a valid spec.

Resolves: https://issues.redhat.com/browse/HACBS-2329

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -48,10 +48,10 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		certificateOIDCIssuer       string
 		certificateOIDCIssuerRegExp string
 		effectiveTime               string
-		filePath                    string
+		filePath                    string // Deprecated: images replaced this
 		imageRef                    string
 		info                        bool
-		input                       string
+		input                       string // Deprecated: images replaced this
 		output                      []string
 		outputFile                  string
 		policy                      policy.Policy
@@ -102,11 +102,11 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			Validate multiple images from an ApplicationSnapshot Spec file:
 
-			  ec validate image --file-path my-app.yaml
+			  ec validate image --images my-app.yaml
 
 			Validate attestation of images from an inline ApplicationSnapshot Spec:
 
-			  ec validate image --json-input '{"components":[{"containerImage":"<image url>"}]}'
+			  ec validate image --images '{"components":[{"containerImage":"<image url>"}]}'
 
 			Use a different public key than the one from the EnterpriseContractPolicy resource:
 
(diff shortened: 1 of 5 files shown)


cmd/validate/image.go
```diff
@@ -374,6 +376,9 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 	cmd.Flags().StringVarP(&data.input, "json-input", "j", data.input,
 		"JSON representation of an ApplicationSnapshot Spec")
 
+	cmd.Flags().StringVarP(&data.snapshotSource, "snapshot-source", "a", data.snapshotSource,
```
lcarva, line 379: Can we mark `file-path` and `json-input` as deprecated?

lcarva, line 379: What if this was called `images`? I'm thinking how `ApplicationSnapshot` means absolutely nothing to anyone outside RHTAP. We may want to not mention that here at all. But that's a different PR. I'm just trying to make the case for not adding a new parameter named after `ApplicationSnapshot`. (Descriptions can be changed later on with ease, parameter names not so much.)

NOTE: We don't actually have to have a long and short version of the parameter name. Allowing just the long version is perfectly ok.

cmd/validate/image.go
```diff
@@ -177,10 +178,11 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
...
 			if s, err := applicationsnapshot.DetermineInputSpec(ctx, applicationsnapshot.Input{
-				File:     data.filePath,
-				JSON:     data.input,
-				Image:    data.imageRef,
-				Snapshot: data.snapshot,
+				File:           data.filePath,
+				JSON:           data.input,
+				Image:          data.imageRef,
+				Snapshot:       data.snapshot,
+				SnapshotSource: data.snapshotSource,
```
lcarva, line 185: There's a conditional required flag check at the end of this file that I think needs to be updated to take this into account, 

```
if len(data.input) > 0 || len(data.filePath) > 0 {
		if err := cmd.MarkFlagRequired("image"); err != nil {
			panic(err)
		}
	}
```

internal/applicationsnapshot/input.go
```diff
@@ -81,37 +82,46 @@ func DetermineInputSpec(ctx context.Context, input Input) (*app.SnapshotSpec, er
 	var snapshot snapshot
 	provided := false
 
+	if input.SnapshotSource != "" {
+		var content []byte
+		var err error
+		fs := utils.FS(ctx)
+		content, err = afero.ReadFile(fs, input.SnapshotSource)
+		if err != nil {
+			log.Debugf("could not read --snapshot-source as a file: %v", err)
```
lcarva, line 91: The actual name of the parameter is not known by this package. Consider removing it from the error message.

internal/applicationsnapshot/input.go
```diff
@@ -154,3 +164,15 @@ func DetermineInputSpec(ctx context.Context, input Input) (*app.SnapshotSpec, er
 
 	return &snapshot.SnapshotSpec, nil
 }
+
+func readSnapshotFile(input []byte) (app.SnapshotSpec, error) {
```
lcarva, line 168: It's not a file though.

lcarva (approved):

PR #836, opened 2023-07-18
Add ec test command that mimics conftest test & supports the appstudio output format

simonbaird:
The new `ec test` command acts like plain conftest, but supports an additional output format `appstudio` which creates a `TEST_OUTPUT` style json for use in a Tekton result.

[HACBS-2376](https://issues.redhat.com/browse/HACBS-2376)


cmd/test/test.go
```diff
@@ -0,0 +1,103 @@
...
+			fmt.Print(string(jsonOut))
+			return
+		},
+	}
+
+	cmd.Flags().StringArrayVarP(&flagPolicy, "policy", "p", flagPolicy, "path to the Rego policy files directory")
+	cmd.Flags().StringArrayVarP(&flagData, "data", "d", flagData, "a list of paths from which data for the rego policies will be recursively loaded")
+	cmd.Flags().StringArrayVarP(&flagNamespace, "namespace", "n", flagNamespace, "test policies in a specific namespace")
+	cmd.Flags().BoolVar(&flagAllNamespaces, "all-namespaces", flagAllNamespaces, "test policies found in all namespaces")
+	cmd.Flags().BoolVar(&flagNoFail, "no-fail", flagNoFail, "return an exit code of zero even if a policy fails") // Todo: Does it work?
```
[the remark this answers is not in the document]
simonbaird: Fixed now.

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
+// Copyright 2022 Red Hat, Inc.
```
lcarva, line 1: Update year?

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
...
+
+const (
+	OutputAppstudio = "appstudio"
+)
+
+var TestCmd *cobra.Command
+
+func fetchTestCmd() *cobra.Command {
+	flags := struct {
+		Policy             []string
```
lcarva, line 44: Could these be lower case?

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
...
+	}{}
+
+	cmd := &cobra.Command{
+		Use:   "test [flags] input",
+		Short: "A wrapper for the conftest test command",
+
+		Long: hd.Doc(`
+			The ec test command is meant to work as a drop-in replacement for conftest test.
+
+			It supports most but not all of the flags supported by Conftest. Its supports one
```
lcarva, line 64: `s/Its/It`

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
...
+				ec test --policy my_policies.rego --data my_data.json --namespace main input.json
+
+			Like Conftest it will traverse directories to find multiple policies and data files:
+
+				ec test --policy ./policy --data ./data --namespace main input.json
+		`),
+
+		RunE: func(cmd *cobra.Command, fileList []string) (err error) {
+			if len(fileList) < 1 {
+				return fmt.Errorf("at least one input file must be specified")
```
lcarva, line 81: Consider adding something like this to the cmd object instead:
```go
Args: cobra.MaximumNArgs(1)
```

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
...
+
+		Long: hd.Doc(`
+			The ec test command is meant to work as a drop-in replacement for conftest test.
+
+			It supports most but not all of the flags supported by Conftest. Its supports one
+			more output format than Conftest, the 'appstudio' format, which is suitable for use
+			with the standard AppStudio TEST_OUTPUT Tekton result format.
+		`),
+
+		Example: hd.Doc(`
```
lcarva, line 69: It might be a good idea to add an example of using multiple policies and multiple data.

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+
+			r := &runner.TestRunner{
+				Policy:             flags.Policy,
+				Data:               flags.Data,
+				Namespace:          flags.Namespace,
+				AllNamespaces:      flags.AllNamespaces,
+				NoFail:             flags.NoFail,
+				FailOnWarn:         flags.FailOnWarn,
+				Output:             flags.Output,
```
lcarva, line 91: Is it a problem if `flags.Output` is `appstudio`?

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+
+			if flags.Output == OutputAppstudio {
+				// The appstudio format is unknown to Conftest so we handle it ourselves
+				report := applicationsnapshot.AppstudioReportFromCheckResults(results)
+				reportOutput, err := json.Marshal(report)
+				if err != nil {
+					return fmt.Errorf("output results: %w", err)
+				}
+				fmt.Printf("%s", reportOutput)
```
lcarva, line 110: For easier testing, consider using `cmd.OutOrStdout()`.

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
...
+				})
+
+				if err := o.Output(results); err != nil {
+					return fmt.Errorf("output results: %w", err)
+				}
+			}
+
+			exitCode := chooseExitCode(r.NoFail, r.FailOnWarn, results)
+			os.Exit(exitCode)
+			return
```
lcarva, line 128: Consider something like this:
```go
return chooseExitError(r.NoFail, r.FailOnWarn, results)
```

cmd/test/test.go
```diff
@@ -0,0 +1,165 @@
...
+		Trace              bool
+		JunitHideMessage   bool
+	}{}
+
+	cmd := &cobra.Command{
+		Use:   "test [flags] input",
+		Short: "A wrapper for the conftest test command",
+
+		Long: hd.Doc(`
+			The ec test command is meant to work as a drop-in replacement for conftest test.
```
lcarva, line 62: I do like this approach, but I wonder if this will be confusing to users familiar with the `validate image` command. We do, in some sense, extend conftest to provide things like filtering, rule codes, descriptions, etc. It would be nice to make those features applicable here as well at some point, maybe?

(This is not a blocker for this PR, of course.)

cmd/test/test_test.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2022 Red Hat, Inc.
```
lcarva, line 1: Update year?

cmd/test/test_test.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build unit
+
+package test
+
+// See test coverage in features/conftest_test.feature
```
lcarva, line 21: Do we need this file at all then?

lcarva (approved):

cmd/test/test.go
simonbaird: https://github.com/enterprise-contract/ec-cli/pull/840

Did I overdo it?

cmd/test/test.go
simonbaird: Yep. Thanks.

cmd/test/test.go
simonbaird: I had that to begin with but switched it out because the error message is less friendly. I've got a better version in the next revision that uses the `cobra.MinimumNArgs(1)` but with a customized message.

cmd/test/test.go
simonbaird: Done.

simonbaird: Beware I don't think we can send the standard conftest output to anything other than os.Stdout.

cmd/test/test.go
simonbaird: I'm not sure how to do that, but I agree it is something to explore. Maybe we introduce a new command `ec validate input` that has a conftest output mode.

cmd/test/test_test.go
simonbaird: I was thinking it would help a reader find the real test coverage, but on second thoughts it's not needed. Removed.

cmd/test/test.go
simonbaird: Not applicable now that I'm trying to use conftest code as-is.

cmd/test/test.go
```diff
@@ -0,0 +1,228 @@
...
+	cmd.Flags().StringSlice("proto-file-dirs", []string{}, "A list of directories containing Protocol Buffer definitions")
+
+	return &cmd
+}
+
+var TestCmd *cobra.Command
+
+func init() {
+	TestCmd = NewTestCommand()
+}
```
simonbaird, line 228: These lines are not in the conftest version since they do the context handling slightly differently. We could tweak that maybe, but I don't think it's needed in this PR.

cmd/test/test.go
```diff
@@ -0,0 +1,228 @@
...
+See the pull command for more details on supported protocols for fetching policies.
+
+When debugging policies it can be useful to use a more verbose policy evaluation output. By using the '--trace' flag
+the output will include a detailed trace of how the policy was evaluated, e.g.
+
+	$ ec test --trace <input-file>
+`
+
+// NewTestCommand creates a new test command.
+func NewTestCommand() *cobra.Command {
```
simonbaird, line 100: This could be `newTestCommand` but I'm leaving it like this so the diff is easier to read...

robnester-rh: Capitalization invokes exportation, if that's not the intent, this should be left lowercased.

cmd/test/test.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/open-policy-agent/conftest/runner"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/enterprise-contract/ec-cli/internal/applicationsnapshot"
+)
+
+const (
+	OutputAppstudio = "appstudio"
+)
```
robnester-rh, line 39: Parentheses are not needed when defining a single constant as you're doing here. Either remove parentheses or include line numbers 41-97 within parenthetical enclosures.

cmd/test/test.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	cmd.Flags().Bool("fail-on-warn", false, "Return a non-zero exit code if warnings or errors are found")
+	cmd.Flags().Bool("no-fail", false, "Return an exit code of zero even if a policy fails")
+	cmd.Flags().Bool("no-color", false, "Disable color when printing")
+	cmd.Flags().Bool("suppress-exceptions", false, "Do not include exceptions in output")
+	cmd.Flags().Bool("all-namespaces", false, "Test policies found in all namespaces")
+	cmd.Flags().Bool("quiet", false, "Disable successful test output")
+
+	cmd.Flags().BoolP("trace", "", false, "Enable more verbose trace output for Rego queries")
```
robnester-rh, line 203: You're not specifying a single letter option here. Linked 203-205 can use `cmd.Flags().Bool` in that case.

cmd/test/test.go
```diff
@@ -0,0 +1,228 @@
...
+	cmd.Flags().BoolP("strict", "", false, "Enable strict mode for Rego policies")
+	cmd.Flags().BoolP("combine", "", false, "Combine all config files to be evaluated together")
+
+	cmd.Flags().String("ignore", "", "A regex pattern which can be used for ignoring paths")
+	cmd.Flags().String("parser", "", fmt.Sprintf("Parser to use to parse the configurations. Valid parsers: %s", parser.Parsers()))
+	cmd.Flags().String("capabilities", "", "Path to JSON file that can restrict opa functionality against a given policy. Default: all operations allowed")
+	validOutputs := append(output.Outputs(), OutputAppstudio)
```
robnester-rh, line 210: I'd rather see this above this section dealing with flags. I understand that the next flag concerns output, but it feels unusual for this to be tucked in here.

cmd/test/test.go
```diff
@@ -0,0 +1,228 @@
...
+	$ ec test --trace <input-file>
+`
+
+// NewTestCommand creates a new test command.
+func NewTestCommand() *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "test <path> [path [...]]",
+		Short: "Test your configuration files using Open Policy Agent",
+		Long:  testDesc,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
robnester-rh, line 105: If the intention is to check that each of these flags have a value or to make them required, see https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired and https://github.com/spf13/cobra/blob/v1.7.0/shell_completions.go#L38 for better way of achieving this.

robnester-rh (commented):
This seems to be bringing direct dependencies on `viper` which I don't think we need to introduce. It seems that the usage of functions from the `viper` package can be easily replaced with standard library packages.

cmd/test/test.go
simonbaird: Yeah, I did it on purpose since it reduces the diff between the conftest file and our file. I could maybe add a comment to explain.

cmd/test/test.go
simonbaird: I agree, but this is verbatim upstream Conftest code, so the fix should be in conftest.

cmd/test/test.go
simonbaird: Ack.

cmd/test/test.go
robnester-rh: I think it would be better to follow the idioms of Go and not capitalize function names if they're not intended to be exported.

simonbaird: Yeah ok.

cmd/test/test.go
simonbaird: Here we go: https://github.com/open-policy-agent/conftest/pull/847 :grin:

Ps, if our code diverges too much from the original code it becomes harder to maintain when/if the original code changes. So there's a tradeoff I think.

cmd/test/test.go
simonbaird: As mentioned, the idea is to make it the same as conftest. I want to avoid refactoring conftest's flag handling if I can.

cmd/test/test.go
simonbaird: No, it's defined as a string.

cmd/test/test.go
simonbaird: Done in latest revision.

cmd/test/test.go
simonbaird: Fixed in latest revision.

cmd/test/test.go
simonbaird: Fixed in latest revision.

simonbaird:
Latest revision has suggestions from @robnester-rh included, but so far it's sticks with the viper dependency and the flag verbatim flag parsing from [Conftest](https://raw.githubusercontent.com/open-policy-agent/conftest/v0.44.1/internal/commands/test.go) .

zregvart (approved):
I'm not 100% sure that we'll stick with this, is this something we want to have in the future or are we experimenting with this? If it's the latter we might want to put this behind the `EC_EXPERIMENTAL` flag.
Other than that LGTM

zregvart:
A followup might be adding support for restricting capabilities of OPA.

internal/applicationsnapshot/report.go
```diff
@@ -254,16 +258,47 @@ func (r *Report) toAppstudioReport() testReport {
...
+	if len(namespaces) > 0 {
+		// The first namespace only
+		useNamespace = namespaces[0]
+	}
+	report := testReport{
+		Timestamp: fmt.Sprint(time.Now().UTC().Unix()),
+		Namespace: useNamespace,
+	}
+
+	for _, result := range results {
```
lcarva, line 284: Do we need to handle skipped here?

lcarva (approved):

internal/applicationsnapshot/report.go
simonbaird: (After some slack discussion) I did add extra test coverage for result "SKIPPED". Might need some more work in future around Conftest exceptions.

simonbaird:
Let's merge.

PR #838, opened 2023-07-18
Add new policy-input output format

lcarva:
The policy-input format exports the image data that is provided to the policy as `input` in the `validate image` command. Given that multiple images can be validated at the same time, and that ec generates a unique `input` for each image, the output is in the JSONL format. A policy, however, only sees one of the objects at a time.

This PR also changes the variables `REGISTRY_*_HASH` to `REGISTRY_*_DIGEST` in the acceptance tests. The value of this new variable does not include the algorithm, i.e. `sha256`. This is needed because the image-data output contains fields where only the digest is present, without the algorithm. Having both variables is problematic because their values overlap. This means that whether or not the value is replaced with one vs the other is not deterministic.

https://issues.redhat.com/browse/HACBS-2393

### DIFF
--- a/acceptance/cli/cli.go
+++ b/acceptance/cli/cli.go
@@ -300,7 +300,11 @@ func setupRegistry(ctx context.Context, vars map[string]string, environment []st
 	}
 
 	for repositoryAndTag, hash := range hashes {
-		vars[fmt.Sprintf("REGISTRY_%s_HASH", repositoryAndTag)] = hash
+		_, digest, found := strings.Cut(hash, ":")
+		if !found {
+			return environment, vars, fmt.Errorf("hash %q does not contain digest", hash)
+		}
+		vars[fmt.Sprintf("REGISTRY_%s_DIGEST", repositoryAndTag)] = digest
 	}
 
 	return environment, vars, nil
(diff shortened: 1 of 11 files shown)


lcarva:
I'll share my Jira comment that explains my reasoning for taking this approach:
> An additional format to `ec validate image` is an interesting idea. On one hand, it leverages the existing code that fetches the data for validation making it less likely to diverge over time. On the other hand, it requires signature checks to succeed, making it harder to use in some contexts. It's important to note that the cosign calls that fetches the signatures and attestations are the same calls that validate against a signature (or fulcio identity information). Separating those may introduce security flaws.
>
> I'm gonna give this approach a try given that it sounds like a really simple way to solve a problem without adding long-term maintenance, and without potentially introducing a security flaw.

zregvart (commented):
Why not call it `policy-input`?

lcarva:
I think I'm gonna change the output format to be JSONL.

zregvart (approved):

simonbaird (approved):
One question: Does it handle the case where there are multiple images being checked, e.g. when the input is a snapshot image list?

lcarva:
> One question: Does it handle the case where there are multiple images being checked, e.g. when the input is a snapshot image list?

Yup! Each line would be a `policy-input` json.

PR #845, opened 2023-07-20
Rework key/signature variables

zregvart:
This reworks the replacement variables in acceptance tests and removes the special cases for JSON and XML. Given that only values are replaced we no longer need to render the whole JSON/XML structure.


joejstuart (approved):
outstanding. thank you @zregvart.

lcarva (approved):

lcarva:
The acceptance tests pass for me locally. We have a known issue where they're not working on GitHub which is unrelated to this PR. I'm gonna merge it as this solves other issues.

PR #846, opened 2023-07-20
Tweak the licence preamble

simonbaird:
The guidance at [1] suggests removing the year entirely so let's do that. Also let's changed the copyright holder to "The Enterprise Contract Contributors" which we decided on over "The Enterprise Contract Authors" or just "Red Hat". (The guide recommends against "Red Hat, Inc".)

Adjust the Makefile options to use that format for new files, and modify the existing preambles with some handy sed -i commands, something like this:

    git grep -l ... | xargs -n 1 sed -i 's/.../.../'

[1] https://source.redhat.com/departments/legal/redhatintellectualproperty/intellectual_property_and_ip_litigation_wiki/copyright_notices_in_source_code

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,4 @@
-# Copyright 2022 Red Hat, Inc.
+# Copyright The Enterprise Contract Contributors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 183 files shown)


lcarva (approved):

simonbaird:
Test failure is a known issue with docker in acceptance tests on GitHub.

PR #852, opened 2023-07-21
Fix acceptance tests container launching issue

zregvart:
A change[1] was backported to golang 1.19 that introduces additional checks on the `Host` HTTP header. The docker client connecting via Unix domain socket to the docker daemon does not specify the `Host` header and that triggers the error `http: invalid Host header`.

The fix[2] was introduced in an unreleased version[3] of docker client which we use via commit id here to resolve the issue.

[1] https://github.com/golang/go/issues/61075
[2] https://github.com/moby/moby/issues/45935
[3] https://github.com/moby/moby/pull/45971


zregvart:
Included #848 to reduce the CI churn

zregvart:
Merging to unblock other PRs

PR #854, opened 2023-07-21
Document policy-input

lcarva:
https://issues.redhat.com/browse/HACBS-2388

### DIFF
--- a/docs/modules/ROOT/pages/policy_input.adoc
+++ b/docs/modules/ROOT/pages/policy_input.adoc
@@ -1,0 +1,74 @@
+= Policy Input
+
+The policy input refers the variable `input` in rego. This page describes the structure of the input
+available for policy evaluation when using the Enterprise Contract CLI. The input is different for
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/policy_input.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+The input format for the `validate image` command describes all the information related to the image
+being available. Not all information is available for all images. This is noted below in the
+description of each attribute.
+
+It is important to note that this command may proccess multiple images at the same time. In such
+cases, a different input is generated for each image. In other words, policy rules only eve have
+access to the information about a single image.
+
+[,json]
```
lcarva, line 17: We could have a proper jsonschema but this might be sufficient until that need becomes obvious. What do other think of this approach?

zregvart: Works for now

zregvart (approved):

PR #855, opened 2023-07-21
Fix random order of var replacement in snapshots

lcarva:
Prior to this change, some variables were mapping to overlappling values:

```
REGISTRY_acceptance/image/parent:latest_JSON_DIGEST👉"c88bcb8e9c4827b15e33db83849b678220f1a88fc654090a0c9ac08038b73721"}
REGISTRY_acceptance/image/parent:latest_DIGEST👉c88bcb8e9c4827b15e33db83849b678220f1a88fc654090a0c9ac08038b73721
```

This caused two problems. First, the value of the variables ending in JSON_DIGEST causes the underlying JSON to break due to the closing `}` at the end. It also includes quotes which could potentially cause other issues. Second, because values overlap, and the order in which variable replacement happens in unpredictable, re-running the acceptance tests may result in a different variable being used. This resulted in random snapshot mismatches in CI.

With this change, we do away with the `JSON_DIGEST` variables altogether and instead rely on just the `DIGEST` variables.

For the `validate_image` acceptance tests, the variables are also now logged when a failure occurs. This makes issues like this much easier to debug.

### DIFF
--- a/acceptance/cli/cli.go
+++ b/acceptance/cli/cli.go
@@ -294,16 +294,12 @@ func setupRegistry(ctx context.Context, vars map[string]string, environment []st
 
 	vars["REGISTRY"] = registryURL
 
-	hashes, err := registry.AllHashes(ctx)
+	digests, err := registry.AllDigests(ctx)
 	if err != nil {
 		return environment, vars, err
 	}
 
-	for repositoryAndTag, hash := range hashes {
-		_, digest, found := strings.Cut(hash, ":")
-		if !found {
-			return environment, vars, fmt.Errorf("hash %q does not contain digest", hash)
-		}
+	for repositoryAndTag, digest := range digests {
 		vars[fmt.Sprintf("REGISTRY_%s_DIGEST", repositoryAndTag)] = digest
 	}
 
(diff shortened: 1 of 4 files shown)


acceptance/cli/cli.go
```diff
@@ -294,16 +294,12 @@ func setupRegistry(ctx context.Context, vars map[string]string, environment []st
 
 	vars["REGISTRY"] = registryURL
 
-	hashes, err := registry.AllHashes(ctx)
+	hashes, err := registry.AllDigests(ctx)
```
robnester-rh, line 297: For consistency should this be updated to a variable name of `digests`?

acceptance/kubernetes/kubernetes.go
```diff
@@ -270,13 +270,13 @@ func taskLogsShouldMatchTheSnapshot(ctx context.Context, stepName string) error
 		vars[fmt.Sprintf("____%s_PUBLIC_KEY", name)] = indent(key, 4)
 	}
 
-	hashes, err := registry.AllHashes(ctx)
+	hashes, err := registry.AllDigests(ctx)
```
robnester-rh, line 273: Nit: as above, consider variable renaming.

robnester-rh (approved):
Just a couple of nits, otherwise LGTM

zregvart (commented):
Thanks :100:

PR #856, opened 2023-07-24
Upload statistics to website

zregvart:
This will publish a JSONL file to

    https://enterprisecontract.dev/ec-cli/stats.json

Which we can use to monitor the download statistics which are currently lost each time we perform a release.

### DIFF
--- a/hack/stats.sh
+++ b/hack/stats.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+# Copyright The Enterprise Contract Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


hack/stats.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o posix
+
+mkdir -p stats
+# shellcheck disable=SC2016
+{
+    curl --silent --fail https://enterprisecontract.dev/ec-cli/stats.json || echo -n ''
```
lcarva, line 26: What's the purpose of this curl? The output doesn't seem to be used.

lcarva (approved):

PR #858, opened 2023-07-25
Increase number of returned assets in stats

zregvart:
We limit at 10 and we, currently, have 14 assets. This bumps to 50.


zregvart:
Merging to fix the stats.

PR #860, opened 2023-07-25
Refactor fetchers

zregvart:
Moves `internal/fetcher/image_config` to `internal/fetchers/oci/[config|fake]`.

This allows evolution of fetchers and implementation of different OCI related fetchers.


lcarva (approved):
Nice!

PR #863, opened 2023-07-25
Small NewPolicy refactor

lcarva:
This PR includes the following:

```
Add policy.Options to more easily call NewPolicy

The amount of parameters to `policy.NewPolicy` is high. Most of them are
of string type. This makes it easy to make mistakes when calling the
function.

This commit changes the function signature to take two parameters, a
context and a new struct `policy.Options`.
```

```
Update enterprise-contract-controller

This update brings in a new resource that cannot be loaded on a
non-OpenShift cluster. Thus, it is removed in the kustomization used
during the acceptance tests.
```

```
Use identity from ECP

The EnterpriseContractPolicy now includes a new attribute, Identity.
This commit uses that information if it is provided.
```

This is in preparation to https://issues.redhat.com/browse/HACBS-2165 where we may add a new CLI parameter to ignore Rekor instead of inferring the behavior from the Rekor URL parameter. Adding yet another parameter to `policy.NewPolicy` didn't seem right.


internal/policy/policy.go
```diff
@@ -410,6 +429,13 @@ func validateIdentity(identity cosign.Identity) error {
 	return errs
 }
 
+func emptyIdentity(identity cosign.Identity) bool {
```
robnester-rh, line 432: If you're looking to check if all fields in this `identity` struct are their zero value (empty string in the case of these fields), you might consider using the `.isZero()` method from the `reflect` package. This checks to see if the I think this could eliminate the need for this function, as well as change the call on line 191 to something like:

```
if !reflect.ValueOf(opts.Identity).isZero(){
    p.identity = opts.Identity
```

robnester-rh (approved):
One small suggestion regarding checking empty structs, but not a blocker. LGTM

internal/policy/policy.go
lcarva: Thank you for this!

I started reading up on the reflect package and this caught my eye:
> It panics if the argument is invalid. 

Given that reflection is quite slow and the possibility of panicking, I decided to go with a different approach:

```go
opts.Identity != (cosign.Identity{})
```

That seems safer to me. Let me know what you think.

robnester-rh: I think this should work in this context as well. :+1:

PR #866, opened 2023-07-26
Add manifest files from the image to input

zregvart:
This adds manifest files from the image to the input under `image.manifests`. Files that are located in the `manifests` directory and can be parsed as JSON or YAML are considered.

Each file is parsed and converted to JSON and included in the input JSON tree verbatim.

Ref. https://issues.redhat.com/browse/HACBS-2379


zregvart:
This includes #865 as using the `FakeImage` mock resulted in a panic when trying to access `image.Layers()`.

Not 100% sure we want this data in `image.manifests`, perhaps `image.fs.manifests` could be better place?

Not sure if we need to limit the memory usage, say no manifest file larger than 100Kb will be added to the input?

We should, perhaps as a followup, look into caching the fetching of images, we probably do it more than once now.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -369,6 +383,7 @@ func (a *ApplicationSnapshotImage) WriteInputFile(ctx context.Context) (string,
 			Ref:        a.reference.String(),
 			Signatures: a.signatures,
 			Config:     a.configJSON,
+			Manifests:  a.manifests,
```
lcarva, line 386: I think these manifests should go under something like `olm` (Operator Lifecycle Manager). The term `manifests` on its own is too ambiguous. For example, Red Hat stuffs certain SBOM-like manifests under `/root/buildconfig`.

internal/fetchers/oci/manifest/manifest.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/enterprise-contract/ec-cli/internal/fetchers/oci"
+)
+
+func ImageManifests(ctx context.Context, ref name.Reference, opts ...remote.Option) (map[string]json.RawMessage, error) {
+	img, err := oci.NewClient(ctx).Image(ref, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	content := mutate.Extract(img)
```
lcarva, line 42: I'm concerned about doing this for every image ec process. It seems unnecessary. Can we check for the presence of a certain label first? Even just any label that starts with `operatorframework.io/`.

internal/fetchers/oci/manifest/manifest.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	log "github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
+
+	"github.com/enterprise-contract/ec-cli/internal/fetchers/oci"
+)
+
+func ImageManifests(ctx context.Context, ref name.Reference, opts ...remote.Option) (map[string]json.RawMessage, error) {
```
lcarva, line 36: This should probably be named to reflect that these are olm manifests.

internal/fetchers/oci/manifest/manifest.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, err
+	}
+
+	content := mutate.Extract(img)
+	defer content.Close()
+	archive := tar.NewReader(content)
+
+	manifests := map[string]json.RawMessage{}
+	for {
+		header, err := archive.Next()
```
lcarva, line 48: It would be nice if we could use `range` to iterate over this. I don't think it's possible though.

internal/fetchers/oci/manifest/manifest.go
```diff
@@ -0,0 +1,89 @@
...
+		if !strings.EqualFold(path.Dir(name), "manifests") {
+			continue
+		}
+
+		// we're only interested in JSON or YAML files
+		if ext := path.Ext(name); !strings.EqualFold(ext, ".yaml") && !strings.EqualFold(ext, ".json") {
+			continue
+		}
+
+		// TODO: large files could be an issue. We do need to read the archive
```
lcarva, line 68: [io.LimitedReader](https://pkg.go.dev/io#LimitedReader) might be something to consider later on.

internal/fetchers/oci/manifest/manifest.go
```diff
@@ -0,0 +1,89 @@
...
+
+		name := header.Name
+
+		// we're only interested in files in manifests/*
+		if !strings.EqualFold(path.Dir(name), "manifests") {
+			continue
+		}
+
+		// we're only interested in JSON or YAML files
+		if ext := path.Ext(name); !strings.EqualFold(ext, ".yaml") && !strings.EqualFold(ext, ".json") {
```
lcarva, line 64: Maybe also consider `.yml`?

features/validate_image.feature
```diff
@@ -757,3 +757,20 @@ Feature: evaluate enterprise contract
...
+      | manifests/some.crd.yaml                   | examples/some.crd.yaml                   |
+      | manifests/some.clusterserviceversion.yaml | examples/some.clusterserviceversion.yaml |
+      And a valid image signature of "acceptance/image" image signed by the "known" key
+      And a valid attestation of "acceptance/image" signed by the "known" key
+      And a git repository named "olm-manifests" with
+      | image_config.rego | examples/olm_manifests.rego |
+      And policy configuration named "policy" with specification
+    """
+    {"sources": [{"policy": ["git::https://${GITHOST}/git/olm-manifests.git"]}]}
+    """
```
simonbaird, line 772: Indent mismatch here maybe?

simonbaird: I mean line 770.

simonbaird:
Cool stuff.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
zregvart: I kinda think we want to be even broader than that, we could go with something like `image.fs`, and keep exposing just the `/manifests/*.(yaml|json)` for now. Once we have the ability to add custom functions to OPA this could be something like `oci.files("registry.io/repository...", ...)`

internal/fetchers/oci/manifest/manifest.go
zregvart: We check for presence of `operators.operatorframework.io.bundle.manifests.v1` in the rules, not saying that'll avoid doing unnecessary work, let me thinker a bit on this, I think I can make this semi-generic

internal/fetchers/oci/manifest/manifest.go
zregvart: I'm thinking more of `ImageFiles` now...

internal/fetchers/oci/manifest/manifest.go
zregvart: Not seeing a way to do it

internal/fetchers/oci/manifest/manifest.go
zregvart: I much rather have a (OPA) function and let the policy pull in what it needs, if we had a metric of how big the manifest files can get then perhaps, left a TODO to think about this in the future...

lcarva: works for now.

internal/fetchers/oci/files/files.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+			return nil, err
+		}
+
+		for _, f := range fetchers {
+			if data, err := f.fetch(header, archive); err != nil {
+				return nil, err
+			} else if data != nil {
+				manifests[header.Name] = data
+				break
```
lcarva, line 83: So we only care about the first successful fetcher? That seems odd. What if two fetchers want to fetch the same file? If a fetcher can manipulate the file it is fetching, `olmManifestExtractor` does, this will be a problem.

If we want to have multiple fetchers, we should probably namespace their result, e.g. `manifests.olm["manifests/crd.yaml"]`, where `olm` is the name of the fetcher.

internal/fetchers/oci/files/olm_manifest.go
```diff
@@ -0,0 +1,91 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	log "github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
+)
+
+const olm_manifest_v1 = "operators.operatorframework.io.bundle.manifests.v1"
+
+type olmManifestExtractor struct{}
+
+func (olmManifestExtractor) fetcherFor(img v1.Image) (fetcher, error) {
```
lcarva, line 35: Do we need a struct here? Maybe this can just be a standalone function?

internal/fetchers/oci/files/olm_manifest.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if path, ok := config.Config.Labels[olm_manifest_v1]; ok {
+		return &olmManifestFetcher{path}, nil
+	}
+
+	return nil, nil
+}
+
+type olmManifestFetcher struct {
```
lcarva, line 52: This is more like a filter/mapper than a fetcher.

internal/fetchers/oci/files/files.go
lcarva: Yup, we either have all the files being fetched the same way, or we need to namespace them. Either approach is fine IMO.

(We could also make the fetcher just return a list of files to fetch.)

lcarva (approved):
Very nice!

PR #867, opened 2023-07-26
Enable manual runs for "checks" and "lint" workflows

mbestavros:
This can be helpful for manually running tests before submitting a pull request to this repo.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -24,6 +24,7 @@ name: Checks
   push:
     branches:
       - main
+  workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


zregvart:
Given that these run whenever a change is pushed to (any) repository and if the checks fail they can be re-run, what is the utility of this?

lcarva (approved):
FWIW, I try to run `make ci` locally before opening a PR, or at least a relevant subset of it. I don't really like waiting for GitHub unless I need to, but that's just my personal preference. If this change makes your workflow easier, then by all means let's get this in.

PR #868, opened 2023-07-26
Allow duplicate digests in bundles tracked with `ec track bundle`

mbestavros:
This is a fairly simple patch that changes the logic for filtering out duplicate bundles. Now a bundle is only dropped if its digest is the same as _the previous_ bundle, instead of if its digest is the same as _any_ bundle.

Ref: https://issues.redhat.com/browse/HACBS-2394

### DIFF
--- a/internal/tracker/tracker.go
+++ b/internal/tracker/tracker.go
@@ -175,18 +175,19 @@ func filterRecords(records []bundleRecord, prune bool) []bundleRecord {
 	now := time.Now().UTC()
 
 	unique := make([]bundleRecord, 0, len(records))
-	keys := map[string]bool{}
-	for i := len(records) - 1; i >= 0; i-- {
+	last_index := len(records) - 1
+	for i := last_index; i >= 0; i-- {
 		r := records[i]
 		// NOTE: Newly added records will have a repository, but existing ones
 		// will not. This is expected because the output does not persist the
 		// repository for each record. Instead, the repository is the attribute
 		// which references the list of records.
-		key := r.Digest
-		if _, ok := keys[key]; ok {
-			continue
+		if i < last_index {
+			previous := records[i+1]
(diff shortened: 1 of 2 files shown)


lcarva (approved):

internal/tracker/tracker.go
```diff
@@ -175,18 +175,18 @@ func filterRecords(records []bundleRecord, prune bool) []bundleRecord {
...
 	for i := len(records) - 1; i >= 0; i-- {
 		r := records[i]
 		// NOTE: Newly added records will have a repository, but existing ones
 		// will not. This is expected because the output does not persist the
 		// repository for each record. Instead, the repository is the attribute
 		// which references the list of records.
-		key := r.Digest
-		if _, ok := keys[key]; ok {
-			continue
+		if i < len(records)-1 {
```
simonbaird, line 184: I would expect the `go fmt` would change this, but I might be wrong..

```suggestion
		if i < len(records) - 1 {
```

internal/tracker/tracker.go
```diff
@@ -175,18 +175,18 @@ func filterRecords(records []bundleRecord, prune bool) []bundleRecord {
 	now := time.Now().UTC()
 
 	unique := make([]bundleRecord, 0, len(records))
-	keys := map[string]bool{}
 	for i := len(records) - 1; i >= 0; i-- {
```
simonbaird, line 178: Maybe..?

    last_index := len(records) - 1

mbestavros: Good catch, added.

simonbaird (approved):
Lgtm, but consider nitpick.

Ps, another nitpick - if the tests are for the functionality your adding, it's better to put them in the same commit. Someone looking at the commit in future will hopefully get a better understanding.

simonbaird (approved):

PR #873, opened 2023-07-27
Make Rekor stubs more precise

lcarva:
Previously, the Rekor stubs for the `/api/v1/log/entries/retrieve` API endpoint were matching any requests that included the public key.

This commit changes the stubs to only match the expected content. Queries for attestations and signature are different, so a different selector is used for each.

### DIFF
--- a/acceptance/rekor/rekor.go
+++ b/acceptance/rekor/rekor.go
@@ -28,6 +28,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/cucumber/godog"
@@ -54,14 +55,6 @@ const rekorStateKey = key(0) // we store the gitState struct under this key in C
 type rekorState struct {
 	KeyPair *cosign.KeysBytes
 }
-
-type rekorImpl struct{}
-
-func (r rekorImpl) StubRekorEntryFor(ctx context.Context, data []byte) error {
-	return stubRekorEntryFor(ctx, data)
-}
-
-var impl testenv.Rekor = rekorImpl{}
 
 func (r rekorState) Key() any {
 	return rekorStateKey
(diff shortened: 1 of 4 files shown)


acceptance/rekor/rekor.go
```diff
@@ -280,17 +273,44 @@ func stubRekorEntryFor(ctx context.Context, data []byte) error {
...
 			200,
 		))
 }
 
+type jsonPathExtractor func([]byte) (string, error)
+
+// jsonPathFromSignature returns the JSON Path expression to be used in the wiremock stub
+// for a signature query. The expression matches the value of the signature's content.
+func jsonPathFromSignature(data []byte) (string, error) {
+	blah := cosign.Signatures{}
```
zregvart, line 295: (nit) `blah` looks a bit off
```suggestion
	signatures := cosign.Signatures{}
```

acceptance/rekor/rekor.go
```diff
@@ -280,17 +273,44 @@ func stubRekorEntryFor(ctx context.Context, data []byte) error {
...
+	if err := json.Unmarshal(data, &blah); err != nil {
+		return "", fmt.Errorf("unmarshalling signature: %w", err)
+	}
+
+	if blah.Sig == "" {
+		return "", fmt.Errorf("data missing 'sig' key: %s", data)
+	}
+
+	jsonPath := fmt.Sprintf("$..[?(@.content=='%s')]", blah.Sig)
+	return jsonPath, nil
```
zregvart, line 305: (nit)
```suggestion
	return fmt.Sprintf("$..[?(@.content=='%s')]", blah.Sig), nil
```

zregvart (approved):
:+1:

acceptance/rekor/rekor.go
lcarva: oops.. fixed it. Thanks!

PR #874, opened 2023-07-27
Add --ignore-rekor flag

lcarva:
Prior to this change, the flag --rekor-url had two purposes. To provide the URL for the Rekor service, and to determine whether or not Rekor should be used in the verification at all. This was problematic.

Depending on the item being verified, cosign may not use the provided Rekor URL during verification. Instead, it may switch to offline verification. This uses the Rekor public keys fetched from TUF. This led to the odd usage of having to provide a dummy value as the Rekor URL to trigger Rekor verification.

This commit introduces a new flag, `--ignore-rekor`, to explicitly skip Rekor during verification. The default value is `false` as that is the most secure option.

This has implications with backwards compatibility. Someone that wants to ignore Rekor when running `ec validate image` must now use the new flag.

This commit also adds a new parameter to the
`verify-enterprise-contract` task, aptly named `IGNORE_REKOR`. Contrary to the CLI flag, this parameter defaults to `true`. This is done for a temporary amount of time to allow users of the Task to start using the new parameter if needed. (Tekton does not allow passing parmeters that don't exist to a Task.) After a certain amount of time, the default value will be updated to `false` for security reasons.

https://issues.redhat.com/browse/HACBS-2165

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -52,6 +52,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		imageRef                    string
 		info                        bool
 		input                       string // Deprecated: images replaced this
+		ignoreRekor                 bool
 		output                      []string
 		outputFile                  string
 		policy                      policy.Policy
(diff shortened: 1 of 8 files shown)


lcarva:
Filed a follow up [issue](https://issues.redhat.com/browse/HACBS-2412) that will help us improve acceptance test coverage for `task_validate_image`. I do think there's sufficient coverage in this PR as is.

cmd/validate/image.go
```diff
@@ -363,6 +365,9 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 	cmd.Flags().StringVarP(&data.rekorURL, "rekor-url", "r", data.rekorURL,
 		"Rekor URL. Overrides rekorURL from EnterpriseContractPolicy")
 
+	cmd.Flags().BoolVar(&data.ignoreRekor, "ignore-rekor", data.ignoreRekor,
+		"Skip Rekor checks during validation.")
```
simonbaird, line 369: Maybe..?

```suggestion
		"Skip Rekor transparency log checks during validation.")
```

simonbaird (approved):

cmd/validate/image.go
lcarva: Good idea. Also added to the task param.

cmd/validate/image_test.go
zregvart, line 1: :+1: nice

zregvart (approved):

PR #876, opened 2023-07-28
Use filesystem image cache

zregvart:
This makes sure that we cache remote image fetches by using a filesystem cache set to a path in the user cache directory. The cache is disabled if `EC_CACHE` environment variable is set to to `false` according to `strconv.ParseBool`.

### DIFF
--- a/internal/fetchers/oci/client.go
+++ b/internal/fetchers/oci/client.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Red Hat, Inc.
+// Copyright The Enterprise Contract Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


internal/fetchers/oci/client.go
```diff
@@ -18,27 +18,58 @@ package oci
...
+var defaultClient = remoteClient{}
+
 func NewClient(ctx context.Context) client {
 	c, ok := ctx.Value(clientKey).(client)
 	if ok && c != nil {
 		return c
 	}
 
-	return &remoteClient{}
+	return &defaultClient
```
zregvart, line 72: I thought this might make sense to prevent allocation

zregvart:
I had this on #866 it is unrelated so I created a separate PR

internal/fetchers/oci/client_test.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2023 Red Hat, Inc.
```
lcarva, line 1: `Copyright The Enterprise Contract Contributors`

We should probably fix it on the other files under fetchers as well. I think those were missed after the mass update.

lcarva (approved):

lcarva (approved):

PR #879, opened 2023-07-28
Isolate dependency on conftest

lcarva:
This commit improves the Evaluator interface to use its own types instead of the ones defined by conftest.

Conftest is now only used on the places that leverage it, such as `cmd/test`, `internal/downloader`, and `conftest_evaluator`.

### DIFF
--- a/internal/evaluator/evaluator.go
+++ b/internal/evaluator/evaluator.go
@@ -21,12 +21,29 @@ import (
 )
 
 type Evaluator interface {
-	// TODO refactor not to expose Conftest type here
-	Evaluate(ctx context.Context, inputs []string) (CheckResults, Data, error)
+	Evaluate(ctx context.Context, inputs []string) ([]Outcome, Data, error)
 
 	// Destroy performs any cleanup needed
 	Destroy()
 
 	// CapabilitiesPath returns the path to the file where capabilities are defined
 	CapabilitiesPath() string
+}
+
+type Data map[string]any
+
+type Outcome struct {
+	FileName   string   `json:"filename"`
+	Namespace  string   `json:"namespace"`
+	Successes  []Result `json:"successes,omitempty"`
+	Skipped    []Result `json:"skipped,omitempty"`
+	Warnings   []Result `json:"warnings,omitempty"`
+	Failures   []Result `json:"failures,omitempty"`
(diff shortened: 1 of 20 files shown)


internal/evaluator/evaluator.go
```diff
@@ -21,12 +21,29 @@ import (
...
 	// Destroy performs any cleanup needed
 	Destroy()
 
 	// CapabilitiesPath returns the path to the file where capabilities are defined
 	CapabilitiesPath() string
 }
+
+type Data map[string]any
+
+type Result struct {
```
zregvart, line 35: Perhaps
```suggestion
type Outcome struct {
```

internal/evaluator/evaluator.go
```diff
@@ -21,12 +21,29 @@ import (
...
+	FileName   string `json:"filename"`
+	Namespace  string `json:"namespace"`
+	Successes  []Rule `json:"successes,omitempty"`
+	Skipped    []Rule `json:"skipped,omitempty"`
+	Warnings   []Rule `json:"warnings,omitempty"`
+	Failures   []Rule `json:"failures,omitempty"`
+	Exceptions []Rule `json:"exceptions,omitempty"`
+}
+
+type Rule struct {
```
zregvart, line 45: and then:

```suggestion
type Result struct {
```

internal/evaluator/evaluator.go
```diff
@@ -21,12 +21,29 @@ import (
...
+	Message  string                 `json:"msg"`
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
+	Outputs  []string               `json:"outputs,omitempty"`
```
zregvart, line 48: Are we using `Outputs`? Perhaps we can remove it?

zregvart (commented):
I think it's a bit awkward to convert `output.Result` to `evaluator.Rule`, I think we might have a `evaluator.Rule` at some point based on the parsed rules.

internal/evaluator/evaluator.go
lcarva: I like it.

internal/evaluator/evaluator.go
lcarva: Probably. I'd like to do that as a follow up given the size of this PR.

PR #893, opened 2023-08-02
Don't depend on redhat-appstudio/application-api

zregvart:
The `github.com/redhat-appstudio/application-api` has older dependencies that are not compatible with the dependencies that we need. This tries to solve that issue by vendoring the specific types we need in the new `application` module.

Ref. https://issues.redhat.com/browse/HACBS-2426


zregvart:
Set in draft so we can see if https://github.com/redhat-appstudio/application-api/pull/63 gets merged so the `replace` is not necessary here.

PR #902, opened 2023-08-04
Don't depend on redhat-appstudio/application-api

zregvart:
The `github.com/redhat-appstudio/application-api` has older dependencies that are not compatible with the dependencies that we need. This tries to solve that issue by vendoring the specific types we need in the new `application` module.

Ref. https://issues.redhat.com/browse/HACBS-2426


.github/workflows/codeql.yaml
```diff
@@ -70,8 +70,9 @@ jobs:
       - name: Download go dependencies
         run: |
           go mod download
-          go mod download -modfile tools/go.mod
-          go mod download -modfile acceptance/go.mod
+          (cd tools && go mod download)
```
lcarva, line 73: Just for my own understanding, why not use `-modfile`?

application/v1alpha1/README
```diff
@@ -0,0 +1,2 @@
+The files in this directory were copied from
+https://github.com/redhat-appstudio/application-api/tree/main/api/v1alpha1
```
lcarva, line 2: As a follow up, we could probably create a little script that automates this process. Then, we can run as a GitHub Workflow, warning us if any changes are found.

hack/boilerplate.go.txt
lcarva, line 1: What's this for?

lcarva (approved):

simonbaird (approved):
Idk, but lgtm.

lcarva:
I think we may want to isolate the usage of the structs from Application API and instead provide our own structs that can be used more deliberately. A user coming to EC has no idea what an ApplicationSnapshotSpec is. I don't really want to send them to Appstudio docs for that. We should own that data structure here and have adapters/converters for the structs from Application API. Thoughts?

.github/workflows/codeql.yaml
zregvart: [This](https://github.com/enterprise-contract/ec-cli/actions/runs/5761947101/job/15620788857?pr=902#step:7:7) happens:
```
go: github.com/enterprise-contract/ec-cli/application@v0.0.0-00010101000000-000000000000 (replaced by ../application/): reading ../application/go.mod: open /home/runner/work/ec-cli/application/go.mod: no such file or directory
```

hack/boilerplate.go.txt
zregvart: Leftover from code generators, will remove

application/v1alpha1/README
zregvart: Created [HACBS-2441](https://issues.redhat.com/browse/HACBS-2441), I think we want to have an approach that is triggered on version updates and doesn't block us in CI

zregvart:
> I think we may want to isolate the usage of the structs from Application API and instead provide our own structs that can be used more deliberately. A user coming to EC has no idea what an ApplicationSnapshotSpec is. I don't really want to send them to Appstudio docs for that. We should own that data structure here and have adapters/converters for the structs from Application API. Thoughts?

How would we maintain compatibility with App Studio in this scenario?

lcarva (approved):

PR #904, opened 2023-08-04
Display component source in report

lcarva:
This change makes the `source` information from the SnapshotComponent appear in the report of the `ec validate image` command.

Eventually, we want to propagate this information to the policy-input, but that is left for HACBS-2440.


lcarva:
Looks like the order of components in the output is random because we process them in parallel. I'll sort this out on Monday.

zregvart (approved):

PR #939, opened 2023-08-15
Allow not-yet-effective successes in output

mbestavros:
Pretty simple patch. Maybe we want to consider some additional annotation to indicate that this rule isn't yet effective?

Ref: https://issues.redhat.com/browse/HACBS-2449


robnester-rh (approved):
LGTM

simonbaird (approved):

PR #955, opened 2023-08-18
Experimental image build using Chainguard

simonbaird:
Based on [these docs](https://edu.chainguard.dev/chainguard/chainguard-images/reference/go/getting-started-go/#example-2--multistage-docker-build-for-go-chainguard-image).

Not sure if we want or need it, but I'll push a draft PR.

Size comparison:
- `snapshot       331 MB`
- `snapshot-cg    89.1 MB`

Notes:
- We include git-core and jq in the current image build, so the size comparison is not really fair.
- The idea of doing a multi-stage build with a build container may have some value, even if we don't end up using these base images.
- Currently the go modules are downloaded every time, which makes the build a bit slower.

### DIFF
--- a/Dockerfile.chainguard
+++ b/Dockerfile.chainguard
@@ -1,0 +1,11 @@
+FROM cgr.dev/chainguard/go AS builder
+ARG VERSION
+COPY . /app
+RUN cd /app && go build \
+  -trimpath \
+  -ldflags="-s -w -X github.com/enterprise-contract/ec-cli/internal/version.Version=${VERSION}" \
+  -o dist/ec
+
+FROM cgr.dev/chainguard/glibc-dynamic
+COPY --from=builder /app/dist/ec /usr/bin/
+ENTRYPOINT ["/usr/bin/ec"]
--- a/Makefile
+++ b/Makefile
@@ -229,6 +229,12 @@ ifdef ADD_IMAGE_TAG
 	@podman manifest push $(IMAGE_REPO):$(IMAGE_TAG) $(IMAGE_REPO):$(ADD_IMAGE_TAG)
 endif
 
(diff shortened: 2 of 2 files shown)


zregvart:
Note that without `git` executable go-getter will not work for git URLs.

simonbaird:
> Note that without `git` executable go-getter will not work for git URLs.

Yeah, it's a firm blocker. I thought maybe I could do an `apk install git` somewhere in the dockerfile, but iiuc the base image used here doesn't have apk. At this stage I probably won't try to take it further, but I'll leave it open for a bit longer.

lcarva:
For the sake of experimentation. You can build a tiny image that doesn't include any package managers. Something like this:

```Dockerfile
FROM registry.access.redhat.com/ubi9/ubi:9.2 AS micro-build

RUN mkdir -p /mnt/rootfs
RUN dnf install --installroot /mnt/rootfs git-core jq --nodocs -y --setopt install_weak_deps=false \
    && dnf --installroot /mnt/rootfs clean all
RUN rm -rf /mnt/rootfs/var/cache/* /mnt/rootfs/var/log/dnf* /mnt/rootfs/var/log/yum.*

FROM scratch

COPY --from=micro-build /mnt/rootfs/ /
```

That should give you an image that is smaller than ubi-minimal, but not by much. This is because git-core (mostly) and jq bring in a bunch of other deps (including bash :disappointed: ). On my local system, the Dockerfile above produces an image that is 107MB. The Dockerfile below produces an image that is 147MB. It does seem like a significant gain though.

```Dockerfile
FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2

RUN microdnf --nodocs -y install git-core jq && microdnf clean all
```

simonbaird:
Closing to de-clutter review rot dashboard.

PR #969, opened 2023-08-28
Use `--annotate` in `tkn bundle push`

zregvart:
Instead of downloading, modifying the manifest and re-uploading to the image registry this uses a new feature of `tkn` CLI to add the `org.opencontainers.image.revision` annotation when the image is initially constructed.


simonbaird (approved):

simonbaird:
For a tidier history you could split the TASK/TASKS diff to another commit.

PR #971, opened 2023-08-29
Fix `literals` syntax in kustomize.yaml patch

zregvart:
The data type is `string[]` not `map[string]string`.


zregvart:
Merged to get the infra-deployments PR fixed...

lcarva:
lgtm

PR #975, opened 2023-08-31
SBOM atts

simonbaird:


internal/utils/map_get.go
```diff
@@ -0,0 +1,66 @@
...
+func MapGetList(m map[string]any, key string) ([]any, error) {
+	return mapGet[[]any](m, key)
+}
+
+// Access a map inside a map with string keys
+func MapGetMap(m map[string]any, key string) (map[string]any, error) {
+	return mapGet[map[string]any](m, key)
+}
+
+func mapGet[T string | bool | int | []any | map[string]any](m map[string]any, key string) (T, error) {
```
zregvart, line 50: I'd probably have this as a [public interface](https://go.dev/play/p/UciROptYns7), and move this to `utils/maps`, so we could write something like `maps.Get(...)`

internal/attestation/untyped_test.go
```diff
@@ -0,0 +1,125 @@
...
+		setup func(l *mockSignature)
+		data  string
+	}{
+		// Note: A lot of the testing already done for SLSAProvenanceFromSignature
+		// in particular the error handling is covering code that is now in the
+		// PayloadFromSig function, which is also used by UntypedProvenanceFromSignature.
+		// So I don't think we should duplicate that error handling test coverage here.
+		// Additionally I don't want to refactor the existing SLSAProvenanceFromSignature
+		// testing, e.g. to split it up and move it around. So that's why the tests here
+		// are covering only some "happy path" scenarios and not the error handling.
```
zregvart, line 74: Nice pattern of putting an explainer here

internal/attestation/untyped.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Some important keys we're hoping to find in the opaque data
+const (
+	predicateTypeKey = "predicateType"
+	predicateKey     = "predicate"
+)
+
+// Similar function to in_toto.ProvenanceStatementSLSA02 but we'll use
+// an opaque untyped map
+type UntypedAttestation map[string]any
```
zregvart, line 40: Could  this be an array in some cases?

internal/attestation/untyped.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, AT005.CausedBy(err)
+	}
+
+	// Fix string predicate if needed
+	if statement.predicateIsString() {
+		fixedPredicate, err := statement.Predicate()
+		if err != nil {
+			return nil, AT002.CausedBy(err)
+		}
+		statement[predicateKey] = fixedPredicate
```
zregvart, line 72: Would it make sense to store this in a field of `untypedProvenance`?

internal/attestation/untyped.go
```diff
@@ -0,0 +1,161 @@
...
+
+type untypedProvenance struct {
+	statement  UntypedAttestation
+	data       []byte
+	signatures []signature.EntitySignature
+}
+
+func (u untypedProvenance) Type() string {
+	// Maybe we don't know the predicateType, but it's still (presumably) an
+	// in-toto statement. Question: Should we extract it instead of hard coding it?
```
zregvart, line 121: Perhaps hardcoding this could allow someone to make a unsupported claim and trick the rules, perhaps, not sure. Smells a bit to have it hardcoded. Versions also change so it might be an issue in the future that would be a bit difficult to troubleshoot

internal/attestation/untyped.go
```diff
@@ -0,0 +1,161 @@
...
+	return in_toto.StatementInTotoV01
+}
+
+// Extract a predicateType string from the statement
+func (u untypedProvenance) PredicateType() string {
+	if t, err := u.statement.PredicateType(); err != nil {
+		panic(fmt.Errorf("Problem extracting predicateType: %w", err))
```
zregvart, line 128: If the predicate type was in the field, we could return an error when trying to extract it in `UntypedProvenanceFromSignature` instead, and not panic here

internal/attestation/untyped.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (u untypedProvenance) Signatures() []signature.EntitySignature {
+	return u.signatures
+}
+
+func (u untypedProvenance) MarshalJSON() ([]byte, error) {
+	val := struct {
+		Type          string                      `json:"type"`
+		PredicateType string                      `json:"predicateType"`
+		Signatures    []signature.EntitySignature `json:"signatures"`
```
zregvart, line 153: Are we missing the statement here?

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -247,14 +247,39 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
...
+			return fmt.Errorf("Unable to parse untyped provenance: %w", err)
+		}
+		t := att.PredicateType()
+		switch t {
+		case attestation.PredicateSLSAProvenance:
+			// SLSAProvenanceFromSignature does the payload extraction
+			// and decoding that was done in UntypedProvenanceFromSignature
+			// over again. We could refactor so we're not doing that twice,
+			// but it's not super important IMO.
+			sp, err := attestation.SLSAProvenanceFromSignature(sig)
```
zregvart, line 262: Could be something like:
```suggestion
			sp, err := attestation.SLSAProvenanceFromUntyped(att)
```

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -247,14 +247,39 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
...
+		case attestation.PredicateSpdxDocument:
+			// Todo: Try unmarshaling to confirm the format is valid
+
+			// Keep for further use
+			a.attestations = append(a.attestations, att)
+
+		default:
+			// Todo: We should probably keep rather than discard attestations with
+			// an unknown predicateType
+			log.Debugf("Ignoring attestation with predicateType: %s", t)
```
zregvart, line 279: Would make sense for future proofing to not discard here

internal/utils/map_get.go
simonbaird: I think I tried that but was discourage when I learned that you can't do a generic in a interface method. But still the public methods could be done like this. I'll give it a try.

internal/attestation/untyped.go
simonbaird: It's a good question. I think probably no, but I'm not certain.

internal/attestation/untyped.go
simonbaird: Not sure I follow, do you mean to suggest we have the existing `statement` field (which might be "typed") as well as another `untypedStatement` field which is always un"typed"?

internal/attestation/untyped.go
simonbaird: Yeah, I think we can probably revisit this question in future rather than decide on it now.

internal/attestation/untyped.go
simonbaird: Ack, yeah. I think I know what you mean. Kind of like a "fail earlier" approach.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
simonbaird: Ah yes, that's a nice idea.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
simonbaird: Yeah, we could choose to be bold and do it now.

internal/attestation/untyped.go
zregvart: Relates to https://github.com/enterprise-contract/ec-cli/pull/975#discussion_r1312807789, for having the predicate type as a field on `untypedProvenance` and not modify the `statement`, i.e. provide it to the rules as is...

internal/attestation/untyped.go
simonbaird: I added a todo comment to figure this out.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
simonbaird: Decided to leave this for another day.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
simonbaird: Done.

simonbaird:
There are a few todos in the comments, but I think this could conceivably be merged now.

internal/attestation/untyped.go
simonbaird: Latest revision doesn't include this code any more, so resolving.

internal/attestation/attestation.go
```diff
@@ -38,3 +47,146 @@ type Attestation interface {
...
+
+	payloadBytes, err := io.ReadAll(reader)
+	if err != nil {
+		return payload, AT002.CausedBy(err)
+	}
+
+	err = json.Unmarshal(payloadBytes, &payload)
+	if err != nil {
+		return payload, AT002.CausedBy(err)
+	}
```
zregvart, line 82: If we wanted to conserve a bit of memory (*citation needed) this could go through a Decoder, something like:
```suggestion
	if err := json.NewDecoder(reader).Decode(&payload), err != nil {
		return payload, AT002.CausedBy(err)
	}
```

internal/attestation/attestation.go
```diff
@@ -38,3 +47,146 @@ type Attestation interface {
 	Statement() []byte
 	Signatures() []signature.EntitySignature
 }
+
+// Extract the payload from a DSSE signature OCI layer
+func PayloadFromSig(sig oci.Signature) (cosign.AttestationPayload, error) {
```
zregvart, line 52: Perhaps could be private?
```suggestion
func payloadFromSig(sig oci.Signature) (cosign.AttestationPayload, error) {
```

internal/attestation/attestation.go
```diff
@@ -38,3 +47,146 @@ type Attestation interface {
...
+
+// Todo: It seems odd that this does not contain the statement.
+// (See also the equivalent method in slsa_provenance_02.go)
+func (p provenance) MarshalJSON() ([]byte, error) {
+	val := struct {
+		Type          string                      `json:"type"`
+		PredicateType string                      `json:"predicateType"`
+		Signatures    []signature.EntitySignature `json:"signatures"`
+	}{
+		Type:          p.Type(), // ??
```
zregvart, line 186: Why the question marks?

zregvart (approved):
Let's merge this and do any followups afterwards. I think we need to validate the SPDX data, like we do for SLSA Provenance v0.2 using JSON Schema, that might be awkward it supports [many file formats](https://spdx.github.io/spdx-spec/v2.3/conformance/#44-standard-data-format-requirements)...

internal/attestation/attestation.go
simonbaird: You're right, thanks. Done in upcoming revision.

internal/attestation/attestation.go
simonbaird: Removed.

internal/attestation/attestation.go
simonbaird: It works, but I'll put it in a separate PR.

simonbaird:
I'll probably merge when it goes green.

PR #976, opened 2023-08-31
moving all references of hacbs-contract to enterprise-contract quay org

joejstuart:
https://issues.redhat.com/browse/HACBS-2047

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -40,6 +40,7 @@ jobs:
 
     env:
       IMAGE_REPO: quay.io/hacbs-contract/ec-cli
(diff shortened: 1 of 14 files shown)


hack/registry.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/sh
```
zregvart, line 1: Noting that this [duplicates](https://github.com/enterprise-contract/ec-cli/blob/f307f7afc70892a474a61d27ff29d3fbb35863b7/hack/test/kustomization.yaml#L24) the `hack/setup-[dev|test]-environment.sh`

joejstuart: good catch. removed

zregvart (approved):

PR #980, opened 2023-09-01
Set strict mode on by default

seanconroy2021:
From the Slack discussion, we decided to update ec-cli to have strict mode 'true.' 
Jira : [Link here ](https://issues.redhat.com/browse/HACBS-2584)

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -68,6 +68,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 		// Default policy from an ECP cluster resource
 		policyConfiguration: "enterprise-contract-service/default",
+		strict:              true,
 	}
 	cmd := &cobra.Command{
 		Use:   "image",
@@ -120,7 +121,11 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			Return a non-zero status code on validation failure:
 
-			  ec validate image --image registry/name:tag --strict
+			  ec validate image --image registry/name:tag 
+			
+		  	Return a zero status code even if there are validation failures:
+			
+			  ec validate image --image registry/name:tag --strict=false
 
 			Use an EnterpriseContractPolicy resource from the currently active kubernetes context:
 
@@ -402,7 +407,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		"[DEPRECATED] write output to a file. Use empty string for stdout, default behavior")
 
 	cmd.Flags().BoolVarP(&data.strict, "strict", "s", data.strict,
-		"return non-zero status on non-successful validation")
+		"Return non-zero status on non-successful validation. Defaults to true. Use --strict=false to return a zero status code.")
 
 	cmd.Flags().StringVar(&data.effectiveTime, "effective-time", policy.Now, hd.Doc(`
 		Run policy checks with the provided time. Useful for testing rules with
--- a/features/validate_image.feature
+++ b/features/validate_image.feature
@@ -29,7 +29,7 @@ Feature: evaluate enterprise contract
       ]
     }
     """
-    When ec command is run with "validate image --image ${REGISTRY}/acceptance/ec-happy-day --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --rekor-url ${REKOR} --strict --show-successes"
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/ec-happy-day --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --rekor-url ${REKOR}  --show-successes"
     Then the exit status should be 0
     Then the output should match the snapshot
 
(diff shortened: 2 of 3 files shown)


joejstuart:
LGTM. Just found `--strict` being used in a doc. https://github.com/seanconroy2021/ec-cli/blob/main/cmd/validate/image.go#L124

simonbaird (approved):

simonbaird:
> LGTM. Just found `--strict` being used in a doc. https://github.com/seanconroy2021/ec-cli/blob/main/cmd/validate/image.go#L124

Yeah, that should probably be updated a little to indicate the change. Also this part:

```
404     cmd.Flags().BoolVarP(&data.strict, "strict", "s", data.strict,                                                                                
405         "return non-zero status on non-successful validation")   
```

Something like "return non-zero status on non-successful validation. Defaults to true. Use `--strict=false` to return a zero status code" or similar in both places.

simonbaird (changes requested):
Removing approval to indicate I think a few doc wording changes would be a good idea.

simonbaird:
> Something like "return non-zero status on non-successful validation. Defaults to true. Use `--strict=false` to return a zero status code" or similar in both places.

Or maybe just "return non-zero status on non-successful validation. Use `--strict=false` to return a zero status code even when there are policy violations", since I think the default value appears automatically.

seanconroy2021:
> > Something like "return non-zero status on non-successful validation. Defaults to true. Use `--strict=false` to return a zero status code" or similar in both places.
> 
> Or maybe just "return non-zero status on non-successful validation. Use `--strict=false` to return a zero status code even when there are policy violations", since I think the default value appears automatically.

I will work on that this evening @simonbaird

seanconroy2021:
> > Something like "return non-zero status on non-successful validation. Defaults to true. Use `--strict=false` to return a zero status code" or similar in both places.
> 
> Or maybe just "return non-zero status on non-successful validation. Use `--strict=false` to return a zero status code even when there are policy violations", since I think the default value appears automatically.

I will work on that this evening @simonbaird

cmd/validate/image.go
```diff
@@ -402,12 +403,10 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		"[DEPRECATED] write output to a file. Use empty string for stdout, default behavior")
 
 	cmd.Flags().BoolVarP(&data.strict, "strict", "s", data.strict,
-		"return non-zero status on non-successful validation")
+		"Return non-zero status on non-successful validation. Defaults to true. Use --strict=false to return a zero status code.")
```
seanconroy2021, line 406: @simonbaird Is this better

seanconroy2021: comment from simonbarid: Yeah, that should probably be updated a little to indicate the change. Also this part:

404     cmd.Flags().BoolVarP(&data.strict, "strict", "s", data.strict,                                                                                
405         "return non-zero status on non-successful validation")   
Something like "return non-zero status on non-successful validation. Defaults to true. Use --strict=false to return a zero status code" or similar in both places.

cmd/validate/image.go
```diff
@@ -402,12 +403,10 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
...
 
 	cmd.Flags().StringVar(&data.effectiveTime, "effective-time", policy.Now, hd.Doc(`
-		Run policy checks with the provided time. Useful for testing rules with
-		effective dates in the future. The value can be "now" (default) - for
-		current time, "attestation" - for time from the youngest attestation, or
+		Run policy checks with the preturn non-zstation" - for time from the youngest attestation, or
```
zregvart, line 409: Should this be in this change? Seems off

zregvart (commented):
Not sure about the change in the `effective-time` flag

cmd/validate/image.go
seanconroy2021: @zregvart don't know what happened shouldn't have been changed fixed it thank you.

zregvart (approved):
Perhaps a squash before merge?

simonbaird:
Yeah, let's rebase and squash this into one tidy commit on top of main branch.

cmd/validate/image.go
```diff
@@ -120,7 +121,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			Return a non-zero status code on validation failure:
 
-			  ec validate image --image registry/name:tag --strict
+			  ec validate image --image registry/name:tag
```
simonbaird, line 124:
```suggestion
                        Return a zero status code even if there are validation failures:

			  ec validate image --image registry/name:tag --strict=false
```

features/validate_image.feature
```diff
@@ -471,7 +471,7 @@ Feature: evaluate enterprise contract
       ]
     }
     """
-    When ec command is run with "validate image --snapshot acceptance/happy --policy acceptance/ec-policy --rekor-url ${REKOR} --public-key ${known_PUBLIC_KEY} --strict  --show-successes"
+    When ec command is run with "validate image --snapshot acceptance/happy --policy acceptance/ec-policy --rekor-url ${REKOR} --public-key ${known_PUBLIC_KEY}   --show-successes"
```
simonbaird, line 474: Nitpick: There are some extra spaces here and there, e.g. `before --show-successes` in this line and some of the others. It doesn't matter much, but you could consider could tidying them up.

simonbaird:
Lgtm, but let's get rid of the merge commits, squash and rebase.

features/validate_image.feature
seanconroy2021: Ok think i fixed all the spacing sorry for that.

seanconroy2021:
@simonbaird this ready to be merged then

simonbaird (approved):

PR #981, opened 2023-09-01
Add simple demo script for keyless

simonbaird:
I find the demo scripts useful when hacking, so let's make one that can be used with a GitHub built image.

(Split out from #975 for early merge.)

### DIFF
--- a/hack/keyless-demo.sh
+++ b/hack/keyless-demo.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+# Copyright The Enterprise Contract Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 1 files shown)


hack/keyless-demo.sh
```diff
@@ -0,0 +1,37 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# This is an image built in GitHub, see https://github.com/lcarva/festoji
+IMAGE=${IMAGE:-"quay.io/lucarval/festoji:latest"}
+IDENTITY_REGEXP=${IDENTITY_REGEXP:-"https:\/\/github\.com\/(slsa-framework\/slsa-github-generator|lcarva\/festoji)\/"}
+IDENTITY_ISSUER=${IDENTITY_ISSUER:-"https://token.actions.githubusercontent.com"}
+
+# The EC golden container, see https://github.com/enterprise-contract/golden-container/
+#IMAGE=${IMAGE:-"ghcr.io/enterprise-contract/golden-container:latest"}
+#IDENTITY_REGEXP=${IDENTITY_REGEXP:-"https:\/\/github\.com\/"} # Fixme: What should a more restrictive regexp be here?
```
simonbaird, line 25: @seanconroy2021 FYI. Do you have a good regexp for matching the identities we trust to build the golden container?

seanconroy2021: I was doing it similar to Luiz's `Identity`,  to only two specific repositories, the `slsa-github-generator` & `golden-container`. (you may have a better way)
```suggestion
#IDENTITY_REGEXP=${IDENTITY_REGEXP:-"https:\/\/github\.com\/(slsa-framework\/slsa-github-generator|enterprise-contract\/golden-container)\/"}

simonbaird: Thanks @seanconroy2021. I pushed a revision that makes golden-container the default image using that regexp.

zregvart (approved):

PR #982, opened 2023-09-01
Add new map[string]any helper interface

simonbaird:
I think this will help make some code that deals with unmarshalled json cleaner and easier to read. With the unit tests in place I expect the maintenance burden should be low enough.

I had some fun trying out go generics, which seem to work well and make the code nicer.

This was created while hacking on HACBS-2515, which is about working with attestation predicates of an unknown format.


simonbaird:
Split out of #975 for review purposes.

internal/utils/msa/msa.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Short for MapStringAny
+package msa
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type MapStringAny map[string]any
```
zregvart, line 29: Perhaps, since the usefulness of this is related to JSON this could be:

```suggestion
type JSONObject map[string]any
```

internal/utils/msa/msa.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Create a MapStringAny from a plain map[string]any
+func FromMap(m map[string]any) MapStringAny {
+	return MapStringAny(m)
+}
+
+// Create a MapStringAny by unmarshaling json bytes
+func FromJsonBytes(bytes []byte) (MapStringAny, error) {
```
zregvart, line 37: If the struct is named `JSONObject` then this could be called

```suggestion
func FromBytes(data []byte) (JSONObject, error) {
```

internal/utils/msa/msa.go
```diff
@@ -0,0 +1,120 @@
...
+	err := json.Unmarshal(bytes, &m)
+	return m, err
+}
+
+// Create a MapStringAny by unmarshaling a json string
+func FromJson(str string) (MapStringAny, error) {
+	return FromJsonBytes([]byte(str))
+}
+
+// Convert back to a plain map[string]any
```
zregvart, line 48: Changing the resulting `map[string]any` will mutate `MapStringAny` might be unexpected, perhaps add that to the docs?

internal/utils/msa/msa.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Create a MapStringAny by unmarshaling json bytes
+func FromJsonBytes(bytes []byte) (MapStringAny, error) {
+	var m MapStringAny
+	err := json.Unmarshal(bytes, &m)
+	return m, err
+}
+
+// Create a MapStringAny by unmarshaling a json string
+func FromJson(str string) (MapStringAny, error) {
```
zregvart, line 44: For consistency...

```suggestion
func FromJSON(str string) (MapStringAny, error) {
```

internal/utils/msa/msa.go
```diff
@@ -0,0 +1,120 @@
...
+		err    error
+	)
+
+	if len(args) > 0 {
+		result, err = json.MarshalIndent(m, "", args[0])
+	} else {
+		result, err = json.Marshal(m)
+	}
+	if err != nil {
+		panic(fmt.Errorf("Unexpected marshalling problem: %w", err))
```
zregvart, line 66: Let's return an error instead, random panics are not nice to debug

internal/utils/msa/msa.go
```diff
@@ -0,0 +1,120 @@
...
+	return FromJsonBytes([]byte(str))
+}
+
+// Convert back to a plain map[string]any
+func (m MapStringAny) ToMap() map[string]any {
+	return map[string]any(m)
+}
+
+// Marshal to json bytes with optional indenting
+func (m MapStringAny) ToJSON(args ...string) []byte {
```
zregvart, line 54: Not seeing args[n], n > 0 being used, do we need the vararg here? Could use `len(indent) > 0`, or have two methods like `json.Marshal*`

internal/utils/msa/msa.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Get a float64 value from a map with string keys
+// (You might expect it to be an int, but a float64 is what
+// you get from the json decoder.)
+func (m MapStringAny) GetNum(key string) (float64, error) {
+	return mapGet[float64](m, key)
+}
+
+// Get a list from a map with string keys
+func (m MapStringAny) GetList(key string) ([]any, error) {
```
zregvart, line 94: Perhaps

```suggestion
func (m MapStringAny) GetArray(key string) ([]any, error) {
```

... or...

```suggestion
func (m MapStringAny) GetSlice(key string) ([]any, error) {
```

internal/utils/msa/msa.go
```diff
@@ -0,0 +1,120 @@
...
+	return mapGet[[]any](m, key)
+}
+
+// Get a map from a map with string keys
+func (m MapStringAny) GetMap(key string) (map[string]any, error) {
```
zregvart, line 99: This might have better utility

```suggestion
func (m JSONObject) Property(key string) (JSONObject, error) {
```

zregvart (commented):
Mostly nitpickery

internal/utils/msa/msa.go
simonbaird: Or... perhaps we could make a deep clone to prevent that mutation.

internal/utils/msa/msa.go
simonbaird: Done (in upcoming revision).

internal/utils/msa/msa.go
simonbaird: My motivation is succinct and tidy code in the caller. But implementing ToJSONIndent would be more go-ish I guess. I'll do that.

internal/utils/msa/msa.go
simonbaird: Slice is more accurate I think, so I'll go with that.

internal/utils/msa/msa.go
simonbaird: I'll make a separate method for that.

internal/utils/msa/msa.go
simonbaird: I understand we don't like random panics, but I kinda want to keep this. Added some commentary to explain why.

internal/utils/msa/msa.go
simonbaird: Done in latest revision.

internal/utils/msa/msa.go
simonbaird: Done.

internal/utils/msa/msa.go
simonbaird: See latest revision, top commit.

simonbaird:
I'm starting to think this is not needed...

PR #1001, opened 2023-09-08
Upgrade Tekton Pipeline to v0.51.0

zregvart:
Upgrades to latest Tekton Pipeline, switching the API from v1beta1 to v1.

Seems that our Task result snapshotting did not catch the result in failure which could be the consequence of the API upgrade.

Also includes #1000 as this fixes the deprecated API usage found by the linter there.


lcarva:
> Seems that our Task result snapshotting did not catch the result in failure which could be the consequence of the API upgrade.

https://github.com/tektoncd/pipeline/pull/6510 seems to confirm it.

lcarva (approved):

PR #1002, opened 2023-09-08
HACBS-2414 Set IGNORE_REKOR default to 'false'

robnester-rh:
This commit sets the default value for the `IGNORE_REKOR` param to `false`


zregvart:
I see [several scenarios failed](https://github.com/enterprise-contract/ec-cli/actions/runs/6161827601/job/16721934163?pr=1002#step:5:4843), those most likely also need `IGNORE_REKOR=true`

zregvart (approved):

PR #1003, opened 2023-09-11
Run the release workflow only for top of HEAD

zregvart:
When the Checks workflow is run and succeeds for a change that targets the default branch the release workflow will trigger. This includes reruns and doesn't limit to the last commit on the main branch.

This adds a condition so that the release workflow only proceeds if it is triggered for a Checks workflow that ran for the latest commit on the default branch.

Ref. https://issues.redhat.com/browse/HACBS-2585

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -28,6 +28,21 @@ permissions:
   contents: read
 
 jobs:
+  info:
+    name: Info
+    runs-on: ubuntu-latest
+    outputs:
+      head_sha: ${{ steps.head_sha.outputs.head_sha }}
+
+    steps:
+      - name: Git Info
+        id: head_sha
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_COBRA: 1
+        run: |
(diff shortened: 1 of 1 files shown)


lcarva (approved):

lcarva:
We should be able to test this by simply retrying a release workflow. We must be careful to only do so for commits that include this change though.

zregvart:
> We should be able to test this by simply retrying a release workflow. We must be careful to only do so for commits that include this change though.

Just noting that retrying on unchanged `main` should be allowed and this doesn't prevent that, a release triggered by a rerun of the Checks workflow for a older commit should not be allowed, and hopefully this prevents that. I did some testing on my fork, [this one](https://github.com/zregvart/ec-cli/actions/runs/6147607066) was prevented while [this one](https://github.com/zregvart/ec-cli/actions/runs/6147590610) was allowed to continue (and there are many more in there).
One problem I faced was rate limiting, I guess I was testing too frequently, but [in that scenario](https://github.com/zregvart/ec-cli/actions/runs/6147380068/job/16678743304) the `Info` job did not fail, which is odd. I added `GH_COBRA=1` to [force this branch](https://github.com/cli/cli/blob/7e7391bf88b8cf3da0b51b1b9bfc840645d43b32/pkg/cmd/root/root.go#L88) not to happen. That might help...

PR #1016, opened 2023-09-14
Add missing test tags

simonbaird:


simonbaird:
There's a failure which I'm not sure how to fix:

`2023/09/14 20:32:13 failed to parse profiles: open /tmp/tmp.FzbojEwEHG/coverage-acceptance*.out: no such file or directory
exit status 1`

simonbaird:
I removed the extra Makefile changes to see if it would pass...

simonbaird:
:shrug:

acceptance/acceptance_test.go
```diff
@@ -14,6 +14,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+//go:build unit
```
lcarva, line 17: We probably don't want a build tag for the acceptance tests.

simonbaird: :+1:

Makefile
```diff
@@ -77,14 +77,26 @@ clean: ## Delete build output
...
+test-generative:
 # Given the nature of generative tests the test timeout is increased from 500ms
 # to 30s to accommodate many samples being generated and test cases being run.
 	@go test -race -covermode=atomic -coverprofile=coverage-generative.out -timeout 30s -tags=generative ./...
 
+.PHONY: test
+test: generate ## Run all unit tests
+	@$(MAKE) test-unit
+	@$(MAKE) test-integration | grep -v '\[no test files\]'
+	@$(MAKE) test-generative | grep -v '\[no test files\]'
```
zregvart, line 98: Why not?

```suggestion
test: generate test-unit test-integration test-generative
```

Makefile
```diff
@@ -77,14 +77,26 @@ clean: ## Delete build output
 
 ##@ Testing
 
-.PHONY: test
-test: generate ## Run all unit tests
+.PHONY: test-unit
+test-unit:
```
zregvart, line 81: Most likely needs (try with `git clean -dfx`)

```suggestion
test-unit: generate
```

Makefile
simonbaird: See commit message where I answer this question.

Makefile
simonbaird: I was wondering about that. Is generate needed only for test-unit ?

zregvart: Without it on clean checkout there is no code in `/application`...

Makefile
simonbaird: It's because they each run in parallel. There might be another way to make them run sequentially though..

zregvart: Saves us time to run them in parallel, perhaps we could tweak the way results were outputted, though not sure we loose that much on performance...

Makefile
simonbaird: Since generate runs whether it needs to or not, I decide to switch it back to a single task.

lcarva (approved):

zregvart (approved):
Rebase and merge?

lcarva:
Re-opened the PR to trigger the dependency review shared workflow. This is leftover from the initial setup work.

PR #1018, opened 2023-09-14
More efficient json decoding for attestations

simonbaird:
Use json.NewDecoder..Decode.. since it works with the reader directly and hence should be less resource intensive than using io.ReadAll and then json.Marshal.

There were some minor test changes needed to make the tests pass due to the different ways it handles malformed or empty input.


simonbaird:
Suggested by @zregvart in #975 .

zregvart (approved):

PR #1023, opened 2023-09-18
Fail if unsupported Rego rules are loaded

zregvart:
This adds a check when the Rego files are inspected to see if the policy rules return a supported value (string, object, variable or a value from function invocation). This is as fine as we can make this check as there is no type information from OPA.

Ref. https://issues.redhat.com/browse/RHTAPBUGS-736


internal/opa/inspect_test.go
```diff
@@ -148,3 +157,75 @@ func TestInspectDir(t *testing.T) {
...
 	}
 }
+
+func TestCheckRules(t *testing.T) {
+	cases := []struct {
+		name string
+		rego string
+		err  string
+	}{
+		{
```
lcarva, line 167: Let's add something like this for the missing test coverage:

```rego
deny := "value" {
	true
}
```

lcarva (approved):

internal/opa/inspect_test.go
zregvart: Good catch!

lcarva (approved):

lcarva:
Coverage report doesn't seem accurate. Let's merge it.

PR #1026, opened 2023-09-19
Process include/exclude from policy source

lcarva:
https://issues.redhat.com/browse/HACBS-2428


go.mod
```diff
@@ -306,3 +306,6 @@ require (
 )
 
 replace github.com/enterprise-contract/ec-cli/application => ./application
+
+// TODO: Remove this once this is merged: https://github.com/enterprise-contract/enterprise-contract-controller/pull/185
```
lcarva, line 310: Draft until this is addressed.

internal/evaluator/conftest_evaluator.go
```diff
@@ -199,13 +202,13 @@ func (r conftestRunner) Run(ctx context.Context, fileList []string) (result []Ou
 
 // NewConftestEvaluator returns initialized conftestEvaluator implementing
 // Evaluator interface
-func NewConftestEvaluator(ctx context.Context, policySources []source.PolicySource, p policy.Policy) (Evaluator, error) {
-	return NewConftestEvaluatorWithNamespace(ctx, policySources, p, nil)
+func NewConftestEvaluator(ctx context.Context, policySources []source.PolicySource, p policy.Policy, sc *ecc.SourceConfig) (Evaluator, error) {
```
lcarva, line 205: I think this is ok for now, but I want us to reconsider this function signature. It seems redundant to pass in both the `Policy` and the `PolicySource` since the `PolicySource` is within the `Policy`. This is done, of course, so the evaluator knows which source group to use. I don't think the abstraction is quite right here.

We may want to consider this function only taking these parameters:
* `ctx context.Context` obviously
* `src *ecc.Source`. This should include the `SourceConfig` and enough information to build `[]source.PolicySource`.
* `effectiveTime time.Time`. This function doesn't need a whole `Policy`, just the effective time.

There's a wrinkle with this though. We still need something to combine the config from `ecc.Source` with the "global" config `Policy.Spec().Configuration`. That might be better off done prior to calling `NewConftestEvaluator` though which would make this a non-issue.

Anyways, this is somewhat subjective so I didn't want to include in this PR. If there's some agreement here, I will make a follow up PR to do so.

zregvart: :+1: to refactor the signature here, perhaps something as simple as this could work?

```
// NewConftestEvaluator create a Evaluator implemented via Conftest to evaluate the n-th policy source
func NewConftestEvaluator(ctx context.Context, p policy.Policy, n int) (Evaluator, error) {
```

zregvart: Or even:

```
// NewConftestEvaluators create a slice of Evaluator implementations via Conftest
func NewConftestEvaluators(ctx context.Context, p policy.Policy) ([]Evaluator, error) {
```

zregvart (approved):

PR #1033, opened 2023-09-21
Add an asdf config to specify correct Go versions

mbestavros:
Might be helpful for those of us that use `asdf`.


.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.20.8
```
lcarva, line 1: What will keep this up to date?

mbestavros: Good point. I suppose it's something we'd have to update whenever we bump `go` versions? But that can be easily forgotten...

lcarva: Given that it includes a specific version, and not just `1.20`, it would have to be updated more frequently.

Ha! https://docs.renovatebot.com/modules/manager/asdf/

We may need to tweak it to also suport [golang version](https://docs.renovatebot.com/modules/datasource/golang-version/)

lcarva: I think it is already supported: https://github.com/renovatebot/renovate/blob/3d7cf23529371a7960ee217c9218f0d92978f84c/lib/modules/manager/asdf/upgradeable-tooling.ts#L184-L191

mbestavros: How would we update the renovate config? I don't see a `renovate.json` file or anything of the like.

lcarva: Here's an example: https://github.com/redhat-appstudio/build-definitions/blob/main/renovate.json

I just enabled it on this repo.

zregvart: We could also do it by hand, we don't upgrade golang that often... Though, any automatic update should also make sure that the versions in all `go.mod` files are in sync with the `.tool-versions` file...

lcarva: @zregvart , I was thinking renovate should maintain the PATCH portion of the version. MAJOR and MINOR versions would still be maintained by hand.

robnester-rh (approved):

.tool-versions
lcarva: We can ignore this for now.

lcarva (approved):

PR #1045, opened 2023-09-28
Lookup the latest digest of for a image tag

zregvart:
In `ec track bundle`, given a set of images to track either contained in `--input` or provided via `--bundle` parameter, if the `--freshen` flag is provided an additional lookup is done to fetch the latest digest of the image for the provided tag.

Ref. https://issues.redhat.com/browse/HACBS-2594


lcarva (approved):
Outstanding!

PR #1057, opened 2023-10-02
Create auto-merge.yaml

zregvart:
Ref. https://issues.redhat.com/browse/HACBS-2470

### DIFF
--- a/.github/workflows/auto-merge.yaml
+++ b/.github/workflows/auto-merge.yaml
@@ -1,0 +1,29 @@
+# Copyright The Enterprise Contract Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+---
+name: 'Auto merge'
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-merge.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+---
+name: 'Auto merge'
+'on':
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  auto-merge:
+    uses: enterprise-contract/github-workflows/.github/workflows/auto-merge.yaml@main
```
lcarva, line 25: Why not use the `.github` repo since that already has shared workflows?

lcarva: hmm we should probably move the other workflows there then. Having shared workflows in two separate repos is confusing.

zregvart: The [error](https://github.com/zregvart-test/testbed/actions/runs/6378405326) I was getting for reference https://github.com/zregvart-test/.github/blob/main/workflows/auto-merge.yaml

Screenshot 2023-10-02 at 16-31-40 Auto merge · zregvart-test_testbed@b64770c

lcarva: Could it be because it's across orgs/namespaces?

zregvart:
Okay, so this is not merged, why is it showing up [here](https://github.com/enterprise-contract/ec-cli/actions/workflows/auto-merge.yaml)?

lcarva:
> Okay, so this is not merged, why is it showing up [here](https://github.com/enterprise-contract/ec-cli/actions/workflows/auto-merge.yaml)?

Because it's a branch on this repo, instead of a fork.

.github/workflows/auto-merge.yaml
lcarva: oy... let's file a story to move the other shared workflows to the new repo then?

lcarva (approved):

Issue #1065, opened 2023-10-03
Update to OPA 0.57.0


zregvart:
Create a new PR, dependabot keeps closing #1048

PR #1066, opened 2023-10-03
Update to OPA v0.57.0

zregvart:
resolves: gh-1065


acceptance/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/enterprise-contract/ec-cli/acceptance
 
-go 1.20
+go 1.21.0
+
+toolchain go1.21.1
```
lcarva, line 5: Why do we need this? Why does it have a different value than the `go` line above?

zregvart: OPA requires go 1.21

lcarva: That's not what I meant. Why do we have different values for `go` and `toolchain` ?

zregvart: There is explanation of this on [the go blog](https://go.dev/blog/toolchain#toolchain-management)

lcarva: Thanks. That blog is not super clear. I'm trying to understand. It seems to me that the value `1.21.0` could be used in both cases? Is there are reason we want the `.1` patch for toolchain?

zregvart: Just reflects what I had on the command line at the time

Issue #1010, opened 2023-09-13
Restructure `cobra` command hierarchy to allow for more thorough unit testing


mbestavros:
Currently, `ec-cli`'s unit tests rely on creating a mock subcommand, not connected to any higher-level root commands, to run unit tests. This has caused spurious issues occasionally when writing unit tests.

A better approach, outlined [here](https://g14a.dev/posts/Testing-Cobra-Subcommands/), seems to be structuring `cobra`'s command hierarchy to allow the full set of commands to be constructed in unit tests. `ec-cli` cannot currently do this, however, since all commands are defined inline.

Abstracting command construction out into a separate function could improve modularity and make better unit testing possible.

A/C:

- `ec-cli`'s Cobra commands can be constructed independently.

PR #1069, opened 2023-10-04
Refactor tests to utilize full Cobra command backend

### LINKED ISSUE #1011: Run `ec-cli`'s unit tests using the full Cobra command hierarchy
Relies on #1010. Once that is complete, unit tests can be restructured to take advantage of the new modularity by constructing a full command hierarchy.

A/C:

- Unit tests consist of a full command hierarchy, instead of mock subcommands.

mbestavros:
This PR refactors `ec-cli`'s tests to utilize the modular Cobra command hierarchy introduced in #1010. Resolves #1011.

### DIFF
--- a/cmd/inspect/inspect_policy_test.go
+++ b/cmd/inspect/inspect_policy_test.go
@@ -23,10 +23,12 @@ import (
 	"testing"
 
 	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"golang.org/x/net/context"
 
+	"github.com/enterprise-contract/ec-cli/cmd/root"
 	"github.com/enterprise-contract/ec-cli/internal/policy/source"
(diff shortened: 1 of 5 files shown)


cmd/inspect/inspect_policy_test.go
```diff
@@ -157,3 +168,11 @@ func TestSourcesAndPolicyCantBeBothProvided(t *testing.T) {
 
 	assert.Contains(t, buffy.String(), "Usage:")
```
lcarva, line 169: So we need to either remove this check (probably makes the most sense), or assert the inverse. This is because the command chain is being properly constructed, and the root command specifies: `SilenceUsage: true`

mbestavros:
This is now ready. Tests do pass as is.

One potential improvement I could make would be to store the branch commands for each test (eg. `validate image`) in a global variable, then append specific arguments to that per test. Thoughts?

zregvart (approved):

zregvart:
> Thoughts?

Merge as is, if we find the need for global variables we can refactor...

PR #1073, opened 2023-10-04
Relax the SLSA Provenance schema validation

lcarva:
Chains is not populating the `invocation.configSource.uri` attribute correctly, due to https://github.com/tektoncd/chains/issues/934

This change relaxes the schema validation until the issue in Chains is fixed and widely adopted.

Ref: RHTAPBUGS-857


lcarva:
codecov you're drunk.

robnester-rh (approved):
LGTM

PR #1074, opened 2023-10-05
Note how to exclude a violation in the description

zregvart:
This adds the text:

```
To exclude this rule add \"<code>[:term]\" to the `exclude` section of
the policy configuration
```

To the description field in the output when `--info` is provided to `validate image` command.

resolves: HACBS-2665

### DIFF
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -103,8 +103,29 @@ func trim(results *[]Outcome) {
 		return trimmed
 	}
 
+	addNote := func(results []Result) []Result {
+		for i := range results {
+			var description, code string
+			var ok bool
+			if description, ok = results[i].Metadata[metadataDescription].(string); !ok {
+				continue
+			}
+
+			if code, ok = results[i].Metadata[metadataCode].(string); !ok {
+				continue
+			}
+
+			if term, ok := results[i].Metadata[metadataTerm]; ok {
+				code = fmt.Sprintf("%s:%s", code, term)
+			}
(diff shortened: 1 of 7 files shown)


internal/evaluator/conftest_evaluator.go
```diff
@@ -103,8 +103,29 @@ func trim(results *[]Outcome) {
...
+				continue
+			}
+
+			if term, ok := results[i].Metadata[metadataTerm]; ok {
+				code = fmt.Sprintf("%s:%s", code, term)
+			}
+			results[i].Metadata[metadataDescription] = fmt.Sprintf("%s. To exclude this rule add %q to the `exclude` section of the policy configuration", strings.TrimSuffix(description, "."), code)
+		}
+
+		return results
```
lcarva, line 124: No need for this function to return a slice of results since it is modifying the one that is given? Alternatively, we could make it so it creates a copy (which IMO is always safer).

zregvart: Can't compose it below otherwise

internal/evaluator/conftest_evaluator.go
```diff
@@ -103,8 +103,29 @@ func trim(results *[]Outcome) {
...
+	}
+
 	for i, checks := range *results {
-		(*results)[i].Failures = trimOutput(checks.Failures)
+		(*results)[i].Failures = addNote(trimOutput(checks.Failures))
 		(*results)[i].Warnings = trimOutput(checks.Warnings)
```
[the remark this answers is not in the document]
zregvart: Decided it wasn't for the best, skipped warning becomes a skipped failure unintentionally

PR #1078, opened 2023-10-06
Remove internal error package.

robnester-rh:
This commit removes the internal error package which served as the basis of specific error codes for return values.

Resolves: HACBS-2652


internal/attestation/slsa_provenance_02_test.go
```diff
@@ -131,7 +131,7 @@ func (l mockSignature) MediaType() (types.MediaType, error) {
 
 func TestSLSAProvenanceFromSignatureNilSignature(t *testing.T) {
 	sp, err := SLSAProvenanceFromSignature(nil)
-	assert.True(t, AT001.Alike(err), "Expecting `%v` to be alike: `%v`", err, AT001)
+	assert.True(t, strings.Contains(err.Error(), "no attestation found"), "Expecting `%v` to be alike: `%v`", err, "no attestation found")
```
lcarva, line 134: Could use `assert.ErrorContains` here.

lcarva (approved):

lcarva:
codecov failure seems bogus.

robnester-rh:
> codecov failure seems bogus.

agreed

PR #1085, opened 2023-10-10
Use golang 1.21.2 consistently

zregvart:
Makes sure all `go.mod` files use the same golang/go toolbox version and upgrades to the latest 1.21.2.


lcarva (approved):
I guess a different version for `toolchain` wasn't needed after all

PR #1106, opened 2023-10-18
Add newline at end of "ec validate image" outputs

### LINKED ISSUE #1059: Multiple outputs to stdout appear concatenated
If `ec validate --output summary --output json` is invoked the two JSON outputs are outputted on the same line.
We should detect that the output goes to stdout and add a newline between the two outputs to stdout.

mbestavros:
Fixes #1059.


lcarva (commented):
lgtm

I don't think we need to add a new test case for this. Let's just update the current snapshots.

zregvart (approved):

internal/applicationsnapshot/report.go
```diff
@@ -153,7 +153,7 @@ func (r Report) WriteAll(targets []string, p format.TargetParser) (allErrors err
 
 		if data, err := r.toFormat(target.Format); err != nil {
 			allErrors = multierror.Append(allErrors, err)
-		} else if _, err := target.Write(data); err != nil {
+		} else if _, err := target.Write(append(data, "\n"...)); err != nil {
```
simonbaird, line 156: Nitpick: I think I would have made this "smart", i.e. inject a line break only if the output doesn't already end with a line break.

Also I don't understand the `...` syntax here. Is it because data is `[]byte`?

mbestavros:
Acceptance tests now pass, and I've made a tweak so we don't insert as many newlines. @simonbaird look good to you?

simonbaird:
Nice, thanks. Just looking at the snapshot diff I like it a lot more than the previous version.

Issue #1070, opened 2023-10-04
Introduce ec.fetch_oci_blob rego function


lcarva:
# Description

We want the ability to fetch blobs from an OCI registry from policy rules. Let's introduce a new custom built-in rego function to enable this, see example [commit](https://github.com/lcarva/ec-cli/commit/667512aa689779dea4b7f88a3aea77ddf7b49d0a) and the [docs](https://www.openpolicyagent.org/docs/latest/extensions/#custom-built-in-functions-in-go).

Considerations:

* [Memoization](https://en.wikipedia.org/wiki/Memoization). Using the function on the same OCI reference should not cause the SBOM to be fetched more than once from the OCI registry. The rego may provide some utilities for this.
* Error handling.  If the golang implementation of the function returns an error, this causes EC to exit with fatal failure, which is bad. Consider adding a supporting "is_valid" function to ensure the reference is a valid OCI reference by digest, and return an empty string if a runtime error occurs, for example.
* Data limit. Use something like [io.LimitReader](https://pkg.go.dev/io#LimitReader) to set some boundaries on how much data EC is fetching. (10 MB should be enough for SBOMs to start with.) Be sure to validate the fetched content has the expected digest.

# Acceptance Criteria

* A new rego function can be used to fetch a blob from an OCI registry.

zregvart:
What limits are we imposing in the first implementation? Single-layer images only? No manifest/config access?

lcarva:
Good question. The scope here is a "blob", which usually means a single layer from an image manifest. However, the reference being fetched should be to the blob, and not to the wrapper image manifest. In fact, it shouldn't matter if the blob exists on its own in the registry, or if it's part of an image manifest.

I'll share what I wrote somewhere else:

> The nice thing about blobs is that they have the same digest as the underlying content. However, blobs are garbage collected quite aggressively by most registries, including quay.io. The SBOM may not be available in an hour, for example. To get around this, we could wrap the blob in a manifest so we can tag it. This is easily done with [ORAS](https://oras.land/)’s “[push](https://oras.land/docs/commands/oras_push/)” command, instead of the "[blob push](https://oras.land/docs/commands/oras_blob_push)" command.

Notice that in either case, a reference like this is what is provided to the ec function: `<registry>/<repo>@sha256:<digest>`, where `<digest>` is the sha256 digest of the SBOM file. This is important.

PR #1107, opened 2023-10-18
Add ec.fetch_oci_blob rego function

lcarva:
~It implements #1070, but it doesn't resolve it just yet. As a followup PR, I'll add acceptance tests to complete this functionality.~

UPDATE: Acceptance tests are now included.

### DIFF
--- a/internal/evaluator/rego.go
+++ b/internal/evaluator/rego.go
@@ -1,0 +1,123 @@
+// Copyright The Enterprise Contract Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// IMPORTANT: The rego functions in this file never return an error. Instead, they return no value
+// when an error is encountered. If they did return an error, opa would exit abruptly and it would
+// not produce a report of which policy rules succeeded/failed.
+
+package evaluator
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"io"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/types"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/enterprise-contract/ec-cli/internal/fetchers/oci"
+)
+
+const ociBlobName = "ec.oci.blob"
+
+func registerOCIBlob() {
+	decl := rego.Function{
+		Name: ociBlobName,
+		Decl: types.NewFunction(
+			types.Args(
+				types.Named("ref", types.S).Description("OCI blob reference"),
+			),
+			types.Named("blob", types.S).Description("the OCI blob"),
+		),
+		// As per the documentation, enable memoization to ensure function evaluation is
+		// deterministic. But also mark it as non-deterministic because it does rely on external
+		// entities, i.e. OCI registry. https://www.openpolicyagent.org/docs/latest/extensions/
+		Memoize:          true,
+		Nondeterministic: true,
+	}
+
+	rego.RegisterBuiltin1(&decl, ociBlob)
+}
+
+const maxBytes = 10 * 1024 * 1024 // 10 MB
+
+func ociBlob(bctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
+	uri, ok := a.Value.(ast.String)
+	if !ok {
+		return nil, nil
+	}
+
+	ref, err := name.NewDigest(string(uri))
+	if err != nil {
+		log.Errorf("%s new digest: %s", ociBlobName, err)
+		return nil, nil
+	}
+
+	opts := []remote.Option{
+		remote.WithTransport(remote.DefaultTransport),
+		remote.WithContext(bctx.Context),
+		remote.WithAuthFromKeychain(authn.DefaultKeychain),
+	}
+
+	rawLayer, err := oci.NewClient(bctx.Context).Layer(ref, opts...)
+	if err != nil {
+		log.Errorf("%s fetch layer: %s", ociBlobName, err)
+		return nil, nil
+	}
+
+	layer, err := rawLayer.Uncompressed()
+	if err != nil {
+		log.Errorf("%s layer uncompressed: %s", ociBlobName, err)
+		return nil, nil
+	}
+	defer layer.Close()
+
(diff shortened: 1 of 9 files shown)


internal/evaluator/rego.go
```diff
@@ -0,0 +1,134 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// IMPORTANT: The rego functions in this file never return an error. Instead, they return no value
```
lcarva, line 17: Not sure if this is visible enough.

zregvart: I think it is...

internal/evaluator/rego.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	layer, err := rawLayer.Uncompressed()
+	if err != nil {
+		log.Errorf("fetchOCIBlob layer uncompressed: %s", err)
+		return nil, nil
+	}
+	defer layer.Close()
+
+	// Other algorithms are technically supported, e.g. sha512. However, support for those is not
```
zregvart, line 107: Perhaps prefix with `TODO`

internal/evaluator/rego.go
```diff
@@ -0,0 +1,134 @@
...
+const remoteLayerKey key = "ec.evaluator.remote.layer"
+
+type remoteLayerFn func(ref name.Digest, options ...remote.Option) (gcr.Layer, error)
+
+func remoteLayer(ctx context.Context, ref name.Digest, options ...remote.Option) (gcr.Layer, error) {
+	fn, ok := ctx.Value(remoteLayerKey).(remoteLayerFn)
+	if ok && fn != nil {
+		return fn(ref, options...)
+	}
+	return remote.Layer(ref, options...)
```
zregvart, line 51: Could we use cache here? Like what we do in `internal/fetchers/oci/client.go`?

internal/evaluator/rego.go
```diff
@@ -0,0 +1,134 @@
...
+	reader := io.TeeReader(io.LimitReader(layer, maxBytes), hasher)
+
+	var blob bytes.Buffer
+	if _, err := io.Copy(&blob, reader); err != nil {
+		log.Errorf("fetchOCIBlob copy buffer: %s", err)
+		return nil, nil
+	}
+
+	sum := fmt.Sprintf("sha256:%x", hasher.Sum(nil))
+	if sum != ref.DigestStr() {
```
zregvart, line 122: Is this because LimitReader might truncate the stream? If not so, think the [fetcher](https://github.com/google/go-containerregistry/blob/a54d64203cffcbf94146e04069aae4a97f228ee2/pkg/v1/remote/fetcher.go#L276) already checks the digest.

zregvart (commented):
I was wondering about the naming, and if `ec.oci.blob` could be a better name, or even if we want to push this upstream to OPA and have it `opa.blob`?
Some nitpicks inside, other than that :+1:

internal/evaluator/rego.go
lcarva: That's exactly why. I wish LimitReader would return whether or not content was truncated. If it did, this wouldn't be needed.

internal/evaluator/rego.go
lcarva: +1 done

internal/evaluator/rego.go
lcarva: Added a comment to make it more obvious.

lcarva:
> I was wondering about the naming, and if `ec.oci.blob` could be a better name, or even if we want to push this upstream to OPA and have it `opa.blob`? 

Updated to go with `ec.oci.blob`.

I think you probably meant `oci.blob` on your second suggestion. If so, that sounds like a good upstream name if/when this function is part of opa itself. For now, since it is implemented in EC, let's make it super obvious that it is related to EC by having the `ec.` prefix.

lcarva:
Caching is not working. I need to rethink the approach. Put it in draft for now.

lcarva:
I'm gonna hold off on the cache approach for this PR due to https://github.com/google/go-containerregistry/issues/1821.

lcarva:
FYI, code coverage is failing to be uploaded. It is a known issue and it happens from time to time...

internal/evaluator/rego.go
```diff
@@ -0,0 +1,120 @@
...
+		Name: ociBlobName,
+		Decl: types.NewFunction(
+			types.Args(
+				types.Named("ref", types.S).Description("OCI blob reference"),
+			),
+			types.Named("blob", types.S).Description("the OCI blob"),
+		),
+		Memoize:          true,
+		Nondeterministic: true,
+	}
```
simonbaird, line 53: If it's safe to memoize, could we consider it safe to call it deterministic as well? (I didn't look at the docs for these options, just wondering about it...) Consider adding a comment to explain if it's not sufficiently obvious.

internal/evaluator/rego.go
```diff
@@ -0,0 +1,120 @@
...
+	reader := io.TeeReader(io.LimitReader(layer, maxBytes), hasher)
+
+	var blob bytes.Buffer
+	if _, err := io.Copy(&blob, reader); err != nil {
+		log.Errorf("%s copy buffer: %s", ociBlobName, err)
+		return nil, nil
+	}
+
+	sum := fmt.Sprintf("sha256:%x", hasher.Sum(nil))
+	// io.LimitReader truncates the layer if it exceeds its limits. This ensures unexpected behavior
```
simonbaird, line 106: I might be misunderstanding it, but...

```suggestion
	// io.LimitReader truncates the layer if it exceeds its limits. This avoids unexpected behavior
```

internal/evaluator/rego.go
lcarva: It is `Nondeterministic` because it makes a network call. This is what the [docs](https://www.openpolicyagent.org/docs/latest/extensions/) recommend:

> The declaration also sets rego.Function#Memoize to true to enable memoization across multiple calls in the same query. If your built-in function performs I/O, you should enable memoization as it ensures function evaluation is deterministic.
> 
> Since this built-in could have non-deterministic results, depending on network conditions, the declaration also sets rego.Function#Nondeterministic to true. This provides basic safety information to the runtime, so that the function isn’t accidentally run during bundle builds or partial evaluation. If your builtin can have non-deterministic results, you should mark it appropriately to avoid surprises.

I'll add a comment explaining as it is likely not obvious to others.

simonbaird (approved):

internal/evaluator/rego.go
lcarva: I re-wrote this sentence because it was just awful.

internal/evaluator/rego.go
```diff
@@ -0,0 +1,124 @@
...
+func registerOCIBlob() {
+	decl := rego.Function{
+		Name: ociBlobName,
+		Decl: types.NewFunction(
+			types.Args(
+				types.Named("ref", types.S).Description("OCI blob reference"),
+			),
+			types.Named("blob", types.S).Description("the OCI blob"),
+		),
+		// As per the docs documentation, enable memoization to ensure function evaluation
```
simonbaird, line 51:
```suggestion
		// As per the documentation, enable memoization to ensure function evaluation
```

internal/evaluator/rego.go
simonbaird: Awesome.

internal/evaluator/rego.go
lcarva: :facepalm: 

Thanks!

zregvart (approved):

PR #1108, opened 2023-10-19
Fix tag ref in tkn bundle

lcarva:
At some point, we changed the variable in the Makefile from TASK to TASKS to support more than one task definition. We forgot to update the GitHub workflow to use the new variable name.


simonbaird (approved):

zregvart (commented):
Thanks!

Issue #1111, opened 2023-10-23
Patch version is always set to 1


zregvart:
Seems that all released versions have the patched version set to 1 instead of the number of commits on the default branch.

zregvart:
Verified on the latest release:

<pre><code>$ podman run quay.io/enterprise-contract/ec-cli:snapshot version
Trying to pull quay.io/enterprise-contract/ec-cli:snapshot...
Getting image source signatures
Copying blob 6ae19ef0c28b skipped: already exists  
Copying blob 130b059cdb64 done   | 
Copying blob caff82f63312 done   | 
Copying blob 14b65f7f6b94 done   | 
Copying blob 49108d4ea228 done   | 
Copying config 848c1e3282 done   | 
Writing manifest to image destination
Version            <b>v0.1.2161-d6b54a8</b>
Source ID          d6b54a8280fec5b8b9909c4a9cfc1c1e69d8f499
Change date        2023-10-23 13:24:17 +0000 UTC (1 hour ago)
ECC                v0.0.0-20230928144500-ee994416688e
OPA                v0.57.1
Conftest           v0.46.0
Cosign             N/A
Sigstore           v1.7.4
Rekor              v1.2.2
Tekton Pipeline    v0.51.0
Kubernetes Client  v0.28.2
</code></pre>

PR #1112, opened 2023-10-23
Set fetch depth to 0 for the release workflow

zregvart:
Given a set of versions it is difficult to compare them, for that reason we introduced the patch version having the number of commits on the release branch. The fetch depth set to 1 by default sets the patch version to `1`.

This sets the fetch depth to `0`, i.e. all commits, so that the command `git rev-list --count HEAD` returns the number of commits on the release branch instead of `1`.

Resolves: #1111


robnester-rh (approved):
LGTM

PR #1123, opened 2023-10-23
Add application.go

lcarva:
This commit adds an application.go file so dependencies can be properly tracked in the corresponding go.mod file.


zregvart (approved):
Pragmatic, I like it. We are taking on a tiny bit of dept in making sure we update `application.go` when dependencies of application-api change.

Issue #1052, opened 2023-09-29
Handle effective* in include/exclude


lcarva:
With https://github.com/enterprise-contract/enterprise-contract-controller/issues/191, the include/exclude list may contain objects instead of just strings. See that issue for more details.

This is about teaching the CLI to handle those correctly.

Consider the following example:
```yaml
sources:
  - config:
      exclude:
        - step_image_registries
        - name: tasks
          effectiveUntil: "2024-01-01T00:00:00Z"
        - name: tests
          effectiveOn: "2023-10-01T00:00:00Z"
      include:
        - '@redhat'
    data:
      ...
    policy:
      ...
```

`step_image_registries` is always excluded. `tasks` is only excluded until January 1st, 2024. `tests` is excluded starting on January 1st, 2023.

Eventually, we may want to emit a warning in certain cases. For example, if a certain violation that is being excluded today will be included in a few days (either via exclude entry with effectiveUntil, or include entry with effectiveOn). Let's keep that out of scope for this issue.

PR #1126, opened 2023-10-25
Support volatile include/exclude configuration

zregvart:
With this the volatile configuration affects the rule inclusion and
exclusion.

resolves: #1052

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	cuelang.org/go v0.6.0
 	github.com/MakeNowJust/heredoc v1.0.0
 	github.com/enterprise-contract/ec-cli/application v0.0.0-00010101000000-000000000000
-	github.com/enterprise-contract/enterprise-contract-controller/api v0.0.0-20230928144500-ee994416688e
+	github.com/enterprise-contract/enterprise-contract-controller/api v0.0.0-20231027095011-f06fe20fb615
 	github.com/evanphx/json-patch v5.7.0+incompatible
 	github.com/gkampitakis/go-snaps v0.4.11
 	github.com/go-logr/logr v1.2.4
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -306,3 +306,5 @@ require (
 )
 
 replace github.com/enterprise-contract/ec-cli/application => ./application
+
+replace github.com/enterprise-contract/enterprise-contract-controller/api => github.com/zregvart/enterprise-contract-controller/api v0.0.0-20231025114412-77586c67f7ca
```
lcarva, line 310: Adding a comment here for tracking purposes. We should remove this once https://github.com/enterprise-contract/enterprise-contract-controller/pull/212 is merged. Ideally, before merging this PR.

internal/evaluator/conftest_evaluator.go
```diff
@@ -872,15 +872,46 @@ func strictCapabilities(ctx context.Context) (string, error) {
...
+				if err != nil {
+					log.Warnf("unable to parse time for criteria %q, was given %q: %v", c.Value, c.EffectiveOn, err)
+					from = at.Add(-1)
+				}
+				until, err := time.Parse(time.RFC3339, c.EffectiveUntil)
+				if err != nil {
+					log.Warnf("unable to parse time for criteria %q, was given %q: %v", c.Value, c.EffectiveUntil, err)
+					until = at.Add(1)
+				}
+				if from.Before(at) && until.After(at) {
```
lcarva, line 902: I wonder if this should be `!from.After(at) && !until.Before(at)` in order to create an inclusive range which seems less surprising, I think.

It would also allow both `from` and `until` to have the default value of `at`.

go.mod
simonbaird, line 310: Maybe a comment to explain why and when it could be removed.

simonbaird: I see it's in the commit message, so you can disregard the above.

simonbaird: Some might say the commit message is enough, but I'd probably add a comment here too.

simonbaird (approved):

internal/evaluator/conftest_evaluator.go
zregvart: Yeah, I think I want to use Compare instead, makes things more explicit

lcarva (approved):

PR #1141, opened 2023-11-02
This adds code to generate a VSA

joejstuart:
This adds code to generate a VSA under the attestations package.

https://issues.redhat.com/browse/EC-189


internal/attestation/vsa.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	var slsaVersion string
+	var digest map[string]string
+	var subject []in_toto.Subject
+	for _, sp := range attestations {
+		slsaVersion = sp.PredicateType()
+		digest = sp.Digest()
+		subject = sp.Subject()
```
lcarva, line 97: The last attestation wins? That doesn't sound right.

internal/attestation/vsa.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return ProvenanceStatementVSA{
+		StatementHeader: in_toto.StatementHeader{
+			Type:          StatmentVSA,
+			PredicateType: PredicateVSAProvenance,
```
lcarva, line 103: So we're claiming that this is a predicate of type `https://slsa.dev/verification_summary/v1`. However, there are various attributes here that are not part of that schema, e.g. policies, verifiedRules, verifiedCollections, etc. We would be doing a disservice to our users by taking this approach.

To me that's a clear sign that we should be producing something else...

internal/attestation/vsa.go
joejstuart: it sure doesn't! That's an oversight. Will fix.

internal/attestation/vsa.go
joejstuart: I've changed it to `"https://enterprisecontract.dev//verification_summary/v1"`

internal/attestation/vsa.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/in-toto/in-toto-golang/in_toto"
+
+	"github.com/enterprise-contract/ec-cli/internal/evaluator"
+	"github.com/enterprise-contract/ec-cli/internal/policy"
+	"github.com/enterprise-contract/ec-cli/internal/policy/source"
+)
+
+const (
+	// Make it visible elsewhere
+	PredicateVSAProvenance = "https://enterprisecontract.dev//verification_summary/v1"
```
zregvart, line 31: I don't think we should deviate from the SLSA VSA spec, we'll end up being locked in our own ecosystem

internal/attestation/vsa.go
```diff
@@ -0,0 +1,113 @@
...
+	Verifier            map[string]string   `json:"verifier,omitempty"`
+	TimeVerified        string              `json:"timeVerified,omitempty"`
+	Policies            []policySource      `json:"policies,omitempty"`
+	InputAttestations   []attestationSource `json:"intputAttestations,omitempty"`
+	VerificationResult  string              `json:"verificationResult,omitempty"`
+	VerifiedRules       []string            `json:"verifiedRules,omitempty"`
+	VerifiedCollections []string            `json:"verfiedCollection,omitempty"`
+}
+
+func VsaFromImageValidation(verifiedTime string, results []evaluator.Outcome, policies []source.PolicySource, policy policy.Policy, attestations []Attestation) ProvenanceStatementVSA {
```
zregvart, line 59: For me it makes more sense for this to be in `internal/appicationsnapshot` and operated on Report there.

Having both policy.Policy and source.PolicySource in the function signature seems redundant

internal/attestation/vsa.go
joejstuart: My thought is we can produce the SLSA spec and our own, since the results we produce don't conform exactly to what the slsa spec calls for. For instance, we allow for an array of policies where as the slsa spec only allows for one.

zregvart: Let's try to influence the spec first before inventing our own. What we might consider instead is to use the value passed in via `--policy` argument, which is a single string and it's digest.

lcarva: IMO, the SLSA VSA spec as is today is not sufficient for our needs. We may want to produce one, but that's mostly for compatibility with other tools that do expect a SLSA VSA. My understanding is that this is not the requirement as per EC-189, but I could be wrong. For EC-189, I believe the ask was to produce a very EC specific predicate, _inspired_ by SLSA's VSA.

I don't know if it makes sense to try to influence the SLSA VSA to meet our needs. EC handles policies, to meet business requirements, beyond what SLSA provides, e.g. asserting SLSA Level 3 is not sufficient for the use cases where EC is currently used. But it's worth having that conversation with the SLSA community. I just wouldn't block any efforts on that. It's perfectly fine to introduce an experimental predicate type.

A minor commentary is where the code is placed. `internal/attestation` is meant for parsing input predicates, and here we are emitting a predicate. This might be better suited as just another `--output` format.

internal/attestation/vsa.go
joejstuart: yes, makes much more sense operating on report. Just moved it. Thanks.

internal/attestation/vsa.go
joejstuart: Yep, made more sense under `internal/applicationsnapshot` . thanks!

simonbaird (approved):
Lgtm.

Issue #1071, opened 2023-10-04
Figure out how to run unit tests in ec-policies that use custom function


lcarva:
# Description

The function introduced in https://github.com/enterprise-contract/ec-cli/issues/1070 means that a policy rule could write something like this:

``` 
ec.fetch_oci_bundle(ref) 
```

Plain OPA doesn't know anything about that function. If a policy rule in ec-policies wants to use that function, this error will occur in the tests:

``` 
rego_unsafe_var_error: var ec is unsafe 
```

If we try to mock that in the tests by using the `with` keyword:

``` 
rego_type_error: with keyword target must reference existing input, data, or a function
```

One option could be to wrap the opa command in ec-cli.

This also raises the question of how we can mock those functions. We don't want to actually hit a registry during the unit tests, for example.

# Acceptance Criteria

- ec-provided custom rego functions can be used in ec-policies without
  breaking tests.

zregvart:
Do we need `ec policy test` command?

lcarva:
Maybe? That might be nicer than just wrapping the whole opa command. It should be compatible with `opa test` though. So wrapping may still be involved here.

zregvart:
I've asked OPA maintainers about bringing back shared library support for loading plugins into OPA. This is not something that will be considered in OPA's future.

Based on this I think the best option would be to add support for running tests into ec, most likely based on [this pattern](https://www.openpolicyagent.org/docs/latest/extensions/#adding-built-in-functions-to-the-opa-runtime).

Perhaps we can have `ec opa` command that would behave exactly the same as `opa`?

PR #1145, opened 2023-11-06
Add `ec opa` subcommand to run embedded OPA

zregvart:
This includes the `opa` CLI as a subcommand of `ec` CLI. Any registered builtins from the `internal/evaluator` package should be available.

resolves: #1071


lcarva (approved):
:taco:

PR #1149, opened 2023-11-07
Use upstream of application-api directly

zregvart:
We used to copy/vendor application-api to go around dependency alignment issues. Now that has been resolved upstream and we can use the upstream dependency directly.
This greatly simplifies the build and consumption of ec-cli as a dependency.

reference: EC-241


zregvart:
Merged to unblock https://github.com/enterprise-contract/ec-policies/pull/792, feel free to review after the fact, happy to address any issues found.

lcarva:
:tada:

PR #1152, opened 2023-11-07
Update snapshots to reflect removed collections

lcarva:
The slsa1 and slsa2 collection have been removed from ec-policies. This commit updates the snapshots in the acceptance tests to no longer expecte them.


zregvart (approved):
Thanks!

Issue #1113, opened 2023-10-23
Per source group configuration might not be included in the reports


zregvart:
Seems probable based on the [failure of the e2e tests](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_infra-deployments/2602/pull-ci-redhat-appstudio-infra-deployments-main-appstudio-e2e-tests/1716437031288049664) in EC update.

lcarva:
The problem here is that the e2e-tests are using a version of the enterprise-contract-controller which does not have support for the per source group config.

Since this issue was reported, the test case was removed from the e2e tests under the premise that we already have sufficient coverage for it. Regardless, I'm gonna update the version of the controller there to avoid this happening again in the future.

NOTE: It may be worth considering providing versions for the controller so consumers can leverage dependabot/renovatebot to bring in updates.

zregvart:
Looks this fixed itself, let's close

PR #1153, opened 2023-11-07
Add test for config per policy source

lcarva:
Relates #1113


lcarva:
This test demonstrates that the config per source does show up on the report. It means that something else is amiss in #1113. It is a good test to have nonetheless.

zregvart (approved):

PR #1164, opened 2023-11-15
free disk space before build and release

joejstuart:


zregvart:
I think with this we will have out of disk happen later. The stopgap measure to delete the caches every couple of months is best we can do as far as I’ve looked at this.

zregvart (approved):
Let’s have a go at this, not sure if it’ll solve the issue though…

PR #1168, opened 2023-11-20
Request 300MiB of memory for the validate step

zregvart:
The task is being OOM reaped, let's increase the requested memory so that the task is scheduled on a node that has at least 300MiB of memory available.

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -189,7 +189,9 @@ spec:
           value: "false"
       computeResources:
         requests:
-          memory: 300m
+          memory: 300Mi
+        limits:
+          memory: 300Mi
     - name: report
       image: quay.io/enterprise-contract/ec-cli:snapshot
       command: [cat]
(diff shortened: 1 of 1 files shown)


tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -187,6 +187,9 @@ spec:
         # files from an image, see https://github.com/enterprise-contract/ec-cli/issues/1109
         - name: EC_CACHE
           value: "false"
+      computeResources:
+        requests:
```
lcarva, line 191: Maybe also set a limit? https://home.robusta.dev/blog/kubernetes-memory-limit

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -187,6 +187,9 @@ spec:
         # files from an image, see https://github.com/enterprise-contract/ec-cli/issues/1109
         - name: EC_CACHE
           value: "false"
+      computeResources:
```
lcarva, line 190: I think we should try to come up with a way to make this configurable since it depends very much on how many images are being verified, and what kind of data is being loaded. Ideally, this is set at the pipeline level so the release service can adjust as needed.

(Not a blocker for this PR, of course)

PR #1173, opened 2023-11-21
Implement custom rego functions for PURL handling

### LINKED ISSUE #1053: Create a custom built-in rego function to parse PURLs
We want to write policies that evaluate the components in a CycloneDX SBOM. Those components are identified by their [purl](https://github.com/package-url/purl-spec). To facilitate policy authoring, let's introduce two rego functions:

`ec.purl.is_valid` to evaluate if a given string is a valid purl.

`ec.purl.parse` to deconstruct a valid purl string into an object that represents its various fields.

We should consider using https://github.com/package-url/packageurl-go.

https://github.com/open-policy-agent/conftest/issues/856 is a good resource for how to write a custom built-in rego function.

mbestavros:
This PR adds two basic custom Rego functions to handle PURLs: `ec.purl.is_valid` and `ec.purl.parse`. 

Marking as WIP due to lacking tests.

Resolves #1053.

### DIFF
--- a/internal/evaluator/rego.go
+++ b/internal/evaluator/rego.go
@@ -32,12 +32,15 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/types"
+	"github.com/package-url/packageurl-go"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/enterprise-contract/ec-cli/internal/fetchers/oci"
 )
 
 const ociBlobName = "ec.oci.blob"
+const purlIsValidName = "ec.purl.is_valid"
+const purlParseName = "ec.purl.parse"
 
 func registerOCIBlob() {
 	decl := rego.Function{
@@ -56,6 +59,62 @@ func registerOCIBlob() {
 	}
 
 	rego.RegisterBuiltin1(&decl, ociBlob)
+}
+
+func registerPURLIsValid() {
+	decl := rego.Function{
+		Name: purlIsValidName,
+		Decl: types.NewFunction(
+			types.Args(
+				types.Named("purl", types.S).Description("the PURL"),
+			),
+			types.Named("result", types.S).Description("PURL validity"),
+		),
+		// As per the documentation, enable memoization to ensure function evaluation is
+		// deterministic.
+		Memoize:          true,
+		Nondeterministic: false,
+	}
+
+	rego.RegisterBuiltin1(&decl, purlIsValid)
+}
+
+func registerPURLParse() {
+	decl := rego.Function{
+		Name: purlParseName,
+		Decl: types.NewFunction(
+			types.Args(
+				types.Named("purl", types.S).Description("the PURL"),
+			),
+			types.Named("object", types.NewObject(
+				[]*types.StaticProperty{
+					// Specifying the properties like this ensure the compiler catches typos when
+					// evaluating rego functions.
+					{Key: "type", Value: types.S},
+					{Key: "namespace", Value: types.S},
+					{Key: "name", Value: types.S},
+					{Key: "version", Value: types.S},
+					{Key: "qualifiers", Value: types.NewArray(
+						nil, types.NewObject(
+							[]*types.StaticProperty{
+								{Key: "key", Value: types.S},
+								{Key: "value", Value: types.S},
+							},
+							nil,
+						),
+					)},
+					{Key: "subpath", Value: types.S},
+				},
+				nil,
+			)).Description("the parsed PURL object"),
+		),
+		// As per the documentation, enable memoization to ensure function evaluation is
+		// deterministic.
+		Memoize:          true,
+		Nondeterministic: false,
+	}
+
+	rego.RegisterBuiltin1(&decl, purlParse)
 }
 
 const maxBytes = 10 * 1024 * 1024 // 10 MB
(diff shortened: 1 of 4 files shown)


internal/evaluator/rego.go
```diff
@@ -58,6 +61,44 @@ func registerOCIBlob() {
...
+			types.Args(
+				types.Named("purl", types.S).Description("the PURL"),
+			),
+			types.Named("result", types.S).Description("PURL validity"),
+		),
+		// As per the documentation, enable memoization to ensure function evaluation is
+		// deterministic. But also mark it as non-deterministic because it does rely on external
+		// entities, i.e. OCI registry. https://www.openpolicyagent.org/docs/latest/extensions/
+		Memoize:          true,
+		Nondeterministic: true,
```
lcarva, line 77: I don't think we want this to be `Nondeterministic`. It's just a string parsing method.
(Let's also update the comment above.)

Same for `registerPURLParse`.

internal/evaluator/rego.go
```diff
@@ -118,6 +159,30 @@ func ociBlob(bctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
+func purlIsValid(bctx rego.BuiltinContext, a *ast.Term) bool {
+	uri, ok := a.Value.(ast.String)
+	if !ok {
+		return false
+	}
+	instance, err := packageurl.FromString(uri)
+	if err != nil {
+		return false
+	}
+	return true
```
lcarva, line 171: I wonder if you need something like this: 

```go
return ast.BooleanTerm(true)
```

internal/evaluator/rego.go
```diff
@@ -118,6 +159,30 @@ func ociBlob(bctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
 	return ast.StringTerm(blob.String()), nil
 }
 
+func purlIsValid(bctx rego.BuiltinContext, a *ast.Term) bool {
```
lcarva, line 162: This must implement this signature:

```go
Builtin1 func(bctx BuiltinContext, op1 *ast.Term) (*ast.Term, error)
```

It's missing an error parameter (which will always be nil :disappointed:)

internal/evaluator/rego.go
```diff
@@ -118,6 +159,30 @@ func ociBlob(bctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
+		return false
+	}
+	instance, err := packageurl.FromString(uri)
+	if err != nil {
+		return false
+	}
+	return true
+}
+
+func purlParse(bctx rego.BuiltinContext, a *ast.Term) packageurl.instance {
```
lcarva, line 174: We may need to implement a special type here.

internal/evaluator/rego.go
```diff
@@ -58,6 +61,44 @@ func registerOCIBlob() {
...
+	decl := rego.Function{
+		Name: purlIsValidName,
+		Decl: types.NewFunction(
+			types.Args(
+				types.Named("purl", types.S).Description("the PURL"),
+			),
+			types.Named("result", types.S).Description("PURL validity"),
+		),
+		// As per the documentation, enable memoization to ensure function evaluation is
+		// deterministic. But also mark it as non-deterministic because it does rely on external
```
lcarva, line 74: Let's update the comment here and in registerPURLParse

internal/evaluator/rego.go
mbestavros: Done!

internal/evaluator/rego.go
mbestavros: Done!

internal/evaluator/rego.go
mbestavros: Done!

internal/evaluator/rego.go
```diff
@@ -118,6 +157,39 @@ func ociBlob(bctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
 	return ast.StringTerm(blob.String()), nil
 }
 
+func purlIsValid(bctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
+	uri, ok := a.Value.(ast.String)
+	if !ok {
+		return ast.BooleanTerm(false), nil
+	}
+	_, err := packageurl.FromString(uri.String())
```
lcarva, line 165: This needs to be `string(uri)` instead of `uri.String()`. The second form adds quotes. I ran into this with the fetch oci blob function as well. 

(same for the PURL parse function)

internal/evaluator/rego.go
```diff
@@ -49,15 +52,51 @@ func registerOCIBlob() {
...
+	decl := rego.Function{
+		Name: purlParseName,
+		Decl: types.NewFunction(
+			types.Args(
+				types.Named("purl", types.S).Description("the PURL"),
+			),
+			types.Named("object", types.S).Description("the parsed PURL object"),
+		),
+		// As per the documentation, enable memoization to ensure function evaluation is
+		// deterministic. But also mark it as non-deterministic because it does rely on external
```
lcarva, line 91: Outdated comment.

mbestavros: Fixed!

internal/evaluator/rego_test.go
```diff
@@ -106,9 +106,97 @@ func TestOCIBlob(t *testing.T) {
...
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			ctx := context.Background()
+			bctx := rego.BuiltinContext{Context: ctx}
+
+			isValid, err := purlIsValid(bctx, c.uri)
+			require.NoError(t, err)
+			if c.err {
```
lcarva, line 138: Consider changing the `cases` struct to contain specify an `expected` attribute, then change this to:

```go
require.NotNil(t, isValid)
require.Equal(t, isValid, ast.BooleanTerm(c.expected))
```

That should work for both cases.

mbestavros: That's definitely cleaner. Done.

zregvart (approved):
LGTM

lcarva (approved):

PR #1181, opened 2023-12-01
Change log regex to avoid year overlap

lcarva:


simonbaird:
Diff looks much better when whitespace is hidden.

PR #1186, opened 2023-12-04
Make rego landing page list functions

lcarva:
This adds a summary view to the docs page where rego functions are introduced.


zregvart (approved):
I'd probably call the page/file something like `rego_builtins.adoc`. Other than that LGTM

PR #1190, opened 2023-12-06
Bump conftest to 0.47.0 to unblock `EC-55`

mbestavros:
https://issues.redhat.com/browse/EC-55 was blocked on the inclusion of https://github.com/open-policy-agent/conftest/pull/877 in Conftest. With the release of Conftest `0.47.0`, the patch is now in upstream, and can now unblock EC-55.


simonbaird:
Any ideas about the failure?

```
Error: ../../../go/pkg/mod/github.com/moby/buildkit@v0.12.4/frontend/dockerfile/instructions/parse.go:592:15: healthcheck.StartInterval undefined (type container.HealthConfig has no field or method StartInterval)
```

zregvart:
I guess we need to downgrade github.com/moby/buildkit to [v0.11.5](https://github.com/open-policy-agent/conftest/blob/v0.47.0/go.mod#L22)

lcarva:
I'm likely missing something, but can someone explain how that conftest change allows "ec to produce both kinds of formatted output without needing to run twice." (Quote from the linked Jira issue.)

mbestavros:
> I'm likely missing something, but can someone explain how that conftest change allows "ec to produce both kinds of formatted output without needing to run twice." (Quote from the linked Jira issue.)

The patch I'm writing will add a flag to `ec test` that will write the `conftest` output to a file path. The issue is that, as it stands currently, `conftest` will only output to `stdout` instead of to a file. I tried capturing that `stdout` output and writing it to a file, but I figured the true solution would be an upstream patch to allow `conftest` to write directly to a file-like object when used as a library.

If that file object is something EC controls directly, we could output to both stdout and to a file at the same time.

lcarva:
> The patch I'm writing will add a flag to `ec test` that will write the `conftest` output to a file path. The issue is that, as it stands currently, `conftest` will only output to `stdout` instead of to a file. I tried capturing that `stdout` output and writing it to a file, but couldn't get it working when I tried it. I figured the true solution would be an upstream patch to allow `conftest` to write directly to a file-like object when used as a library.
> 
> If that file object is something EC controls directly, we could output to both stdout and to a file at the same time.

Ok, is there any progress in making that happen? It looks like this work is blocked on something that doesn't exist yet.

mbestavros:
@lcarva I've got a draft patch locally that doesn't work yet, because it's waiting on the updated Conftest dependency (this PR). Once this gets merged, my work on that can be unblocked.

This patch should just be a dependency bump, with no code changes.

mbestavros:
@lcarva I realized I was unclear - the aforementioned "upstream patch to allow conftest to write directly to a file-like object when used as a library" is already done and merged as https://github.com/open-policy-agent/conftest/pull/877. Conftest 0.47 (this PR) includes the patch.

lcarva:
@mbestavros , thank you for the clarification!

lcarva:
> I guess we need to downgrade github.com/moby/buildkit to [v0.11.5](https://github.com/open-policy-agent/conftest/blob/v0.47.0/go.mod#L22)

Indeed. It is odd that an older version of conftest worked with a newer version of buildkit though.

mbestavros:
@zregvart @lcarva Tests pass with the `moby/buildkit` downgrade. Any objections, or is this good to go?

lcarva (approved):
Let's do it!

PR #1192, opened 2023-12-07
Add CI triggers for redhat-* branches

simonbaird:
We want to make sure any changes made in the downstream release branches also pass the test suites.

I wasn't sure about CodeQL, so I didn't add it in this commit, but perhaps it should be added too.


zregvart (approved):

robnester-rh (approved):
LGTM

PR #1196, opened 2023-12-07
Improve Dockerfile for RHTAP builds

simonbaird:
Make the new Docker file used by the RHTAP downstream build produce an image more similar to the existing upstream step runner image.

It's not tested yet, but this image should now be useable as the step runner image for our Tekton tasks, as well as provide a place for the OpenShift cli downloader to access a binary cli for download.

It now also just builds one binary specific for the container image, rather than cross-compiling multiple binaries and making them available in a single image.

See also the patch
https://github.com/redhat-appstudio/infra-deployments/pull/2883 which demonstrates how we're likely to provide access to the cli downloads.

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -14,11 +14,24 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-# Note: There's more work to do before this image is usable for productized ec,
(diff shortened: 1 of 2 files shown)


simonbaird:
Replaced by #1197 with rebase on the redhat-v0.1-alpha branch.

PR #1197, opened 2023-12-07
Improve Dockerfile for RHTAP builds

simonbaird:
Make the new Docker file used by the RHTAP downstream build produce an image more similar to the existing upstream step runner image.

It's not tested yet, but this image should now be useable as the step runner image for our Tekton tasks, as well as provide a place for the OpenShift cli downloader to access a binary cli for download.

It now also just builds one binary specific for the container image, rather than cross-compiling multiple binaries and making them available in a single image.

See also the patch
https://github.com/redhat-appstudio/infra-deployments/pull/2883 which demonstrates how we're likely to provide access to the cli downloads.

Ref: https://issues.redhat.com/browse/EC-290

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -14,11 +14,24 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-# Note: There's more work to do before this image is usable for productized ec,
-# but it builds in RHTAP so it's a good starting point. To make this useful as
-# a Tekton step runnner, we need cosign, jq, and git, and possibly to build and
-# push it as a full multi-arch image, in which case we would do just one build
-# per container.
+## Downloads
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.3@sha256:c77792b8084ce5946c68f39024fa460ef7769c0eef3fce995e70299e21a7e166 as download
+
+ARG TARGETOS
+ARG TARGETARCH
+
+ARG COSIGN_VERSION
+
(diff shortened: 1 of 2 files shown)


Dockerfile.dist
```diff
@@ -14,11 +14,24 @@
...
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.3@sha256:c77792b8084ce5946c68f39024fa460ef7769c0eef3fce995e70299e21a7e166 as download
+
+ARG TARGETOS
+ARG TARGETARCH
+
+ARG COSIGN_VERSION
+
+ADD https://github.com/sigstore/cosign/releases/download/${COSIGN_VERSION}/cosign-${TARGETOS}-${TARGETARCH} /opt/
+ADD https://github.com/sigstore/cosign/releases/download/${COSIGN_VERSION}/cosign_checksums.txt /opt/
```
zregvart, line 27: We could copy cosign out of the securesign image, like here https://github.com/securesign/sigstore-ocp/blob/914bdbc5f29c40dcc8fc73989dd3561663324758/images/Dockerfile-clientserver#L1

simonbaird: I like the idea, but I don't think I want to do it in this PR.

simonbaird:
Went green at https://console.redhat.com/preview/application-pipeline/workspaces/rhtap-contract/applications/ec-redhat-v01-alpha/pipelineruns/ec-cli-on-pull-request-77m58/logs .

So I'm thinking let's merge.

Dockerfile.dist
simonbaird: Yeah, that is very cool.

PR #1199, opened 2023-12-07
Increase requests and limits for the validate step

ralphbean:
This is a follow up to #1168, in reaction to https://issues.redhat.com/browse/RHTAPBUGS-1045

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -189,9 +189,10 @@ spec:
           value: "false"
       computeResources:
         requests:
-          memory: 300Mi
+          cpu: 250m
+          memory: 500Mi
         limits:
-          memory: 300Mi
+          memory: 2Gi
     - name: report
       image: quay.io/enterprise-contract/ec-cli:snapshot
       command: [cat]
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -190,7 +190,7 @@ spec:
       computeResources:
         requests:
           cpu: 250m
-          memory: 500Mi
+          memory: 2Gi
         limits:
           memory: 2Gi
     - name: report


simonbaird (approved):

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -189,9 +189,11 @@ spec:
           value: "false"
       computeResources:
         requests:
-          memory: 300Mi
+          cpu: 250m
+          memory: 500Mi
         limits:
-          memory: 300Mi
+          cpu: 2
```
lcarva, line 195: I hear CPU limits is discouraged? See changes in build-definitions: https://github.com/redhat-appstudio/build-definitions/pull/673

lcarva: Actually... it's not clear to me why we are setting CPU requests/limits on the PR at all. The reported issue seems to be about memory.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -189,9 +189,11 @@ spec:
...
+          memory: 2Gi
```
lcarva, line 196: This should match the request as memory is not a resource that can be easily deallocated without killing the pod.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
ralphbean: The reported issue _is_ about memory, however we are seeing cluster-wide issues due to tasks in general not setting cpu _requests_: over-subscribed nodes.

I'll remove the cpu limits.

lcarva:
We're getting a failure in RHTAP:

```
*Warning*: This is an experimental command, it's usage and behavior can change in the next release(s)
	- Added Task: verify-enterprise-contract to image
Error: could not push image to registry as "quay.io/redhat-appstudio/pull-request-builds:task-verify-ec-c5b7178a7da68bde30a94107b943d4ba232521e2": POST https://quay.io/v2/redhat-appstudio/pull-request-builds/blobs/uploads/: UNAUTHORIZED: access to the requested resource is not authorized; map[]
```

I think that's bogus and may have to do with the experimentation @simonbaird is doing. I'm gonna go ahead and merge this.

PR #1201, opened 2023-12-08
Merge branch 'redhat-v0.1-alpha' back into main

simonbaird:
```
    Merge branch 'redhat-v0.1-alpha' back into main
    
    We're still figuring out the best way to handle the Red Hat specific
    release branches, but for now let's bring the RHTAP release branch
    changes back into main.
    
    The main changes coming in with this commit:
    - The new docker file Dockerfile.dist used by TAP builds
    - The two new pipelines under .tekton created by TAP
    - The removal of the old tekton-ci pipelines under .tekton
    
    I'm planning to update the redhat-v0.1-alpha branch to the top of main
    after this.
```


simonbaird:
There's no new code here except the typo fix in the top commit, so no need to review the diff.

There could be other ways to do this, e.g. cherry pick all the changes and scrap the existing release branch, but I think this is better since the history PRs from this week in the release branch is preserved.

Anyway, let's merge. :cowboy_hat_face:

PR #1230, opened 2023-12-15
Remove scheme from policy URL

robnester-rh:
This commit strips the scheme (https:// or http://) from the repository URL for the policy to ensure that it doesn't break go-getter when attempting to download.

ref: EC-285

### DIFF
--- a/internal/policy/source/source.go
+++ b/internal/policy/source/source.go
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
 	"time"
 
 	ecc "github.com/enterprise-contract/enterprise-contract-controller/api/v1alpha1"
@@ -67,7 +68,10 @@ type PolicyUrl struct {
 
 // GetPolicies clones the repository for a given PolicyUrl
 func (p *PolicyUrl) GetPolicy(ctx context.Context, workDir string, showMsg bool) (string, error) {
+	// Trim scheme from repo url as this breaks go-getter.
 	sourceUrl := p.PolicyUrl()
+	sourceUrl = strings.TrimPrefix(sourceUrl, "https://")
+	sourceUrl = strings.TrimPrefix(sourceUrl, "http://")
 
 	dest := uniqueDestination(workDir, p.Subdir(), sourceUrl)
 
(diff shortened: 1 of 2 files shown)


joejstuart (approved):

lcarva:
Is this something we want to do in all cases? Or just when GitHub is used? It'll be good to provide the reasoning for the solution. "breaks go-getter" is quite vague and won't really help us figure out why this was added a few months from now.

internal/policy/source/source.go
```diff
@@ -67,7 +68,10 @@ type PolicyUrl struct {
 
 // GetPolicies clones the repository for a given PolicyUrl
 func (p *PolicyUrl) GetPolicy(ctx context.Context, workDir string, showMsg bool) (string, error) {
+	// Trim scheme from repo url as this breaks go-getter.
 	sourceUrl := p.PolicyUrl()
+	sourceUrl = strings.TrimPrefix(sourceUrl, "https://")
```
zregvart, line 73: We support fetching policies via HTTP as well, so this will break that.

internal/policy/source/source.go
```diff
@@ -67,7 +68,10 @@ type PolicyUrl struct {
...
+	sourceUrl = strings.TrimPrefix(sourceUrl, "http://")
```
zregvart, line 74: Not allowed, see https://github.com/enterprise-contract/ec-cli/blob/2bbed8f8295f2b87fc5c64fb7e7c0bb774a25443/internal/downloader/downloader.go#L49C1-L51.

zregvart (changes requested):
I don't think this is the path we want to go down on. This switches protocol from HTTP to git. It is not the go-getter that is broken here it is the assumption to whatever the URL pointed to would contain certain directory structure, i.e. a `policy` directory.

internal/policy/source/source.go
robnester-rh: Ack. From discussion with @joejstuart, the issue was only seen when the URL had an `https://` scheme with no prepended `git::`. When a URL beginning with `github.com` was used, there was no issue.

internal/policy/source/source.go
robnester-rh: Ack.

PR #1244, opened 2023-12-19
Implement option to write output to a file in `ec test`

mbestavros:
This PR adds an option to `ec test` to allow output to a file. Users can use the new option with `ec test ... --file <file path>`.

The PR utilizes [new functionality](https://github.com/open-policy-agent/conftest/pull/877) available in Conftest `0.47` to write Conftest output to a file.

Resolves [EC-55](https://issues.redhat.com/browse/EC-55).

### DIFF
--- a/cmd/test/test.go
+++ b/cmd/test/test.go
@@ -118,6 +118,7 @@ func newTestCommand() *cobra.Command {
 				"no-fail",
 				"suppress-exceptions",
 				"output",
+				"file",
 				"parser",
 				"policy",
 				"proto-file-dirs",
@@ -150,6 +151,11 @@ func newTestCommand() *cobra.Command {
 				return fmt.Errorf("unmarshal parameters: %w", err)
 			}
 
+			outputFilePath, err := cmd.Flags().GetString("file")
+			if err != nil {
+				return fmt.Errorf("reading flag: %w", err)
+			}
+
 			results, resultsErr := runner.Run(ctx, fileList)
 			var exitCode int
 			if runner.FailOnWarn {
@@ -171,7 +177,14 @@ func newTestCommand() *cobra.Command {
 					if err != nil {
 						return appstudioErrorHandler(runner.NoFail, "output results", err)
 					}
-					fmt.Printf("%s\n", reportOutput)
+
+					if outputFilePath != "" {
+						err := os.WriteFile(outputFilePath, reportOutput, 0600)
+						if err != nil {
+							return fmt.Errorf("creating output file: %w", err)
+						}
+					}
+					fmt.Fprintln(cmd.OutOrStdout(), string(reportOutput)+"\n")
 
 				} else {
 					// Conftest handles the output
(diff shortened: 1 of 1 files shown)


cmd/test/test.go
```diff
@@ -180,15 +194,32 @@ func newTestCommand() *cobra.Command {
...
 						Tracing:            runner.Trace,
 						JUnitHideMessage:   viper.GetBool("junit-hide-message"),
+						File:               outputFile,
 					})
 					if err := outputter.Output(results); err != nil {
 						return fmt.Errorf("output results: %w", err)
 					}
+
+					if outputFilePath != "" {
+						_, err := io.Copy(outputFile, os.Stdout)
```
lcarva, line 218: I think you meant to swap these parameters :wink:

cmd/test/test.go
```diff
@@ -180,15 +194,32 @@ func newTestCommand() *cobra.Command {
...
 						SuppressExceptions: runner.SuppressExceptions,
 						Tracing:            runner.Trace,
 						JUnitHideMessage:   viper.GetBool("junit-hide-message"),
+						File:               outputFile,
 					})
 					if err := outputter.Output(results); err != nil {
 						return fmt.Errorf("output results: %w", err)
 					}
+
+					if outputFilePath != "" {
```
lcarva, line 217: Consider instead check if `outputFile` is `nil` since that's the requirement for the code block below.

lcarva: Actually... I don't think you need this block at all. Shouldn't the code above handle if `ouput.Options.File` is `nil`?

lcarva (commented):
It's unfortunate that this part of the code base has no unit test at all. Be sure to perform some manual tests.

cmd/test/test.go
mbestavros: @lcarva I put this block in to copy the output we wrote to the file to `stdout` as well. That was part of the acceptance criteria for the linked issue: that it would be possible to output to a file and `stdout` at the same time.

I'm fine leaving it out if we don't need that behavior though.

lcarva: I commented out this bit of the code and it still behaved the same way. (With `-file` output went just to the file. Without it, output went just to stdout.)

mbestavros: @lcarva I'm realizing we aren't even touching that part of the code when we pass `--output appstudio`, so it wouldn't matter. You need to do something like `--output json` to make Conftest handle output.

If we do that, running without the code block makes output happen on either the file or `stdout`, but not both - which was part of the acceptance criteria. However, leaving the code in appears to hang when I run it, so that's not good either...

lcarva: It hangs because of [this](https://github.com/enterprise-contract/ec-cli/pull/1244#discussion_r1431967846). Let me run a couple more tests here.

simonbaird:
Not sure if the previous comments were addressed or not. Is this ready for fresh reviews @mbestavros ?

mbestavros:
@simonbaird I wasn't able to get `io.Copy()` working like I expected, so I pushed another workaround. Should be good to go for another review.

cmd/test/test.go
```diff
@@ -171,6 +177,13 @@ func newTestCommand() *cobra.Command {
...
 						return appstudioErrorHandler(runner.NoFail, "output results", err)
 					}
+
+					if outputFilePath != "" {
+						err := os.WriteFile(outputFilePath, reportOutput, 0600)
+						if err != nil {
+							return fmt.Errorf("creating output file: %w", err)
+						}
+					}
 					fmt.Printf("%s\n", reportOutput)
```
zregvart, line 187: Just noticed we're not using `cmd.OutOrStdout()` here, we should, could be in a followup PR as this wasn't added here.

cmd/test/test.go
```diff
@@ -180,15 +193,33 @@ func newTestCommand() *cobra.Command {
...
 					if err := outputter.Output(results); err != nil {
 						return fmt.Errorf("output results: %w", err)
 					}
+
+					if outputFilePath != "" {
+						contents, err := os.ReadFile(outputFile.Name())
+						if err != nil {
+							return fmt.Errorf("copying output file to stdout: %w", err)
+						}
+						fmt.Println(string(contents))
```
zregvart, line 221: I think we want to use `cmd.OutOrStdout()` here
```suggestion
						fmt.Fprintln(cmd.OutOrStdout(), string(contents))
```

zregvart (commented):
Some nitpickery, otherwise LGTM

cmd/test/test.go
mbestavros: Done!

cmd/test/test.go
mbestavros: I tossed it in!

cmd/test/test.go
```diff
@@ -171,7 +177,14 @@ func newTestCommand() *cobra.Command {
...
 					}
-					fmt.Printf("%s\n", reportOutput)
+
+					if outputFilePath != "" {
+						err := os.WriteFile(outputFilePath, reportOutput, 0600)
+						if err != nil {
+							return fmt.Errorf("creating output file: %w", err)
+						}
+					}
+					fmt.Fprintln(cmd.OutOrStdout(), string(reportOutput))
```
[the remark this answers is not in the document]
mbestavros: @lcarva Added that in!

lcarva (approved):

PR #1262, opened 2024-01-03
Script for applying RHTAP pipeline changes

simonbaird:
Since each release requires a new RHTAP application and a new pipeline definition, we need a way to apply the required pipeline changes for each release in the branch for that release.

This commit adds some bash scripting to do that. It's expected the script will be run by a human so they can review the result manually.

Ref: [EC-342](https://issues.redhat.com/browse/EC-342)

### DIFF
--- a/hack/cut-release.sh
+++ b/hack/cut-release.sh
@@ -86,6 +86,12 @@ Edit ${RHTAP_APPLICATION_NAME}-enterprise-contract and add a parameter as follow
   Value: github.com/enterprise-contract/config//redhat-no-hermetic
 Save changes
 
+# Apply pipeline modifications
+git checkout ${BRANCH_NAME}
+hack/patch-release-pipelines.sh
+Review the generated commit and then create a PR for the ${BRANCH_NAME} branch with that commit
+(Todo maybe: If you want, try adding this commit to the PR created by RHTAP before merging that PR.)
+
 EOT
 
 # Todo: What about the RPA?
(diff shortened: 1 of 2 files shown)


simonbaird:
I exercised it with a "release-v0.1-alpha2" branch and I believe it worked reasonably well. Found one little bug which is now fixed via:

```diff
-  release_pipeline=".tekton/cli-${release_name}-$p.yaml"                                                             
+  release_pipeline=".tekton/cli-${release_name/./}-$p.yaml"   
```

zregvart:
LGTM, though if we create release branches from main, wouldn't it already contain the pipelines we need in the state we need? Then it's renaming them and modifying the target branch which could be done in the first commit to the new release branch.
I see that there are pipeline definitions for main branch in https://github.com/enterprise-contract/ec-cli/tree/release-v0.1-alpha2/.tekton, do we want those?

simonbaird:
> LGTM, though if we create release branches from main, wouldn't it already contain the pipelines we need in the state we need? Then it's renaming them and modifying the target branch which could be done in the first commit to the new release branch.

Yeah, that's a highly reasonable suggestion and we were heading in that direction at one point. As mentioned in slack my thinking is for now we should do things the "RHTAP way", i.e. let RHTAP create the PR with the pipelines. 

> I see that there are pipeline definitions for main branch in https://github.com/enterprise-contract/ec-cli/tree/release-v0.1-alpha2/.tekton, do we want those?

We removed them for the `v0.1-alpha` branch iirc, so yeah, let's do that also. Will revise this PR to git rm them.

PR #1268, opened 2024-01-04
Use git for-each-ref to detect branch name

simonbaird:
We're using the branch name to derive the release name, but it doesn't work in the TAP build pipeline because the git-clone task checks out the sha directly and not the branch.

With this patch it should find the branch name correctly anyway. We're making some assumptions: there is only one branch name that points to the HEAD sha, and the git-clone task pulled down the branch refs, which does seem to be happening as a side-effect of fetching down the tags. (See aso the fetchTags param in the git-clone task which we set to "true".)

Ref: [EC-343](https://issues.redhat.com/browse/EC-343)

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -40,6 +40,7 @@ WORKDIR /build
 
 RUN go mod download
 
+# Notes:
 # Derive the version from the branch name with the "release-" prefix removed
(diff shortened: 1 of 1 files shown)


simonbaird:
I realized this doesn't work until after it is merged, so I'm going to merge and see how it goes in the post-merge build.

PR #1269, opened 2024-01-04
Set git-clone task refspec param to fetch branches

simonbaird:
Because we want to use the branch names to extract a version name when building we need to have the branch refs available.

Ref: [EC-343](https://issues.redhat.com/browse/EC-343)


simonbaird:
Refspec doesn't work like I though it did, and actually I don't think we need it.

PR #1277, opened 2024-01-10
wip: Introduce the concept of a validation intent

lcarva:
This commit adds a new flag to the `ec validate command`, `--intent` which is propagated to rego as config value - similarly to how effectiveTime is propagated.

The idea is to provide a mechanism to change the behavior of a policy rule depending on why the user is validating an image.

For example, the
[schedule](https://enterprisecontract.dev/docs/ec-policies/release_policy.html#schedule_package) package contains policy rules that are meant to enforce restrictions depending on the day. This is meant to disallow releases on the weekends or on certain days (holidays).

Those policy rules are NOT meant to disallow changes from being merged.

This allows the user to use the same policy config for multiple purposes.

Ok, so why not just have multiple policy configs?

This is certainly a possibility! It does come with its own challenges:

1. Deviation in what is checked during integration vs release is prone to cause unexpected issues at release time.
1. Maintenance. Duplicating every policy config sounds like a good way to increase toil for everyone.

The "intent" appraoch does away with the maintenance cost for users, at the expense of making certain policy rules slightly more complex. (I think that's a good trade-off.)

But it only reduces the deviation problem. You'd still get a different result in integration than in release. But... that's what the requested feature is all about so not sure if we can do much about that. This approach does use the same policy config in both cases so at least it removes issues caused by those not being in sync.

Anyways, this is a proof of concept meant to help the conversation.

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -62,6 +62,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		spec                        *app.SnapshotSpec
 		strict                      bool
 		images                      string
+		intent                      string
 	}{
 		strict: true,
 	}
@@ -207,6 +208,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 				PolicyRef:   data.policyConfiguration,
 				PublicKey:   data.publicKey,
 				RekorURL:    data.rekorURL,
+				Intent:      data.intent,
 			}); err != nil {
 				allErrors = multierror.Append(allErrors, err)
 			} else {
@@ -384,6 +386,10 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		Include additional information on the failures. For instance for policy
 		violations, include the title and the description of the failed policy
 		rule.`))
+
+	cmd.Flags().StringVar(&data.intent, "intent", data.intent, hd.Doc(`
+		Specify the intent of validating an image.
+	`))
 
 	if len(data.input) > 0 || len(data.filePath) > 0 || len(data.images) > 0 {
 		if err := cmd.MarkFlagRequired("image"); err != nil {
--- a/internal/applicationsnapshot/report.go
+++ b/internal/applicationsnapshot/report.go
@@ -57,6 +57,7 @@ type Report struct {
 	EcVersion     string                           `json:"ec-version"`
 	Data          any                              `json:"-"`
 	EffectiveTime time.Time                        `json:"effective-time"`
+	Intent        string                           `json:"intent"`
 	PolicyInput   [][]byte                         `json:"-"`
 }
 
@@ -141,6 +142,7 @@ func NewReport(snapshot string, components []Component, policy policy.Policy, da
 		Data:          data,
 		PolicyInput:   policyInput,
 		EffectiveTime: policy.EffectiveTime().UTC(),
+		Intent:        policy.Intent(),
 	}, nil
 }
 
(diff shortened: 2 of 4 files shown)


lcarva:
You can try this out by building EC from this PR and using this policy config:

```yaml
---
publicKey: |
  -----BEGIN PUBLIC KEY-----
  MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEZP/0htjhVt2y0ohjgtIIgICOtQtA
  naYJRuLprwIv6FDhZ5yFjYUEtsmoNcW7rx2KM6FOXGsCX3BNc7qhHELT+g==
  -----END PUBLIC KEY-----
sources:
  - policy:
      # - /home/lucarval/src/enterprise-contract/ec-policies/policy/lib
      # - /home/lucarval/src/enterprise-contract/ec-policies/policy/release
      - github.com/lcarva/ec-policies//policy/lib?ref=intent-poc
      - github.com/lcarva/ec-policies//policy/release?ref=intent-poc
    ruleData:
      disallowed_weekdays:
        - Sunday
        - Monday
        - Tuesday
        - Wednesday
        - Thursday
        - Friday
        - Saturday
    config:
      include:
        - schedule

```

```bash
# This should work
ec validate image --image quay.io/redhat-appstudio/ec-golden-image:latest --ignore-rekor --policy policy.yaml

# So should this
ec validate image --image quay.io/redhat-appstudio/ec-golden-image:latest --ignore-rekor --policy policy.yaml

# But not this
ec validate image --image quay.io/redhat-appstudio/ec-golden-image:latest --ignore-rekor --policy policy.yaml --intent release
```

zregvart:
Perhaps we could benefit from having this as a generic mechanism. I'm thinking `intent` could be one kind of pivot for the rules and there might be other kinds as well. First thought for me was to include the notion of a `profile` and that the sources within the policy configuration could say they belong to a profile (by default all profiles).

Example:

```
$ ec validate image --profile release ...
```

That would include `schedule.weekday_restriction` via configuration like:

```
sources:
  - name: Release profile
    policy:
    data:
      - oci::quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest
      - github.com/release-engineering/rhtap-ec-policy//data
    profile: release
	include:
      - schedule
  - name: Default
    policy:
    data:
      - oci::quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest
      - github.com/release-engineering/rhtap-ec-policy//data
	exclude:
      - schedule
```

lcarva:
Profiles are an interesting idea. They would get applied for each source group. We don't have to run the `schedule` rules as a separate source group, but we sure can!

I like that this requires no changes to the policy themselves, so it's a little bit more flexible IMO.

I think the additional source group in the policy config is not too bad. It does duplicate the policy/data sources but I think that's ok. And if `ruleData` is used, that doesn't have to be duplicated because it's not relevant to the release source group, presumably.

simonbaird:
Re the idea of command line influence on the policies, we could make it more even more generic - I'm thinking about the way that ansible-playbook supports -e/--extra-var. Then we could do `ec ... -e intent=release` for example and it would set `data.rule_data.intent` or maybe just `data.intent` to "release" so it's visible to rego rules.

lcarva:
> Re the idea of command line influence on the policies, we could make it more even more generic - I'm thinking about the way that ansible-playbook supports -e/--extra-var. Then we could do `ec ... -e intent=release` for example and it would set `data.rule_data.intent` or maybe just `data.intent` to "release" so it's visible to rego rules.

Really like this idea, or some variation of it. We just have to make sure those passed in values are displayed properly in the report to avoid ambiguity on what was used to validate.

lcarva:
I'm gonna close this out as we don't want to merge this as is. I think this poc has served its purpose, which was to kick off the conversation. thanks all.

PR #1281, opened 2024-01-11
Fix acceptance test and snapshots

zregvart:
With EC-224 and the change in [1] we time-bombed the acceptance tests. This modifies the scenario slightly so a different rule, one emitting a warning, is used. I believe the original intent of the test is maintained.

[1] https://github.com/enterprise-contract/ec-policies/pull/793


lcarva (approved):
I think this actually restores the original intent of the those tests, `Non strict with warnings` and `Strict with warnings`. It sounds like they should've had a warning all along.

PR #1283, opened 2024-01-15
Fix issues reported by Snyk

zregvart:


lcarva:
Oh I think we need to update the regex for the EC version

lcarva (approved):

zregvart:
The checks pass now, the issue was that the `task` variable was redefined within `image.go`, so when I changed all references to `task.Name()` to `taskName` the new value was lost. SAST Snyk task now reports success. Going to merge this now.

PR #1290, opened 2024-01-16
chore: Modify default pipelines for v0.1-alpha

simonbaird:
Apply changes to the RHTAP generated default pipelines.

(Commit created with hack/patch-release-pipelines.sh )


simonbaird:
The commit is generated so no need for review. CI is green, so let's merge and see if we push out a release.

PR #1295, opened 2024-01-19
Provide better error messaging around policy URLs

robnester-rh:
This comment addresses EC-285 and attempts to provide a better error message to users when they provide a policy URL in a format that is ambiguous or difficult for go-getter to parse.

Users will now receive suggestions if it appears they were trying to reference a github or gitlab repository, but include a scheme ('https' or 'http').


simonbaird (approved):
Consider adding an acceptance test that demonstrates the new behavior.

PR #1300, opened 2024-01-22
Gzip extra binaries built in Dockerfile.dist

simonbaird:
As mentioned in the comments, the other RHTAS cli binaries get gzipped, so we'll be consistent with those.

Ref: https://issues.redhat.com/browse/EC-292

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -142,6 +142,10 @@ COPY --from=download /download/cosign /usr/local/bin/cosign
 # your dist directory, not just the freshly built binaries.)
 COPY --from=build /build/dist/ /usr/local/bin/
 
+# Gzip them because that's what the cli downloader image expects, see
+# https://github.com/securesign/sigstore-ocp/blob/main/images/Dockerfile-clientserver
+RUN gzip /usr/local/bin/ec_*
+
 # Copy the one ec binary that can run in this container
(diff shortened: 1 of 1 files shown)


simonbaird:
This is fairly trivial, so I'm going to insta-merge.

Dockerfile.dist
```diff
@@ -142,6 +142,10 @@ COPY --from=download /download/cosign /usr/local/bin/cosign
 # your dist directory, not just the freshly built binaries.)
 COPY --from=build /build/dist/ /usr/local/bin/
 
+# Gzip them because that's what the cli downloader image expects, see
+# https://github.com/securesign/sigstore-ocp/blob/main/images/Dockerfile-clientserver
+RUN gzip /usr/local/bin/ec_*
```
lcarva, line 147: Wait! Isn't this going to **replace** the EC binaries with a gzipped version? That is not going to work when someone uses the ec-cli image in a Tekton Task, for example.

simonbaird: In line 146 we copy the one binary that matches the os/arch of the container we're building. So we should be fine.

simonbaird: I mean line 150, see below.

simonbaird: That one won't get gzipped.

lcarva: Aha! Thanks

PR #1303, opened 2024-01-23
Set limit of parallel checks when validating images

lcarva:
In order to validate a container image, EC gathers quite a bit of data, e.g.  SLSA Provenance, SBOM, etc. All of this is fed into rego. At some point it is all loaded into memory at the same time.

When validating multiple images, EC does that for each image.

This can be quite a bit of data depending on what is attached to each image.

Prior to this change, the `ec validate image` command would happily validate all the images in parallel. Given that memory is constrained resource, this is problematic.

This commit adds a limit to the amount of images that are validated in parallel. It hardcodes it to 5. If there are less than 5 images to validate, then the other workers are a no-op. If there are more, at most 5 will be processed at a time. Of course, 5 is a magic number. It is a starting point. Eventually, we need to make this number configurable, likely via a new CLI parameter and/or automatically adjusted based on available resources. That is for another day.

This should resolve RHTAPBUGS-1121.

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -21,12 +21,12 @@ import (
 	"errors"
 	"fmt"
 	"sort"
-	"sync"
 
 	hd "github.com/MakeNowJust/heredoc"
 	"github.com/hashicorp/go-multierror"
 	app "github.com/redhat-appstudio/application-api/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


simonbaird:
I think if it passes CI it's probably okay.

simonbaird (approved):

zregvart (commented):
:+1: Nice!

PR #1304, opened 2024-01-24
ADD acceptance tests for `ec validate input`

robnester-rh:
This commit adds acceptance tests for `ec validate input`.

Resolves: EC-364

### DIFF
--- a/features/__snapshots__/validate_input.snap
+++ b/features/__snapshots__/validate_input.snap
@@ -1,0 +1,55 @@
+
+[invalid policy URL:stdout - 1]
+
+---
+
+[invalid policy URL:stderr - 1]
+Error: 1 error occurred:
+    * error validating file pipeline_definition.yaml: no rego files found in policy subdirectory
+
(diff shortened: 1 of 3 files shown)


lcarva:
@robnester-rh, why was this merged without a review?

features/__snapshots__/validate_input.snap
```diff
@@ -0,0 +1,55 @@
+
+[invalid policy URL:stdout - 1]
```
lcarva, line 2: I don't see a scenario for this. Outdated snapshot?

robnester-rh:
> @robnester-rh, why was this merged without a review?
Apologies, I'd worked with @simonbaird on these directly. I'll make sure there's at least one lgtm next time.

features/__snapshots__/validate_input.snap
robnester-rh: Yes. I'll update and create a PR.

PR #1305, opened 2024-01-24
Prune records by taking their tag into account

lcarva:
https://issues.redhat.com/browse/EC-340


simonbaird (approved):
I didn't test it, or deeply analyze the logic, but lgtm in a shallow sense.

lcarva:
> I didn't test it, or deeply analyze the logic, but lgtm in a shallow sense.

Thanks! I did spot test this against the acceptable bundles image and things looked as expected. Merging.

PR #1307, opened 2024-01-26
Intermediate client server poc

simonbaird:
It's a long story, but this is an alternative workaround for the
problem where images build from
https://github.com/securesign/sigstore-ocp/blob/9024422e356c22660af7870aecd55a0571f07f58/images/Dockerfile-clientserver
have too many base images, so they overflow the tekton task result
that lists all the base images.

Just a POC so far but it works for me locally. Not sure what the
next step would be, but we could make a component and try building
this in Konflux. And we need an updated base image since the one I'm
using here is very old and doesn't have all the different arches.

### DIFF
--- a/Dockerfile.tas-client-server
+++ b/Dockerfile.tas-client-server
@@ -1,0 +1,18 @@
+
+# Built with Dockerfile.dist
+FROM quay.io/redhat-user-workloads/rhtap-contract-tenant/ec-v01-alpha/cli-v01-alpha@sha256:6aca9b2e9fde177d1398820b4b0609865dc2b0573c2d9e6d475e515a6a123f1c AS ec-image
+#FROM registry.redhat.io/rhtas-tech-preview/ec-rhel9:1.0.alpha-1705963542@sha256:6aca9b2e9fde177d1398820b4b0609865dc2b0573c2d9e6d475e515a6a123f1c AS ec-image
+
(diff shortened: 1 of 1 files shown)


simonbaird:
```
podman build -f Dockerfile.tas-client-server -t testeroni && podman run --rm -it --entrypoint /bin/bash testeroni -c "find /var/www/html/clients -type f" 
...
/var/www/html/clients/darwin/cosign.gz
/var/www/html/clients/darwin/gitsign.gz
/var/www/html/clients/darwin/rekor-cli.gz
/var/www/html/clients/darwin/ec-arm64.gz
/var/www/html/clients/darwin/ec-amd64.gz
/var/www/html/clients/linux/cosign.gz
/var/www/html/clients/linux/gitsign.gz
/var/www/html/clients/linux/rekor-cli.gz
/var/www/html/clients/linux/ec-s390x.gz
/var/www/html/clients/linux/ec-ppc64le.gz
/var/www/html/clients/linux/ec-arm64.gz
/var/www/html/clients/linux/ec-amd64.gz
/var/www/html/clients/windows/cosign.gz
/var/www/html/clients/windows/gitsign.gz
/var/www/html/clients/windows/rekor-cli.gz
/var/www/html/clients/windows/ec-amd64.gz
```

Notice the different arches are missing because I'm using an old (from November) base image.

PR #1340, opened 2024-02-12
Fix 'concurrent map writes' error

lcarva:
conftest's Download function leverages oras under the hood to fetch from OCI. It uses the global oras client and sets the user agent to "conftest". This is not a thread safe operation.

This commit uses a mutex to ensure a single download happens at a time.


robnester-rh (approved):
LGTM

PR #1341, opened 2024-02-14
Track bundles in the new unified format

lcarva:
Ref: EC-383

### DIFF
--- a/internal/tracker/tracker.go
+++ b/internal/tracker/tracker.go
@@ -20,6 +20,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"sort"
+	"strings"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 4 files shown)


internal/tracker/tracker.go
```diff
@@ -239,14 +292,75 @@ func filterRecords(records []bundleRecord, prune bool) []bundleRecord {
...
+				ExpiresOn:   task.ExpiresOn,
+			})
+		}
+	}
+
+	for _, bundles := range t.TaskBundles {
+		// Sort the task bundles in reverse order. The first bundle being the most recent one. The
+		// sorting function returns true if the bundle at "i" is considered newer than the bundle at
+		// "j". It is assumed that every bundle has an EffectiveOn date and a Tag, but some bundles
+		// may not have an ExpiresOn date.
```
simonbaird, line 332: Nice commentary.

simonbaird (approved):
I might be misremembering the plan, but are we still intending to phase out the "expires_on" field from the trusted tasks data? If so then perhaps mention that in commit message and/or some commentary.

zregvart (approved):
Do we need the `effective_on` now?

lcarva:
Filed EC-423 and EC-424 to deal with removing `effective_on`.

PR #1356, opened 2024-02-20
New custom rego function ec.oci.image_manifest

lcarva:


lcarva:
This is a draft PR that shows different ways of implementing a new custom rego function. 

The first commit on the PR has the most simplistic implementation. It simply converts the retrieved Image Manifest to JSON and returns it. On the rego side, a policy author then has to call `json.unmarshal` to access attributes on the Image Manifest, e.g. `json.unmarshal(ec.oci.image_manifest(ref)).schemaVersion`.

The last commit (better seen in the PR diff) lays out a rego object that has 1:1 mappings to all the attributes of the Image Manifest. Implementation is quite verbose. The main benefit is on the rego side. If a policy author makes a typo, e.g. `ec.oci.image_manifest(ref).schemaVersionnnnn` instead of `ec.oci.image_manifest(ref).schemaVersion`, the rego compiler complains about it like this:

```
Error: 1 error occurred:
        * error validating file foo.json: load: loading policies: get compiler: 1 error occurred: /tmp/ec-work-111017712/policy/7bf2fdb5b/yolo.rego:25: rego_type_error: undefined ref: im.schemaVerionnnnn
        im.schemaVerionnnnn
           ^
           have: "schemaVerionnnnn"
           want (one of): ["annotations" "config" "layers" "mediaType" "schemaVersion" "subject"]
```

Catching those errors seems like a useful thing to have. The error message from rego is also super helpful. Is it worth the more complex implementation? Any feedback is welcome.

joejstuart:
Whatever makes Rego easier to debug is a good thing. Do you see there being a lot of maintenance with the second implementation? Sounds like maybe we'll create a dependency between ec-cli and the policies? I don't think is a big deal if the code isn't changing a lot. Just curious.

zregvart:
I prefer the type safe variant, if at any point the maintenance of those properties does become a burden we could build a code generator...

lcarva:
> Whatever makes Rego easier to debug is a good thing. 

:100: 


> Do you see there being a lot of maintenance with the second implementation? Sounds like maybe we'll create a dependency between ec-cli and the policies? I don't think is a big deal if the code isn't changing a lot. Just curious.

The dependency that I see is on the policies side. I think both approaches provide a similar level of dependency though. Both rely on the function provided by the ec-cli to return a certain value. If the Image Manifest format changes the second approach makes it more explicit. But ultimately, the policy has to be aware of this one way or another.

lcarva:
I'll go ahead with the static object implementation. Thanks for the feedback! :pray:

PR #1365, opened 2024-02-22
Tasks fail EC validation

joejstuart:
The tasks fail EC validation without the "---" above the comments.


zregvart (approved):

simonbaird:
I can't think of a reason not to instamerge. :cowboy_hat_face: .

PR #1375, opened 2024-02-28
ec test: add option to force appstudio output when writing to file

mbestavros:
This PR adds a new boolean flag to the `ec test` command, `--file-appstudio-output`, that forces output to be written to disk in `appstudio` format. Intended to be used alongside the `--file` parameter.

Follow-up for https://issues.redhat.com/browse/EC-55.


simonbaird:
I'm still not clear on how it should be used. Can you share a command line example?

mbestavros:
@simonbaird Yep!

```
EC_EXPERIMENTAL=1 ec test --policy acceptance/examples/happy_day.rego acceptance/examples/empty_input.json --output json --file test.json --file-appstudio-output
```

This will send JSON output to `stdout`, and appstudio output to `test.json`.

zregvart:
I think it would make more sense to utilize the existing `--output=<format>=<file>` parameters we use in `ec verify` commands.

simonbaird:
> I think it would make more sense to utilize the existing `--output=<format>=<file>` parameters we use in `ec verify` commands.

Yeah, that would definitely be better and more consistent.

zregvart:
We discussed merging this, it is behind the experimental feature flag; and gently encouraging folk to use `ec validate input` instead.

robnester-rh:
As a note, if you intend one flag to be used in conjunction with another, and neither are required, you can use `MarkFlagsRequiredTogether("flag1", "flag2") as seen [here](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#flag-groups)

lcarva:
What's up with this PR?

mbestavros:
Closing in favor of https://github.com/enterprise-contract/ec-cli/pull/1482.

PR #1377, opened 2024-02-28
Build task bundle in release branch

joejstuart:
Build the verify-enterprise-contract task bundle on pull request and push. Also, modify the patch script to support the bundle builds.

https://issues.redhat.com/browse/EC-428

### DIFF
--- a/hack/patch-release-pipelines.sh
+++ b/hack/patch-release-pipelines.sh
@@ -32,41 +32,47 @@ fi
 
 # The release name is the branch name with release- trimmed off the front
 release_name=${current_branch#"release-"}
+main_release_name="main-ci"
+components=(cli verify-enterprise-contract-task)
 
 # One each for the two pipelines created by RHTAP
 for p in pull-request push; do
-  main_pipeline=".tekton/cli-main-ci-$p.yaml"
-  release_pipeline=".tekton/cli-${release_name/./}-$p.yaml"
 
-  if [[ "$digest_bumps" != "digest_bumps" ]]; then
-    # Find all significant changes.
-    # Use grep to exclude digest bumps and initial creation.
-    changes=$( git log main --reverse --pretty=%h --invert-grep --regexp-ignore-case \
-      --grep="Update RHTAP references" \
-      --grep="Red Hat Trusted App Pipeline update" \
-      -- $main_pipeline )
-  else
-    # Find only the digest bumps
-    changes=$( git log main --reverse --pretty=%h --regexp-ignore-case \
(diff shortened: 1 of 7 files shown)


hack/patch-release-pipelines.sh
```diff
@@ -32,41 +32,47 @@ fi
...
-  fi
+  for comp in ${components[@]}; do
+    main_pipeline=".tekton/${comp}-${main_release_name}-$p.yaml"
+    release_pipeline=".tekton/${comp}-${release_name/./}-$p.yaml"
 
-  # Loop over each commit
-  for sha in $changes; do
-    echo "Applying changes from commit $(git log -n1 --pretty="'%h %s'" $sha)"\
-      "to pipeline definition file '$release_pipeline'"
+    if [[ "$digest_bumps" != "digest_bumps" ]]; then
```
zregvart, line 45: (comment) I'm a bit worried that when the commit messages change we'll miss some commits here, something to keep in mind is to keep updating those messages for `--grep` below

zregvart (approved):
I didn't really focus on reviewing the YAML files, the build passed so that what's important.

We get PRs like #1372, do we need the `digest_bumps` if those keep flowing?

simonbaird:
Notes:
* I thought the `*.alpha.yaml` files would only be present in the corresponding release branch..?
* The file name is long, just wondering if "task" would be enough of a distinguisher?

hack/patch-release-pipelines.sh
simonbaird: Yeah, a more robust detection of the different commits would be nice. Not a blocker for this PR of course.

hack/patch-release-pipelines.sh
```diff
@@ -32,41 +32,47 @@ fi
...
+      # Stage the changes
+      git add $release_pipeline
 
-  # Let's keep the main branch pipeline for now since it is useful to diff against
-  #git rm $main_pipeline
+    done
+
+  done
+    # Let's keep the main branch pipeline for now since it is useful to diff against
+    #git rm $main_pipeline
```
simonbaird, line 75: Even though it's commented out, I think maybe this belongs inside the "done" on line 73.

simonbaird (approved):

joejstuart:
@simonbaird @zregvart I've made some changes since you approved, if you wanna take another look before I merge. https://github.com/enterprise-contract/ec-cli/pull/1377/commits/abd81cebad761159751ec43c0a8cca1ea4ed4f0a

PR #1388, opened 2024-03-05
Update version format for Konflux builds

simonbaird:
For example, in release branch `release-0.2`, use `v0.2.<patch>` for the version in Konflux builds, instead of `0.2-<build-number>-<sha>`.

Also bump version from 0.1 to 0.2.

Ref: [EC-438](https://issues.redhat.com/browse/EC-438)


simonbaird:
Let's merge.

PR #1398, opened 2024-03-05
Keep patching pipelines if one patch doesn't apply

simonbaird:
When using the patch pipeline script for the v0.2 release branch, one of the patches, specifically the change from f1a1972, didn't apply. The patch program interactively asked by first if I should "reverse" it (no), then if I should apply it anyway (no).

The behavior is what we want, but it returns an error code, so the script aborts. Make it so the script keeps going in that kind of situation.

It does mean you have to pay attention and look at why and how the patch doesn't apply.

### DIFF
--- a/hack/patch-release-pipelines.sh
+++ b/hack/patch-release-pipelines.sh
@@ -63,7 +63,8 @@ for p in pull-request push; do
 
     # Create a diff file and apply the patch
     # (Can't use git apply since it is a different file)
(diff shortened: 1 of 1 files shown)


simonbaird:
Base branch is correct - I do want to merge to main branch. (I did use this in the v0.2 release branch but I didn't make a commit there.)

joejstuart (approved):

PR #1402, opened 2024-03-05
Bring back preflight check

simonbaird:
Ref: [EC-464](https://issues.redhat.com/browse/EC-464)


simonbaird:
Check is passing now. I'll move the PR out of draft.

simonbaird:
I got some :button: emojis on slack. Let's merge.

simonbaird:
One note: If this does somehow cause problems, it won't impact Konflux prod, since Konflux prod uses the upstream ec image build.

PR #1404, opened 2024-03-06
Add new release branch script for copying task definition pipeline

simonbaird:
Also includes some improvements for the existing script for patching the cli pipeline.

PR #1399, opened 2024-03-05
Apply customizations to Konflux pipelines for release-v0.2

simonbaird:
Apply changes to the Konflux generated default pipelines.

Note there was some manual intervention here. The patch from f1a1972 did not apply since those params were also removed in the release branch pipeline.

(Commit created with hack/patch-release-pipelines.sh )

Ref: [EC-434](https://issues.redhat.com/browse/EC-434)

PR #1411, opened 2024-03-08
Patch pipelines for release v0.2

simonbaird:
FYI I scrapped the original upstream/release-v0.2, so this is like a redo of the #1399 PR with the the Konflux generated pipeline from #1391 included also.

Why? Various reasons, largely related to #1402 and #1404.

Ref: [EC-434](https://issues.redhat.com/browse/EC-434)


simonbaird:
Not much to review here. I'll probably merge once it goes green.

PR #1412, opened 2024-03-08
Explain how to create task definition component

simonbaird:
Some day we'll automate this, but for now we describe the procedure for creating it manually.

Ref: https://issues.redhat.com/browse/EC-472


simonbaird:
As per slack discussions, the component was dropped and recreated. Out theory is that the procedure described here is good enough, but I don't want to redo it all now to double check. Let's assume the best and see how we go next time.

PR #1413, opened 2024-03-08
Fix release script when main pipeline isn't there

simonbaird:
There are cases where you might want to run the patch pipelines script more than once, e.g. to apply some digest bumps after the main customizations have been done.

Now that we're removing the main branch pipelines in the release branch, we get an error running it a second time. This patch fixes that error.


simonbaird:
CI should not matter here. Insta-merging.

PR #1419, opened 2024-03-08
Use source_url not repo_url in git clone task

simonbaird:
I'm not sure how or where to confirm it, but the new pipeline created by Konflux uses source_url instead of repo_url, so I'm assuming that is the preferred way to do it, the repo_url is deprecated.

Minor followup fix while getting the Konflux pipelines ready in the v0.2 release branch in...


simonbaird:
Optimistic merge.

PR #1428, opened 2024-03-11
Only update on a change task definition

joejstuart:

PR #1429, opened 2024-03-12
Avoid unwanted task bundle pipeline runs

simonbaird:


simonbaird:
I think #1428 is better, so let's abandon this one.

simonbaird:
Brought it back via #1523 .

PR #1433, opened 2024-03-12
Change how patch number is derived

simonbaird:
For some reason the git merge-base --fork-point isn't working as expected now. Rather than troubleshoot that I'd rather just switch to a tag based system.

(This will need to to be back ported to main branch later.)

Ref: [EC-394](https://issues.redhat.com/browse/EC-394)


simonbaird:
It's a little hard to know if it works until it's merged, so I'll probably merge as soon as the CI is done.

PR #1435, opened 2024-03-12
Add ec.sigstore.verify* rego functions

lcarva:
This commit introduces two new rego functions.

`ec.sigstore.verify_image` and `ec.sigstore.verify_attestation` can be used to verify the signatures and attestations on an image.

This version of the functions does not return the matched signatures and attestations. This is important and will be done in a follow up commit. Both functions return an verification object which can be enhanced to return these values without breaking users.

Acceptance tests will also be added in a follow up commit.

Ref: EC-447


zregvart (approved):
:partying_face:

PR #1438, opened 2024-03-13
Add `ec validate policy` to validate policyConfig

robnester-rh:
This commit adds the `ec validate policy` subcommand which specifically validates a policy config (yaml, json, or inline) against the schema generated from the Enteprise Contract Controller Policy Spec dependency used by ec.

### DIFF
--- a/cmd/validate/policy.go
+++ b/cmd/validate/policy.go
@@ -1,0 +1,94 @@
+// Copyright The Enterprise Contract Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


cmd/validate/policy.go
```diff
@@ -0,0 +1,94 @@
...
+
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// Policy conforms to the schema.
+			ctx := cmd.Context()
+			err := validate(ctx, data.policyConfiguration)
+			if err != nil {
+				return fmt.Errorf("policy configuration does not conform to the EnterpriseContractPolicy spec")
+			}
+			fmt.Println("Policy configuration conforms to the EnterpriseContractPolicy spec")
```
zregvart, line 78: We want to use the writer configured for the command here

```suggestion
			fmt.Fprintln(cmd.OutOrStdout(), "Policy configuration conforms to the EnterpriseContractPolicy spec")
```

zregvart (commented):
LGTM, one nit...

cmd/validate/policy.go
robnester-rh: Ack. Updated

PR #1446, opened 2024-03-15
Propagate sigstore opts to rego data

lcarva:
Ref: EC-447


zregvart (approved):
LGTM, not sure about the `ConfigProvider` as a name, can't think of anything better though; `PolicyConfig` `PolicyData`...

lcarva:
> LGTM, not sure about the `ConfigProvider` as a name, can't think of anything better though; `PolicyConfig` `PolicyData`...

Yeah. I don't love it either. Let's merge it as is. It should be trivial to rename it later when we're wiser.

PR #1447, opened 2024-03-18
Use schema from the api

zregvart:
We do not want to maintain two copies of schema generation logic. This
builds on PR#286[1] to use the schema generated in the controller/api,
instead of generating the schema at runtime.

[1] enterprise-contract/enterprise-contract-controller#286

### DIFF
--- a/cmd/validate/policy.go
+++ b/cmd/validate/policy.go
@@ -24,7 +24,6 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/spf13/cobra"
 
-	"github.com/enterprise-contract/ec-cli/internal/policy"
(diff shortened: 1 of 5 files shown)


zregvart:
This needs enterprise-contract/enterprise-contract-controller#286 to be merged and the version to be bumped here before merging.

lcarva (approved):
LGTM! (Once the ecp reference is properly updated)

It would be great to add some kind of test to ensure the issue this is fixing is actually fixed. (Or maybe that should go in the controller repo?) Not a blocker for this PR.

robnester-rh (approved):
LGTM!

zregvart:
This is no longer using the version from my fork, merging...

PR #1453, opened 2024-03-19
Support tracking git references in `track bundle`

zregvart:
Also contains some refactoring in separate commits.

### DIFF
--- a/cmd/track/track_bundle.go
+++ b/cmd/track/track_bundle.go
@@ -35,6 +35,7 @@ type pushImageFn func(context.Context, string, []byte, string) error
 func trackBundleCmd(track trackBundleFn, pullImage pullImageFn, pushImage pushImageFn) *cobra.Command {
 	var params = struct {
 		bundles []string
+		gits    []string
 		input   string
 		prune   bool
 		replace bool
@@ -122,7 +123,9 @@ func trackBundleCmd(track trackBundleFn, pullImage pullImageFn, pushImage pushIm
 				return err
 			}
 
-			out, err := track(cmd.Context(), params.bundles, data, params.prune, params.freshen)
+			urls := append(params.bundles, params.gits...)
+
+			out, err := track(cmd.Context(), urls, data, params.prune, params.freshen)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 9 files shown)


internal/tracker/tracker.go
```diff
@@ -182,15 +223,30 @@ func Track(ctx context.Context, urls []string, input []byte, prune bool, freshen
...
+}
 
-	t.setExpiration()
+func trackGitReferences(_ context.Context, urls []string, t *Tracker) error {
+	effective_on := effectiveOn()
 
-	if err := t.convertToOldFormat(); err != nil {
-		return nil, err
+	for _, u := range urls {
+		repository, rest, found := strings.Cut(u, "//")
```
joejstuart, line 233: Does this mean we have to specify `git+ssh` or `git+https` for task references (used by git resolver) when calling the `ec track`  command?

lcarva: https://github.com/enterprise-contract/ec-policies/blob/main/example/data/trusted_tekton_tasks.yml#L3

cmd/track/track_bundle.go
```diff
@@ -170,7 +166,10 @@ func trackBundleCmd(track trackBundleFn, pullImage pullImageFn, pushImage pushIm
 	cmd.Flags().StringVarP(&params.input, "input", "i", params.input, "existing tracking file")
 
 	cmd.Flags().StringSliceVarP(&params.bundles, "bundle", "b", params.bundles,
-		"bundle image reference to track - may be used multiple times (required)")
+		"bundle image reference to track - may be used multiple times (required when --git not provided)")
```
lcarva, line 169: Is this right? It's required if neither `--git` nor `--input` are provided. Maybe we should just remove the "required" text from this description.

internal/tracker/tracker.go
```diff
@@ -154,11 +154,52 @@ func Track(ctx context.Context, urls []string, input []byte, prune bool, freshen
...
+			gits = append(gits, u)
+		} else {
+			imgs = append(imgs, u)
+		}
+	}
+
+	return imgs, gits
+}
+
+func trackImageReferences(ctx context.Context, urls []string, t *Tracker, freshen bool) error {
```
lcarva, line 192: NIT: Why not make this a method on the `Tracker` type? (Same for `trackGitReferences`)

lcarva (approved):

internal/tracker/tracker.go
zregvart: Interesting, when handling on the policy side we might ignore whether the scheme is `git+ssh` or `git+https` or `git+...`, we might not care at all about the transfer protocol...

zregvart:
Nits were easy to address so merging this now.

PR #1459, opened 2024-03-21
Refactor to remove repetition in policy sources

robnester-rh:
This commit refactors the code to remove the fetching of policy sources as a part of the loop for source groups.

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -34,11 +34,14 @@ import (
 	"github.com/enterprise-contract/ec-cli/internal/format"
 	"github.com/enterprise-contract/ec-cli/internal/output"
 	"github.com/enterprise-contract/ec-cli/internal/policy"
+	"github.com/enterprise-contract/ec-cli/internal/policy/source"
 	"github.com/enterprise-contract/ec-cli/internal/utils"
 	validate_utils "github.com/enterprise-contract/ec-cli/internal/validate"
 )
 
-type imageValidationFunc func(context.Context, app.SnapshotComponent, policy.Policy, bool) (*output.Output, error)
+type imageValidationFunc func(context.Context, app.SnapshotComponent, policy.Policy, []evaluator.Evaluator, bool) (*output.Output, error)
+
+var newConftestEvaluator = evaluator.NewConftestEvaluator
 
 func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 	var data = struct {
@@ -225,6 +228,30 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			}
 
 			appComponents := data.spec.Components
+			evaluators := []evaluator.Evaluator{}
+
+			// Return an evaluator for each of these
+			for _, sourceGroup := range data.policy.Spec().Sources {
+				// Todo: Make each fetch run concurrently
+				log.Debugf("Fetching policy source group '%s'", sourceGroup.Name)
+				policySources, err := source.FetchPolicySources(sourceGroup)
+				if err != nil {
+					log.Debugf("Failed to fetch policy source group '%s'!", sourceGroup.Name)
+					return err
+				}
+
+				for _, policySource := range policySources {
+					log.Debugf("policySource: %#v", policySource)
+				}
+
+				c, err := newConftestEvaluator(cmd.Context(), policySources, data.policy, sourceGroup)
+				if err != nil {
+					log.Debug("Failed to initialize the conftest evaluator!")
+					return err
+				}
+
+				evaluators = append(evaluators, c)
+			}
 
 			// worker is responsible for processing one component at a time from the jobs channel,
 			// and for emitting a corresponding result for the component on the results channel.
@@ -233,7 +260,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 				for comp := range jobs {
 					log.Debugf("Worker %d got a component %q", id, comp.ContainerImage)
 					ctx := cmd.Context()
-					out, err := validate(ctx, comp, data.policy, data.info)
+					out, err := validate(ctx, comp, data.policy, evaluators, data.info)
 					res := result{
 						err: err,
 						component: applicationsnapshot.Component{
--- a/cmd/validate/image_test.go
+++ b/cmd/validate/image_test.go
@@ -250,7 +250,7 @@ func Test_determineInputSpec(t *testing.T) {
 }
 
 func Test_ValidateImageCommand(t *testing.T) {
-	validate := func(_ context.Context, component app.SnapshotComponent, _ policy.Policy, _ bool) (*output.Output, error) {
+	validate := func(_ context.Context, component app.SnapshotComponent, _ policy.Policy, _ []evaluator.Evaluator, _ bool) (*output.Output, error) {
 		return &output.Output{
 			ImageSignatureCheck: output.VerificationStatus{
 				Passed: true,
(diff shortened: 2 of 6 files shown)


robnester-rh:
Previously we created our `policySources` array while looping through our `sourceGroups`. That resulted in a lot of unnecessary downloading of the same data. This commit creates an initial `policySources` array in the `RunE` func in the `cmd` and then appends `ruleData` on a per-`sourceGroup` basis. This commit should eliminate the repetitious downloading and should help to reduce network traffic and resource utilization.

zregvart:
Not sure, feels wrong, this would collapse all policy sources from all source groups into one group, or am I reading this wrongly? Do we want to optimize so that the same policy source is not fetched more than once? That could be done by optimizing the individual `GetPolicy` functions, by looking if the policy has already been fetched on disk. I think that could be a much smaller change overall.

robnester-rh:
I’ll look at that approach and update this pr. Thank you.

lcarva:
> That could be done by optimizing the individual GetPolicy functions, by looking if the policy has already been fetched on disk.

Be wary that policy sources are currently fetched in parallel for each container-image/input.

zregvart:
> > That could be done by optimizing the individual GetPolicy functions, by looking if the policy has already been fetched on disk.
> 
> Be wary that policy sources are currently fetched in parallel for each container-image/input.

I’m doing a `sync.Map` with `sync.OnceValues` for values [here](https://github.com/zregvart/ec-cli/commit/029067b424379c27810c9fdfe5e93ec10d646e6d#diff-706edaa24153c20c6964b3b6ad23f48a40b976a64194b7d860e03d52543a1460R87). I think that should be fairly safe approach. Perhaps there is a package/library with this already done, like a caching library, perhaps that’s worth a look…

robnester-rh:
As we discussed in a recent stand-up, we have two simultaneous desires to reduce repetition. One is the repetition of evaluator creation and the other is the repetition of downloading policy sources. The most recent push attempts to consolidate the creation of the evaluators at the `validate image` command level, providing the evaluators to the subsequent validation calls.

simonbaird:
Needs a Jira ref also.

cmd/validate/image.go
```diff
@@ -226,14 +231,59 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			appComponents := data.spec.Components
 
+			var wg sync.WaitGroup
+			errChan := make(chan error, len(data.policy.Spec().Sources))
+			evaluatorsChan := make(chan *evaluator.Evaluator, len(data.policy.Spec().Sources))
+			for _, sourceGroup := range data.policy.Spec().Sources {
+				wg.Add(1)
+				go func(sourceGroup v1alpha1.Source) {
```
lcarva, line 239: Likely to have a less impact here, but blindly starting an unbounded number of goroutines is problematic. See the worker/jobs pattern used below for an alternative.

Also, there's probably not a whole lot of value in fetching multiple source groups in parallel.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -107,29 +104,6 @@ func NewApplicationSnapshotImage(ctx context.Context, component app.SnapshotComp
 		return nil, err
 	}
 
-	// Return an evaluator for each of these
```
lcarva, line 110: Removal of this block, also means we can probably remove the `p policy.Policy` parameter from this fucntion.

lcarva (commented):
This seems reasonable.

internal/image/validate.go
```diff
@@ -107,9 +107,11 @@ func ValidateImage(ctx context.Context, comp app.SnapshotComponent, p policy.Pol
 	}
 
 	var allResults []evaluator.Outcome
-	for _, e := range a.Evaluators {
+
+	for _, e := range evaluators {
 		// Todo maybe: Handle each one concurrently
 		results, data, err := e.Evaluate(ctx, []string{inputPath})
+		log.Debug("\n\nRunning conftest policy check\n\n")
```
zregvart, line 114: Nit, this feels a bit noisy without adding any information useful for debugging, perhaps:

```suggestion
	for i, e := range evaluators {
		// Todo maybe: Handle each one concurrently
		log.Debugf("Running evaluator %d over %q", i, inputPath)
		results, data, err := e.Evaluate(ctx, []string{inputPath})
```

internal/image/validate_test.go
zregvart, line 1: Perhaps renaming the `TestValidateImage` to something like `TestBuiltinChecks`

cmd/validate/image_test.go
zregvart, line 1: We also need some coverage on the `RunE`s creation of evaluators

internal/policy/source/source.go
```diff
@@ -65,12 +67,30 @@ type PolicyUrl struct {
...
 	sourceUrl := p.PolicyUrl()
 
 	dest := uniqueDestination(workDir, p.Subdir(), sourceUrl)
 
+	// Check if the policy has already been downloaded.
+	// If it has try to create a symlink. If that passes return the symlink path.
+	// If the symlink creation fails, fall back to downloading the policy again.
+	if d, loaded := downloadCache.LoadOrStore(sourceUrl, dest); loaded {
+		log.Debugf("Download cache hit: %s", sourceUrl)
+		fs := new(afero.OsFs)
```
zregvart, line 84: Doing this helps with tests (using [utils](https://github.com/enterprise-contract/ec-cli/blob/01b6c39ad710b0c6d8e3a70885557c36343e0db5/internal/utils/helpers.go#L94)):

```suggestion
		fs := utils.FS(ctx)
```

internal/policy/source/source.go
```diff
@@ -65,12 +67,30 @@ type PolicyUrl struct {
...
+		err := fs.SymlinkIfPossible(d.(string), dest)
+		if err == nil {
```
zregvart, line 86: I guess it is intentional to re-download when symlink fails? Perhaps we should log the `err` so we understand why the source was re-downloaded?

internal/policy/source/source.go
```diff
@@ -65,12 +67,30 @@ type PolicyUrl struct {
...
 // GetPolicies clones the repository for a given PolicyUrl
 func (p *PolicyUrl) GetPolicy(ctx context.Context, workDir string, showMsg bool) (string, error) {
 	sourceUrl := p.PolicyUrl()
 
 	dest := uniqueDestination(workDir, p.Subdir(), sourceUrl)
 
+	// Check if the policy has already been downloaded.
+	// If it has try to create a symlink. If that passes return the symlink path.
+	// If the symlink creation fails, fall back to downloading the policy again.
+	if d, loaded := downloadCache.LoadOrStore(sourceUrl, dest); loaded {
```
zregvart, line 82: Having a `sourceUrl` key in `downloadCache` does not guarantee that the source was indeed downloaded, the download could have failed or could be still in progress.
I think the value in `downloadCache` should be `sync.OnceValues`.

Perhaps something like:

```suggestion
	dl, _ := downloadCache.LoadOrStore(sourceUrl, sync.OnceValues(func() (string, error) {
		dest := uniqueDestination(workDir, p.Subdir(), sourceUrl)
		log.Debugf("Download cache miss: %s", sourceUrl)
	
		// Checkout policy repo into work directory.
		log.Debugf("Downloading policy files from source url %s to destination %s", sourceUrl, dest)
	
		x := ctx.Value(DownloaderFuncKey)
		if dl, ok := x.(downloaderFunc); ok {
			return dest, dl.Download(ctx, dest, sourceUrl, showMsg)
		}
		return dest, downloader.Download(ctx, dest, sourceUrl, showMsg)
	}))
	
	if dest, err := dl.(func() (string, error))(); err != nil {
		return "", err
	}
	
	if filepath.Dir(dest) != workDir {
		fs := utils.FS(ctx)
		workdirDest := uniqueDestination(workDir, p.Subdir(), sourceUrl)
		if err := fs.SymlinkIfPossible(dest, workdirDest); err != nil {
			return "", err
		}
		
		return workdirDest, nil
	}
	
	return dest, nil
```

zregvart (approved):

PR #1472, opened 2024-03-25
Consolidate build process across each Dockerfile

joejstuart:
Each Dockerfile has a different way of building ec. This moves the build to the Dockerfile and also pulls the cosign version from the mod file.

https://issues.redhat.com/browse/EC-470

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,6 @@
 
 FROM docker.io/library/golang:1.21 AS build
 
-ARG BUILD_LIST="darwin_amd64 darwin_arm64 linux_amd64 linux_arm64 linux_ppc64le linux_s390x windows_amd64"
 ARG TARGETOS
 ARG TARGETARCH
 
@@ -31,7 +30,7 @@ RUN go mod download
 # Now copy everything including .git
 COPY . .
 
-RUN ["/usr/bin/bash", "-c", "/build/build.sh \"${BUILD_LIST}\" \"${TARGETOS}\" \"${TARGETARCH}\""]
+RUN /build/build.sh "${TARGETOS}_${TARGETARCH}"
 
 # Extract this so we can download the matching cosign version below
 RUN go list --mod=readonly -f '{{.Version}}' -m github.com/sigstore/cosign/v2 | tee cosign_version.txt
@@ -64,8 +63,6 @@ COPY --from=download /download/cosign /usr/bin/cosign
 RUN cosign version
 
 RUN microdnf -y install git-core jq && microdnf clean all
-
-COPY --from=build /build/dist/ /usr/bin/
 
 # Copy the one ec binary that can run in this container
 COPY --from=build "/build/dist/ec_${TARGETOS}_${TARGETARCH}" /usr/bin/ec
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,31 +14,60 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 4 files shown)


joejstuart:
If this is an acceptable way to consolidate the container build process, there is cleanup to do in the Makefile which I'll get to.

Dockerfile
```diff
@@ -14,31 +14,84 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.3@sha256:bc552efb4966aaa44b02532be3168ac1ff18e2af299d0fe89502a1d9fabafbc5 AS downloads
+## Build
+
+FROM docker.io/library/golang:1.21 AS build
+
+ARG BUILD_LIST="darwin_amd64 darwin_arm64 linux_amd64 linux_arm64 linux_ppc64le linux_s390x windows_amd64"
```
zregvart, line 21: Is this used? I can see `build.sh` redefines it.

Dockerfile
```diff
@@ -14,31 +14,60 @@
...
+WORKDIR /build
+
+# Copy just the mod file for better layer caching when building locally
+COPY go.mod go.sum .
+RUN go mod download
+
+# Now copy everything including .git
+COPY . .
+
+RUN ["/usr/bin/bash", "-c", "/build/build.sh \"${BUILD_LIST}\" \"${TARGETOS}\" \"${TARGETARCH}\""]
```
lcarva, line 34: Hmm why not just:

```
RUN /build/build.sh "${BUILD_LIST}" "${TARGETOS}" "${TARGETARCH}"
```

lcarva (commented):
I don't really the idea of stuffing a bunch of different binaries for different OSes and arches into a single image. There's a very specific reason we do this in Dockerfile.dst which is to support the downloads workflow in OpenShift. This really doesn't belong anywhere else. It's confusing, ugly, and makes the image unnecessarily large.

Would it be possible to isolate the "downloader" bits so it can be an addon to Dockerfile.dst and not Dockerfile?

simonbaird (commented):
I wanna try it out locally, give me a little more time.

Dockerfile
joejstuart: The `RUN` command defaults to `/bin/sh`. I think I can change the default shell though so that isn't so messy.

joejstuart:
> I don't really the idea of stuffing a bunch of different binaries for different OSes and arches into a single image. There's a very specific reason we do this in Dockerfile.dst which is to support the downloads workflow in OpenShift. This really doesn't belong anywhere else. It's confusing, ugly, and makes the image unnecessarily large.
> 
> Would it be possible to isolate the "downloader" bits so it can be an addon to Dockerfile.dst and not Dockerfile?

The story is about making the two Dockerfile's more alike which is why the ARGS are the same. The build.sh script only built all the binaries if a `release-*` branch was checked out though. This commit should make things more clear. https://github.com/enterprise-contract/ec-cli/pull/1472/commits/ebf139f65c40ed639eb46e3589aaf6f0ed5448d6

Dockerfile
lcarva: Why do you need bash for this simple command?

`build/build.sh` itself should run with what is specified in that file which is `/usr/bin/bash`.

joejstuart: and you're right. thank you for that. I'll remove the SHELL

simonbaird (approved):

PR #1474, opened 2024-03-26
`validate image`: Allow custom Rego variables with new `--extra` flag

mbestavros:
This PR adds a new option, `--extra`, to the `ec validate image` command. This argument allows users to inject additional variables into the Rego rule data per source. Syntax is simple: `--extra key=value`. The flag can also be used multiple times.

This is an evolution of the idea that @lcarva proposed in https://github.com/enterprise-contract/ec-cli/pull/1277.

Here's an example output from `ec validate image` with `--extra var=value`:

```yaml
policy:
  description: Includes rules for levels 1, 2 & 3 of SLSA v0.1. This is the default
    config used for new Konflux applications. Available collections are defined in
    https://redhat-appstudio.github.io/docs.stonesoup.io/ec-policies/release_policy.html#_available_rule_collections.
    If a different policy configuration is desired, this resource can serve as a starting
    point. See the docs on how to include and exclude rules https://redhat-appstudio.github.io/docs.stonesoup.io/ec-policies/policy_configuration.html#_including_and_excluding_rules.
  name: Default
  publicKey: /dev/fd/63
  sources:
  - config:
      include:
      - '@slsa3'
    data:
    - oci::quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest
    - github.com/release-engineering/rhtap-ec-policy//data
    name: Default
    policy:
    ruleData:
      var: value
success: true
```

This still needs tests added, but it appears to be functional, as seen above.

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -18,9 +18,11 @@ package validate
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 
 	hd "github.com/MakeNowJust/heredoc"
 	"github.com/hashicorp/go-multierror"
@@ -28,6 +30,7 @@ import (
 	"github.com/sigstore/cosign/v2/pkg/cosign"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 
 	"github.com/enterprise-contract/ec-cli/internal/applicationsnapshot"
 	"github.com/enterprise-contract/ec-cli/internal/evaluator"
@@ -50,6 +53,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		certificateOIDCIssuer       string
 		certificateOIDCIssuerRegExp string
 		effectiveTime               string
+		extraRuleData               []string
 		filePath                    string // Deprecated: images replaced this
 		imageRef                    string
 		info                        bool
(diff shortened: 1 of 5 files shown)


cmd/validate/image.go
```diff
@@ -393,6 +444,10 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		a RFC3339 formatted value, e.g. 2022-11-18T00:00:00Z.
 	`))
 
+	cmd.Flags().StringSliceVar(&data.extra, "extra", data.extra, hd.Doc(`
```
simonbaird, line 447: The name "extra" feels not quite specific enough. Something like this perhaps:

```suggestion
	cmd.Flags().StringSliceVarP(&data.extraRuleData, "extra-rule-data", "e", data.extraRuleData, hd.Doc(`
```

simonbaird:
Let's get some test coverage.

simonbaird:
(Wrong button sorry..)

cmd/validate/image.go
mbestavros: Changed it to your suggestion, but open to other ideas as well. I know we floated `override` or `override-rule-data` as well.

mbestavros:
@simonbaird I added a unit and acceptance test for the positive case. Had to get a bit hacky for the unit test to work correctly.

cmd/validate/image_test.go
```diff
@@ -673,6 +674,104 @@ configuration:
 	assert.NoError(t, err)
 }
 
+func Test_ValidateImageCommandExtraData(t *testing.T) {
+	validate := func(_ context.Context, component app.SnapshotComponent, _ policy.Policy, _ bool) (*output.Output, error) {
```
zregvart, line 678: There is a compile error on this line:
```suggestion
	validate := func(_ context.Context, component app.SnapshotComponent, _ policy.Policy, _ []evaluator.Evaluator, _ bool) (*output.Output, error) {
```

lcarva: It looks like @mbestavros did a rebase that brought in the change in the signature for the validate function. I have questions...

- The only CI check failing right now is the linter. How did "Checks / Test" pass?!
- Why was the linter complaining about this line before when it was obviously correct at the time?

simonbaird (approved):

PR #1475, opened 2024-03-26
Update release v0.2 branch with newer content from main

simonbaird:
Done with a merge because we've done some builds already from release branch so we need to preserve the history.

Ref: [EC-504](https://issues.redhat.com/browse/EC-504)


simonbaird:
This PR is to merge a merge commit. Actually the only commit in the PR is the merge commit which is (currently) cafbf77.

simonbaird:
We're going to end up with a patch number 45 or more...

```
$ git rev-list --merges --count v0.2.0..HEAD
45
```

We could figure out a way to bring that down, but I don't think it's worth the trouble.

simonbaird:
Note to self: Don't use "release-" prefix on PR branches.. Because of that prefix the pre-merge build is compiling all the different arches, and the version number is bogus. No harm done but better to remember for next time.

Issue #1465, opened 2024-03-22
Add tests for the antora extensions


lcarva:
Let's make sure we have some basic coverage for the code defined in the antora extensions in this repo.

**Acceptance Criteria**

* Write basic tests to verify the functionality of the antora extensions in this repo.
* These tests should run as part of the CI in this repo. (If this turns out to be too much of an effort, consider splitting this item into another issue.)

zregvart:
Closed in favor of EC-513

PR #1479, opened 2024-03-27
Add tests for the reference-antora-extension

zregvart:
Adds tests that rely on snapshots for the Asciidoc files generated in the reference-antora-extension. The snapshots will need to be updated each time a change is done to the CLI documentation.

Resolves: #1465


zregvart:
Let's close and work on EC-513

PR #1481, opened 2024-03-28
Support for mocking builtin functions

zregvart:
Adds a new builtin: `ec.testing.mock` that can be used only when invoked via `ec opa test` command. When invoked with a name of the function, arguments and the mocked return value, that information is stored in the private static map within the `internal/rego/testing` package, and supporting builtins can check for a short circuit mocked value return by invoking the `testing.Mocked` function.

### DIFF
--- a/acceptance/examples/mock_test.rego
+++ b/acceptance/examples/mock_test.rego
@@ -1,0 +1,9 @@
+package mock_test
+
+import rego.v1
+
+test_mocking_ec_oci_blob if {
+    ec.test.mock("ec.oci.blob", ["ref"], "wubba lubba dub dub")
+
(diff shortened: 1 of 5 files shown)


zregvart:
Created as a draft, to gather early feedback. I'm not 100% sure on the use of the `ast.Value.Hash()`, I've seen one other place in OPA (with annotations) where it is used as a key to the map.

Only `ec.oci.blob` has support for mocking. The acceptance test should provide a good overview how this can be used to create unit tests in ec-policies for code that uses builtins.

lcarva:
I don't think this is needed. The incantation in rego should be sufficient to mock whatever we need when running unit tests.

zregvart:
TIL `with` can mock builtins. Thanks!

PR #1482, opened 2024-03-28
Implement support for multiple output formats in ec test

mbestavros:
This PR revamps the `--output` flag in `ec test` to accept `format=file` syntax. This will allow multiple output formats to different files on disk.

For example: `--output json=test.json --output appstudio=appstudio.txt` will produce two files, `test.json` and `appstudio.txt`, in JSON and Appstudio format, respectively.

Ref: https://issues.redhat.com/browse/EC-55

### DIFF
--- a/cmd/test/test.go
+++ b/cmd/test/test.go
@@ -26,6 +26,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/open-policy-agent/conftest/output"
 	"github.com/open-policy-agent/conftest/parser"
@@ -117,7 +118,6 @@ func newTestCommand() *cobra.Command {
 				"no-color",
 				"no-fail",
 				"suppress-exceptions",
-				"output",
 				"file",
 				"parser",
 				"policy",
@@ -151,9 +151,12 @@ func newTestCommand() *cobra.Command {
 				return fmt.Errorf("unmarshal parameters: %w", err)
 			}
 
-			outputFilePath, err := cmd.Flags().GetString("file")
+			outputFormats, err := cmd.Flags().GetStringSlice("output")
 			if err != nil {
 				return fmt.Errorf("reading flag: %w", err)
+			}
+			if len(outputFormats) == 0 {
+				outputFormats = []string{output.OutputStandard}
 			}
 
 			results, resultsErr := runner.Run(ctx, fileList)
@@ -165,60 +168,66 @@ func newTestCommand() *cobra.Command {
 			}
 
 			if !runner.Quiet || exitCode != 0 {
-				if runner.Output == OutputAppstudio {
-					// The appstudio format is unknown to Conftest so we handle it ourselves
+				for _, outputAndFormat := range outputFormats {
+					parts := strings.SplitN(outputAndFormat, "=", 2)
 
-					if resultsErr != nil {
-						return appstudioErrorHandler(runner.NoFail, "running test", resultsErr)
-					}
-
-					report := appstudioReport(results, runner.Namespace)
-					reportOutput, err := json.Marshal(report)
-					if err != nil {
-						return appstudioErrorHandler(runner.NoFail, "output results", err)
-					}
-
-					if outputFilePath != "" {
-						err := os.WriteFile(outputFilePath, reportOutput, 0600)
-						if err != nil {
-							return fmt.Errorf("creating output file: %w", err)
+					format := output.OutputStandard
+					outputFilePath := ""
+					if len(parts) > 0 {
+						format = parts[0]
+						if len(parts) == 2 {
+							outputFilePath = parts[1]
 						}
 					}
-					fmt.Fprintln(cmd.OutOrStdout(), string(reportOutput)+"\n")
 
-				} else {
-					// Conftest handles the output
+					if format == OutputAppstudio {
+						// The appstudio format is unknown to Conftest so we handle it ourselves
 
-					if resultsErr != nil {
-						return fmt.Errorf("running test: %w", resultsErr)
-					}
-
-					var outputFile *os.File
-					if outputFilePath != "" {
-						outputFile, err = os.Create(outputFilePath)
-						if err != nil {
-							return fmt.Errorf("creating output file: %w", err)
+						if resultsErr != nil {
+							return appstudioErrorHandler(runner.NoFail, "running test", resultsErr)
 						}
-						defer outputFile.Close()
-					}
 
-					outputter := output.Get(runner.Output, output.Options{
-						NoColor:            runner.NoColor,
-						SuppressExceptions: runner.SuppressExceptions,
-						Tracing:            runner.Trace,
-						JUnitHideMessage:   viper.GetBool("junit-hide-message"),
-						File:               outputFile,
-					})
-					if err := outputter.Output(results); err != nil {
-						return fmt.Errorf("output results: %w", err)
-					}
-
-					if outputFilePath != "" {
-						contents, err := os.ReadFile(outputFile.Name())
+						report := appstudioReport(results, runner.Namespace)
+						reportOutput, err := json.Marshal(report)
 						if err != nil {
-							return fmt.Errorf("copying output file to stdout: %w", err)
+							return appstudioErrorHandler(runner.NoFail, "output results", err)
 						}
-						fmt.Fprintln(cmd.OutOrStdout(), string(contents))
+
+						if outputFilePath != "" {
+							err := os.WriteFile(outputFilePath, reportOutput, 0600)
+							if err != nil {
+								return fmt.Errorf("creating output file: %w", err)
+							}
+						} else {
+							fmt.Fprintln(cmd.OutOrStdout(), string(reportOutput))
+						}
+
+					} else {
+						// Conftest handles the output
+
+						if resultsErr != nil {
+							return fmt.Errorf("running test: %w", resultsErr)
+						}
+
+						var outputFile *os.File
+						if outputFilePath != "" {
+							outputFile, err = os.Create(outputFilePath)
+							if err != nil {
+								return fmt.Errorf("creating output file %s: %w", outputFilePath, err)
+							}
+							defer outputFile.Close()
+						}
+
+						outputter := output.Get(format, output.Options{
+							NoColor:            runner.NoColor,
+							SuppressExceptions: runner.SuppressExceptions,
+							Tracing:            runner.Trace,
+							JUnitHideMessage:   viper.GetBool("junit-hide-message"),
+							File:               outputFile,
+						})
+						if err := outputter.Output(results); err != nil {
+							return fmt.Errorf("output results: %w", err)
+						}
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


simonbaird:
Testing it like this:

Plain old conftest:

```
conftest test -p foo.rego input.json --output json
[
        {
                "filename": "input.json",
                "namespace": "main",
                "successes": 0,
                "failures": [
                        {
                                "msg": "boo"
                        }
                ]
        }
]
```

Plain old ec test (should be the same):

```
EC_EXPERIMENTAL=1 ec test -p foo.rego input.json --output json
[
        {
                "filename": "input.json",
                "namespace": "main",
                "successes": 0,
                "failures": [
                        {
                                "msg": "boo"
                        }
                ]
        }
]
```

So far so good.

simonbaird:
With appstudio format:

```
EC_EXPERIMENTAL=1 ec test -p foo.rego input.json --output appstudio
{"timestamp":"1711656172","namespace":"main","successes":0,"failures":1,"warnings":0,"result":"FAILURE","note":"Failures detected"}
```

Looks good afaict.

With multiple formats:

```
$ EC_EXPERIMENTAL=1 ec test -p foo.rego input.json --output json --output appstudio=as.json
[
        {
                "filename": "input.json",
                "namespace": "main",
                "successes": 0,
                "failures": [
                        {
                                "msg": "boo"
                        }
                ]
        }
]
{"timestamp":"1711656214","namespace":"main","successes":0,"failures":1,"warnings":0,"result":"FAILURE","note":"Failures detected"}

$ cat as.json
{"timestamp":"1711656214","namespace":"main","successes":0,"failures":1,"warnings":0,"result":"FAILURE","note":"Failures detected"}
```

I think that's buggy. It does write to `as.json`, which we want, but it seems to be writing to STDOUT as well.

simonbaird:
This is my foo.rego if you want to try to reproduce the above:

```
package main
deny["boo"] {
        true
}
```

cmd/test/test.go
```diff
@@ -165,60 +165,74 @@ func newTestCommand() *cobra.Command {
...
 
-				} else {
-					// Conftest handles the output
+						if outputFilePath != "" {
+							err := os.WriteFile(outputFilePath, reportOutput, 0600)
+							if err != nil {
+								return fmt.Errorf("creating output file: %w", err)
+							}
+						}
+						fmt.Fprintln(cmd.OutOrStdout(), string(reportOutput)+"\n")
```
simonbaird, line 200: Maybe this would do it:

```suggestion
						else {
						    fmt.Fprintln(cmd.OutOrStdout(), string(reportOutput)+"\n")
						}
```

cmd/test/test.go
```diff
@@ -165,60 +165,74 @@ func newTestCommand() *cobra.Command {
...
+							File:               outputFile,
+						})
+						if err := outputter.Output(results); err != nil {
+							return fmt.Errorf("output results: %w", err)
+						}
+
+						if outputFilePath != "" {
+							contents, err := os.ReadFile(outputFile.Name())
+							if err != nil {
+								return fmt.Errorf("copying output file to stdout: %w", err)
```
simonbaird, line 232: Hold on, I think there's a requirements misunderstanding. If the user specifies a file, then we should write the output to that file only. It looks like you're doing extra work to also print it on stdout.

simonbaird:
Doing this, I can see it writes to both stdout and the filename for all the ouputs:

`EC_EXPERIMENTAL=1 ec test -p foo.rego input.json --output json=j.json --output appstudio=as.json`

The intention iiuc is that we write to the file only. The above command should write nothing on stdout.

simonbaird:
Notice the behavior for ec. This sends nothing to stdout:

```
OUTPUT="json=out.json" hack/simple-demo.sh
```

cmd/test/test.go
mbestavros: @simonbaird Yeah, that's a leftover from the old implementation. I'll adjust.

simonbaird: I think you can just delete line 229 to 235 or so.

cmd/test/test.go
```diff
@@ -249,14 +263,15 @@ func newTestCommand() *cobra.Command {
 	cmd.Flags().String("parser", "", fmt.Sprintf("Parser to use to parse the configurations. Valid parsers: %s", parser.Parsers()))
 	cmd.Flags().String("capabilities", "", "Path to JSON file that can restrict opa functionality against a given policy. Default: all operations allowed")
 
-	cmd.Flags().StringP("output", "o", output.OutputStandard, fmt.Sprintf("Output format for conftest results - valid options are: %s", append(output.Outputs(), OutputAppstudio)))
+	//cmd.Flags().StringP("output", "o", output.OutputStandard, fmt.Sprintf("Output format for conftest results - valid options are: %s", append(output.Outputs(), OutputAppstudio)))
```
simonbaird, line 266: To be removed, right?

simonbaird:
Let's add a Jira in commit message and/or PR description as well.

mbestavros:
@simonbaird I added some updates - shouldn't print to stdout anymore, and I got rid of the commented-out `output` declaration.

cmd/test/test.go
```diff
@@ -165,60 +165,67 @@ func newTestCommand() *cobra.Command {
...
 
-				} else {
-					// Conftest handles the output
+						if outputFilePath != "" {
+							err := os.WriteFile(outputFilePath, reportOutput, 0600)
+							if err != nil {
+								return fmt.Errorf("creating output file: %w", err)
+							}
+						} else {
+							fmt.Fprintln(cmd.OutOrStdout(), string(reportOutput)+"\n")
```
simonbaird, line 200: I feel like Fprintln is going to handle the new line at the end here, so probably no need to add the extra one.

```suggestion
							fmt.Fprintln(cmd.OutOrStdout(), string(reportOutput))
```

cmd/test/test.go
```diff
@@ -165,60 +165,67 @@ func newTestCommand() *cobra.Command {
...
-						File:               outputFile,
-					})
-					if err := outputter.Output(results); err != nil {
-						return fmt.Errorf("output results: %w", err)
-					}
+						var outputFile *os.File
+						if outputFilePath != "" {
+							outputFile, err = os.Create(outputFilePath)
+							if err != nil {
+								return fmt.Errorf("creating output file: %w", err)
```
simonbaird, line 214: Maybe this will be better for debugging:

```suggestion
								return fmt.Errorf("creating output file %s: %w", outputFilePath, err)
```

simonbaird (approved):

zregvart:
The [acceptance test failure](https://github.com/enterprise-contract/ec-cli/actions/runs/8473742295/job/23218724964?pr=1482#step:5:5728) should be investigated.

simonbaird:
This reproduces the failure for me: `make feature_conftest_test`

mbestavros:
Strange... it doesn't reproduce on my machine :confused:

cmd/test/test.go
```diff
@@ -151,10 +151,13 @@ func newTestCommand() *cobra.Command {
...
 			}
 
-			outputFilePath, err := cmd.Flags().GetString("file")
+			outputFormats, err := cmd.Flags().GetStringSlice("output")
 			if err != nil {
 				return fmt.Errorf("reading flag: %w", err)
 			}
+			if len(outputFormats) == 0 {
+				outputFormats = []string{output.OutputStandard}
+			}
```
mbestavros, line 160: @simonbaird @zregvart My latest push should fix acceptance tests. The issue was that the command would no longer default to stdout if no `--output` flag was specified. I added a fallback here.

simonbaird:
There are two minor suggestions from my previous review, did you consider those?

cmd/test/test.go
```diff
@@ -165,60 +168,67 @@ func newTestCommand() *cobra.Command {
 			}
 
 			if !runner.Quiet || exitCode != 0 {
-				if runner.Output == OutputAppstudio {
-					// The appstudio format is unknown to Conftest so we handle it ourselves
+				for _, outputAndFormat := range outputFormats {
+					parts := strings.SplitN(outputAndFormat, "=", 2)
```
zregvart, line 172: `parts` can be `nil` here if there is no `=` in `outputAndFormat`, perhaps switch on `len(parts)`?

zregvart (commented):
One small issue, otherwise we should merge this...

cmd/test/test.go
simonbaird: I might be wrong, I think parts would be just `[outputAndFormat]` if there was no "=" char present, so `parts[0]` is safe to use.

simonbaird: Could still switch on `len(parts)` for extra readability though.

zregvart:
If we're saying that this is still experimental, i.e. we will drop this at some point, then let's merge this as it seems to do what was required.
If, however, we do want to keep this around for a while, then perhaps adding test coverage (acceptance tests) is needed here.

cmd/test/test.go
mbestavros: @zregvart Yep, that makes sense. I refactored that bit of code.

mbestavros:
@simonbaird @zregvart I updated with suggestions. I'm happy merging as is - happy to also write tests, but maybe we do it in a follow-up PR.

zregvart:
Make sure to run `make generate` to update the docs

docs/modules/ROOT/pages/ec_test.adoc
```diff
@@ -83,7 +83,7 @@ the output will include a detailed trace of how the policy was evaluated, e.g.
 -n, --namespace:: Test policies in a specific namespace (Default: [main])
 --no-color:: Disable color when printing (Default: false)
 --no-fail:: Return an exit code of zero even if a policy fails (Default: false)
--o, --output:: Output format for conftest results - valid options are: [stdout json tap table junit github appstudio] (Default: stdout)
+-o, --output:: Output format for conftest results - valid options are: [stdout json tap table junit github appstudio] (Default: [])
```
simonbaird, line 86: I know it's generated, but the "Default" here is not helpful. Can you think of a way to have it still say stdout?

Also, I think it would be good if the help text somehow mention the `format=filename` option. Maybe add a little extra on line 268.

cmd/test/test.go
```diff
@@ -249,14 +258,14 @@ func newTestCommand() *cobra.Command {
...
 
-	cmd.Flags().StringP("output", "o", output.OutputStandard, fmt.Sprintf("Output format for conftest results - valid options are: %s", append(output.Outputs(), OutputAppstudio)))
 	cmd.Flags().String("file", "", "File path to write output to")
 	cmd.Flags().Bool("junit-hide-message", false, "Do not include the violation message in the JUnit test name")
 
 	cmd.Flags().StringSliceP("policy", "p", []string{"policy"}, "Path to the Rego policy files directory")
 	cmd.Flags().StringSliceP("update", "u", []string{}, "A list of URLs can be provided to the update flag, which will download before the tests run")
 	cmd.Flags().StringSliceP("namespace", "n", []string{"main"}, "Test policies in a specific namespace")
 	cmd.Flags().StringSliceP("data", "d", []string{}, "A list of paths from which data for the rego policies will be recursively loaded")
+	cmd.Flags().StringSliceP("output", "o", []string{}, fmt.Sprintf("Output format for conftest results - valid options are: %s", append(output.Outputs(), OutputAppstudio)))
```
simonbaird, line 268:
```suggestion
	cmd.Flags().StringSliceP("output", "o", []string{}, fmt.Sprintf("Output format for conftest results - valid options are: %s. You can optionally specify a file for the output, e.g. -o json=out.json", append(output.Outputs(), OutputAppstudio)))
```

Not sure how that will look - maybe a line break is needed.

simonbaird:
Small nitpick about the command line help, otherwise lgtm.

docs/modules/ROOT/pages/ec_test.adoc
mbestavros: @simonbaird I don't think there's a way to specify a default value with `StringSliceP`, so we may just need to live with the empty default.

We could also manually call it out in the help text, e.g. "defaults to stdout" or something.

cmd/test/test.go
mbestavros: @simonbaird added!

simonbaird:
LGTM.

simonbaird (approved):

PR #1485, opened 2024-03-29
Change tag used for patch num counting

simonbaird:
Calling the v0.2.0 makes it look like a specific release tag, which is not really accurate. Let's use a name that better reflects its purpose.


simonbaird:
Ignoring spurious codecov failure as per usual.

PR #1486, opened 2024-04-01
Add support for multi-arch image index to ec validate command

yashvardhannanavati:
For an image index that is passed to the ec validate command, expand it into its image manifests and replace the original image index spec with the individual manifests in snapshot Spec

### DIFF
--- a/cmd/validate/image_test.go
+++ b/cmd/validate/image_test.go
@@ -31,6 +31,9 @@ import (
 	hd "github.com/MakeNowJust/heredoc"
 	"github.com/enterprise-contract/enterprise-contract-controller/api/v1alpha1"
 	"github.com/gkampitakis/go-snaps/snaps"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
 	app "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/sigstore/cosign/v2/pkg/cosign"
 	"github.com/spf13/afero"
@@ -42,8 +45,6 @@ import (
 	"github.com/enterprise-contract/ec-cli/cmd/root"
 	"github.com/enterprise-contract/ec-cli/internal/applicationsnapshot"
 	"github.com/enterprise-contract/ec-cli/internal/evaluator"
-	"github.com/enterprise-contract/ec-cli/internal/fetchers/oci"
-	"github.com/enterprise-contract/ec-cli/internal/fetchers/oci/fake"
 	"github.com/enterprise-contract/ec-cli/internal/output"
 	"github.com/enterprise-contract/ec-cli/internal/policy"
 	"github.com/enterprise-contract/ec-cli/internal/policy/source"
@@ -60,6 +61,35 @@ type data struct {
 var rootArgs = []string{
 	"validate",
 	"image",
+}
+
+type MockRemoteClient struct {
+	mock.Mock
+}
+
+func (m *MockRemoteClient) Get(ref name.Reference) (*remote.Descriptor, error) {
+	args := m.Called(ref)
+	result := args.Get(0)
+	if result != nil {
+		return result.(*remote.Descriptor), args.Error(1)
+	}
+	return nil, args.Error(1)
+}
+
+func (m *MockRemoteClient) Index(ref name.Reference) (v1.ImageIndex, error) {
+	args := m.Called(ref)
+	result := args.Get(0)
+	if result != nil {
+		return args.Get(0).(v1.ImageIndex), args.Error(1)
+	}
+	return nil, args.Error(1)
+}
+
+func commonMockClient(mockClient *MockRemoteClient) {
+	imageManifestJson := `{"mediaType": "application/vnd.oci.image.manifest.v1+json"}`
+	imageManifestJsonBytes := []byte(imageManifestJson)
+	// TODO: Replace mock.Anything calls with specific values
+	mockClient.On("Get", mock.Anything).Return(&remote.Descriptor{Manifest: imageManifestJsonBytes}, nil)
 }
 
 func Test_determineInputSpec(t *testing.T) {
@@ -238,10 +268,12 @@ func Test_determineInputSpec(t *testing.T) {
 			},
 		},
 	}
-
+	mockRemoteClient := &MockRemoteClient{}
+	commonMockClient(mockRemoteClient)
+	ctx := context.WithValue(context.Background(), applicationsnapshot.RemoteClientKey{}, mockRemoteClient)
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
-			s, err := applicationsnapshot.DetermineInputSpec(context.Background(), applicationsnapshot.Input{
+			s, err := applicationsnapshot.DetermineInputSpec(ctx, applicationsnapshot.Input{
 				File:   c.arguments.filePath,
 				JSON:   c.arguments.input,
 				Image:  c.arguments.imageRef,
@@ -294,7 +326,11 @@ func Test_ValidateImageCommand(t *testing.T) {
 	validateImageCmd := validateImageCmd(validate)
 	cmd := setUpCobra(validateImageCmd)
 
-	cmd.SetContext(utils.WithFS(context.TODO(), afero.NewMemMapFs()))
+	mockRemoteClient := &MockRemoteClient{}
+	commonMockClient(mockRemoteClient)
+	ctx := utils.WithFS(context.TODO(), afero.NewMemMapFs())
+	ctx = context.WithValue(ctx, applicationsnapshot.RemoteClientKey{}, mockRemoteClient)
+	cmd.SetContext(ctx)
 
 	effectiveTimeTest := time.Now().UTC().Format(time.RFC3339Nano)
 
@@ -370,7 +406,11 @@ func Test_ValidateImageCommandImages(t *testing.T) {
 	validateImageCmd := validateImageCmd(validate)
 	cmd := setUpCobra(validateImageCmd)
 
-	cmd.SetContext(utils.WithFS(context.Background(), afero.NewMemMapFs()))
+	mockRemoteClient := &MockRemoteClient{}
+	commonMockClient(mockRemoteClient)
+	ctx := utils.WithFS(context.TODO(), afero.NewMemMapFs())
+	ctx = context.WithValue(ctx, applicationsnapshot.RemoteClientKey{}, mockRemoteClient)
+	cmd.SetContext(ctx)
 
 	effectiveTimeTest := time.Now().UTC().Format(time.RFC3339Nano)
 
@@ -466,7 +506,11 @@ func Test_ValidateImageCommandKeyless(t *testing.T) {
 	})
 	cmd := setUpCobra(validateImageCmd)
 
-	cmd.SetContext(utils.WithFS(context.Background(), afero.NewMemMapFs()))
+	mockRemoteClient := &MockRemoteClient{}
+	commonMockClient(mockRemoteClient)
+	ctx := utils.WithFS(context.TODO(), afero.NewMemMapFs())
+	ctx = context.WithValue(ctx, applicationsnapshot.RemoteClientKey{}, mockRemoteClient)
+	cmd.SetContext(ctx)
 
 	cmd.SetArgs(append(rootArgs, []string{
 		"--image",
@@ -528,9 +572,12 @@ func Test_ValidateImageCommandYAMLPolicyFile(t *testing.T) {
 	validateImageCmd := validateImageCmd(validate)
 	cmd := setUpCobra(validateImageCmd)
 
+	mockRemoteClient := &MockRemoteClient{}
+	commonMockClient(mockRemoteClient)
 	fs := afero.NewMemMapFs()
-
-	cmd.SetContext(utils.WithFS(context.TODO(), fs))
+	ctx := utils.WithFS(context.TODO(), fs)
+	ctx = context.WithValue(ctx, applicationsnapshot.RemoteClientKey{}, mockRemoteClient)
+	cmd.SetContext(ctx)
 
 	cases := []struct {
 		name   string
@@ -642,9 +689,12 @@ func Test_ValidateImageCommandJSONPolicyFile(t *testing.T) {
 	validateImageCmd := validateImageCmd(validate)
 	cmd := setUpCobra(validateImageCmd)
 
+	mockRemoteClient := &MockRemoteClient{}
+	commonMockClient(mockRemoteClient)
 	fs := afero.NewMemMapFs()
-
-	cmd.SetContext(utils.WithFS(context.TODO(), fs))
+	ctx := utils.WithFS(context.TODO(), fs)
+	ctx = context.WithValue(ctx, applicationsnapshot.RemoteClientKey{}, mockRemoteClient)
+	cmd.SetContext(ctx)
 
 	testPolicyJSON := `sources:
   - policy:
@@ -718,9 +768,12 @@ func Test_ValidateImageCommandEmptyPolicyFile(t *testing.T) {
 	validateImageCmd := validateImageCmd(validate)
 	cmd := setUpCobra(validateImageCmd)
 
+	mockRemoteClient := &MockRemoteClient{}
+	commonMockClient(mockRemoteClient)
 	fs := afero.NewMemMapFs()
-
-	cmd.SetContext(utils.WithFS(context.TODO(), fs))
+	ctx := utils.WithFS(context.TODO(), fs)
+	ctx = context.WithValue(ctx, applicationsnapshot.RemoteClientKey{}, mockRemoteClient)
+	cmd.SetContext(ctx)
 
 	err := afero.WriteFile(fs, "/policy.yaml", []byte(nil), 0644)
 	if err != nil {
@@ -813,7 +866,11 @@ func Test_ValidateErrorCommand(t *testing.T) {
 			validateImageCmd := validateImageCmd(validate)
 			cmd := setUpCobra(validateImageCmd)
 
-			cmd.SetContext(utils.WithFS(context.TODO(), afero.NewMemMapFs()))
+			mockRemoteClient := &MockRemoteClient{}
+			commonMockClient(mockRemoteClient)
+			ctx := utils.WithFS(context.TODO(), afero.NewMemMapFs())
+			ctx = context.WithValue(ctx, applicationsnapshot.RemoteClientKey{}, mockRemoteClient)
+			cmd.SetContext(ctx)
 
 			cmd.SetArgs(append([]string{"validate", "image"}, c.args...))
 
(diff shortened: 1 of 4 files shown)


internal/applicationsnapshot/input.go
```diff
@@ -176,3 +210,79 @@ func readSnapshotSource(input []byte) (app.SnapshotSpec, error) {
 	log.Debugf("Read application snapshot from file %s", input)
 	return file, nil
 }
+
+func (*remoteClient) Get(ref name.Reference) (*remote.Descriptor, error) {
+	desc, err := remote.Get(ref)
```
lcarva, line 215: `return remote.Get(ref)` ?

Same for the `Index` and `clientFromContext` methods below.

internal/applicationsnapshot/input.go
```diff
@@ -176,3 +210,79 @@ func readSnapshotSource(input []byte) (app.SnapshotSpec, error) {
...
+
+func expandImageIndex(ctx context.Context, snap *app.SnapshotSpec) error {
+	client, ok := clientFromContext(ctx)
+	if !ok {
+		client = &remoteClient{}
+	}
+	// For an image index, remove the original component and replace it with an expanded component with all its image manifests
+	var removeIndices []int
+	var modifiedComponents []app.SnapshotComponent
+	log.Debugf("Contents of snapComponents are %v", snap.Components)
```
lcarva, line 243: I believe `%#v` should display a more readable representation of the components.

internal/applicationsnapshot/input.go
```diff
@@ -176,3 +210,79 @@ func readSnapshotSource(input []byte) (app.SnapshotSpec, error) {
...
+
+		if desc.MediaType.IsIndex() {
+			index, err := client.Index(ref)
+			if err != nil {
+				return fmt.Errorf("unable to fetch index for container image %s: %w", component.ContainerImage, err)
+			}
+
+			indexManifest, err := index.IndexManifest()
+			if err != nil {
+				return fmt.Errorf("unable to fetch image manifest for container image %s: %w", component.ContainerImage, err)
```
lcarva, line 264: "index manifest"?

internal/applicationsnapshot/input.go
```diff
@@ -176,3 +210,79 @@ func readSnapshotSource(input []byte) (app.SnapshotSpec, error) {
...
+			}
+
+			for _, manifest := range indexManifest.Manifests {
+				archComponent := component
+				archComponent.Name = fmt.Sprintf("%s-%s", component.Name, manifest.Platform.Architecture)
+				archComponent.ContainerImage = fmt.Sprintf("%s@%s", ref.Context().Name(), manifest.Digest)
```
lcarva, line 270: Should it be `ref.Repository.Name()` instead of `ref.Context().Name()`?

internal/applicationsnapshot/input.go
```diff
@@ -176,3 +210,79 @@ func readSnapshotSource(input []byte) (app.SnapshotSpec, error) {
...
+	return cli, ok
+}
+
+func expandImageIndex(ctx context.Context, snap *app.SnapshotSpec) error {
+	client, ok := clientFromContext(ctx)
+	if !ok {
+		client = &remoteClient{}
+	}
+	// For an image index, remove the original component and replace it with an expanded component with all its image manifests
+	var removeIndices []int
```
lcarva, line 241: NIT: Consider a slightly different approach by always building a new list of components:

```go
var components []app.SnapshotComponent
for _, c in range snap.Components {
  // ...
  if !isIndex {
    components = append(components, c)
    continue
  }
  for _, manifest := range expand index {
    // ...
    components = append(components, newComponent)
  }
}
snap.Components = components
```

internal/applicationsnapshot/input_test.go
```diff
@@ -163,13 +242,13 @@ func Test_DetermineInputSpec(t *testing.T) {
...
 				}
 			}
-
 			got, err := DetermineInputSpec(ctx, tc.input)
 			// expect an error so check for nil
 			if tc.want != nil {
 				assert.NoError(t, err)
 			}
 			assert.Equal(t, tc.want, got)
+			// assert.True(t, gock.IsDone(), "Expected http call not intercepted")
```
lcarva, line 251: Leftover?

internal/applicationsnapshot/input_test.go
```diff
@@ -34,6 +40,72 @@ import (
 	"github.com/enterprise-contract/ec-cli/internal/utils"
 )
 
+type MockRemoteClient struct {
```
lcarva, line 43: We use github.com/stretchr/testify/mock in some places to assist in mocking, example https://github.com/enterprise-contract/ec-cli/blob/ca3cf7bd6a864d57cc2edf8e33e1ffd296f7c4f3/internal/fetchers/oci/fake/client.go#L80

See if makes sense here.

internal/applicationsnapshot/input_test.go
```diff
@@ -223,3 +302,42 @@ func TestReadSnapshotFile(t *testing.T) {
 	})
 
 }
+
+func TestExpandImageIndex(t *testing.T) {
```
lcarva, line 306: It would be great to cover some of the error statements.

lcarva (commented):
Thanks for this! Looks pretty good overall, just a few minor comments.

internal/applicationsnapshot/input.go
yashvardhannanavati: Ah, good idea. Makes more sense

internal/applicationsnapshot/input.go
yashvardhannanavati: huh, I thought Repository was the type. I'll check

internal/applicationsnapshot/input_test.go
yashvardhannanavati: Done :)

internal/applicationsnapshot/input.go
yashvardhannanavati: I think it should be `ref.Context().Name()` based on https://pkg.go.dev/github.com/google/go-containerregistry/pkg/name#Reference

internal/applicationsnapshot/input.go
yashvardhannanavati: Nice! I changed `Get` and `Index`.  `clientFromContext` couldn't be changed since Go compiler doesn't allow that saying ctx.Value(key).(client) only returns a single value

zregvart:
@yashvardhannanavati you can run `make lint-fix` to sort the imports and pass the lint check

With this we seem to have lost the ability to gracefully handle cases when the image does not exist. From the [`Non strict with failures`](https://github.com/enterprise-contract/ec-cli/blob/5682f93d16166df81248343b21c5d843522acaf2/features/task_validate_image.feature#L172) scenario that fails in acceptance tests:

```
time="2024-04-03T09:03:30Z" level=debug msg="could not read images from file: open {\"components\": [{\"containerImage\": \"registry.image-registry.svc.cluster.local:44299/acceptance/does-not-exist\"}]}: no such file or directory"
time="2024-04-03T09:03:30Z" level=debug msg="Read application snapshot from file {\"components\": [{\"containerImage\": \"registry.image-registry.svc.cluster.local:44299/acceptance/does-not-exist\"}]}"
time="2024-04-03T09:03:30Z" level=debug msg="Calling expandImageIndex"
time="2024-04-03T09:03:30Z" level=debug msg="Read EnterpriseContractPolicy as k8s resource"
time="2024-04-03T09:03:30Z" level=debug msg="Using in-cluster configuration"
time="2024-04-03T09:03:30Z" level=debug msg="Initialized Kubernetes client"
time="2024-04-03T09:03:30Z" level=debug msg="Raw policy reference: \"acceptance-2n2lp/acceptance-policy-7cqm5\""
time="2024-04-03T09:03:30Z" level=debug msg="Parsed policy reference: acceptance-2n2lp/acceptance-policy-7cqm5"
time="2024-04-03T09:03:30Z" level=debug msg="GET https://10.96.0.1:443/apis/appstudio.redhat.com/v1alpha1/namespaces/acceptance-2n2lp/enterprisecontractpolicies/acceptance-policy-7cqm5 200 OK in 4 milliseconds"
time="2024-04-03T09:03:30Z" level=debug msg="Policy successfully fetched from cluster: v1alpha1.EnterpriseContractPolicy{TypeMeta:v1.TypeMeta{Kind:\"EnterpriseContractPolicy\", APIVersion:\"appstudio.redhat.com/v1alpha1\"}, ObjectMeta:v1.ObjectMeta{Name:\"acceptance-policy-7cqm5\", GenerateName:\"acceptance-policy-\", Namespace:\"acceptance-2n2lp\", SelfLink:\"\", UID:\"aa3f501a-513b-4368-9d92-9792c77308ea\", ResourceVersion:\"876\", Generation:1, CreationTimestamp:time.Date(2024, time.April, 3, 9, 3, 15, 0, time.Local), DeletionTimestamp:<nil>, DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ManagedFields:[]v1.ManagedFieldsEntry{v1.ManagedFieldsEntry{Manager:\"acceptance.test\", Operation:\"Update\", APIVersion:\"appstudio.redhat.com/v1alpha1\", Time:time.Date(2024, time.April, 3, 9, 3, 15, 0, time.Local), FieldsType:\"FieldsV1\", FieldsV1:(*v1.FieldsV1)(0xc0000118c0), Subresource:\"\"}}}, Spec:v1alpha1.EnterpriseContractPolicySpec{Name:\"\", Description:\"\", Sources:[]v1alpha1.Source(nil), Configuration:(*v1alpha1.EnterpriseContractPolicyConfiguration)(nil), RekorUrl:\"\", PublicKey:\"-----BEGIN PUBLIC KEY-----\\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEJ2jfDXZxC0QEl3ftOfhp/VF+wzYM\\nPkTtIY+KyVsP3I4wswF6fKpIRRFrhsPwt7gkdYplk1uNDZswX8v/aZ0Q5w==\\n-----END PUBLIC KEY-----\\n\", Identity:(*v1alpha1.Identity)(nil)}, Status:v1alpha1.EnterpriseContractPolicyStatus{}}"
time="2024-04-03T09:03:30Z" level=debug msg="Chosen to use effective time of `now`, using current time 2024-04-03T09:03:30Z"
time="2024-04-03T09:03:30Z" level=debug msg="Using long-lived key workflow"
Error: 1 error occurred:
	* unable to fetch descriptior for container image registry.image-registry.svc.cluster.local:44299/acceptance/does-not-exist: GET http://registry.image-registry.svc.cluster.local:44299/v2/acceptance/does-not-exist/manifests/latest: MANIFEST_UNKNOWN: manifest unknown; map[Tag:latest]. The client used was *applicationsnapshot.remoteClient
```

See the [snapshot failure](https://github.com/enterprise-contract/ec-cli/actions/runs/8513936201/job/23338747289?pr=1486#step:5:1379); we're expecting violation with code `builtin.image.accessible` to be reported

That scenario can be run with:

```shell
cd acceptance
go test -test.run 'TestFeatures/Non strict with failures' -args -persist
```

And the launched kind cluster can be used with:

```shell
KUBECONFIG=<(kind get kubeconfig --name $(kind get clusters)) kubectl ...
```

lcarva:
The failing acceptance test, `TestFeatures/Strict_with_failures`, is failing due to a change of behavior.

Before these changes, if you pass in an image reference that doesn't exist, EC reports that as a violation. With these changes, EC tries to access the image earlier. If the image doesn't exist, EC fails before it gets to the violation generation code. A missing image is now handled more like an "error" instead of just another violation.

IMO it's important to retain the old behavior. If EC is given 10 images to validate, and only one of them is inaccessible, then EC should still validate the other 9 images.

(We should probably have a unit test to verify this behavior. It is an accident that that particular acceptance test caught this.)

Conceptually, we probably want `expandImage` to return a list of violations that can be added to the report if it fails for whatever reason. But there may be other patterns that might be better suitable.

yashvardhannanavati:
IIUC, It seems like EC doesn't consider `inaccessible image` as an error for `ec validate`. So in order to conform to that pattern, `expandImageIndex` will not give a warning instead of an error if the image cannot be accessed. 

Thoughts? Please let me know if you prefer some other pattern.

internal/applicationsnapshot/input.go
```diff
@@ -97,6 +112,8 @@ func DetermineInputSpec(ctx context.Context, input Input) (*app.SnapshotSpec, er
 		if err != nil {
 			return nil, err
 		}
+		log.Debugf("Calling expandImageIndex")
```
zregvart, line 115: Logging here without any detail and only for one branch of the execution doesn't add value, perhaps adding logging statements to `expandImageIndex` makes more sense

internal/applicationsnapshot/input.go
```diff
@@ -97,6 +112,8 @@ func DetermineInputSpec(ctx context.Context, input Input) (*app.SnapshotSpec, er
...
+		expandImageIndex(ctx, &file)
```
zregvart, line 116: Could expandImageIndex be a function on the `snapshot` struct? Either way calling it once at the end of `DetermineInputSpec` should suffice, having it in each branch here makes it difficult to maintain and could mistakenly lead to doing more work than necessary by mistake (expanding the image index multiple times for the same image index)

internal/applicationsnapshot/input.go
```diff
@@ -176,3 +197,76 @@ func readSnapshotSource(input []byte) (app.SnapshotSpec, error) {
...
+}
+
+func (*remoteClient) Index(ref name.Reference) (v1.ImageIndex, error) {
+	return remote.Index(ref)
+}
+
+func clientFromContext(ctx context.Context) (client, bool) {
+	cli, ok := ctx.Value(RemoteClientKey{}).(client)
+	return cli, ok
+}
```
zregvart, line 212: Perhaps this; it would make the caller unaware of the client selection

```suggestion
var defaultClient = &remoteClient{}

func determineClient(ctx context.Context) client {
	if cli, ok := ctx.Value(RemoteClientKey{}).(client); ok {
		return cli
	}

	return defaultClient
}
```

zregvart:
> IIUC, It seems like EC doesn't consider `inaccessible image` as an error for `ec validate`.

Not sure what is meant with this, if an image of a component cannot be accessed a violation should be reported. [Here](https://github.com/enterprise-contract/ec-cli/blob/f555d51dfd0ef5dd2d56091455dfe0b9bd5e341d/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go#L111) is the bit that checks that.

internal/applicationsnapshot/input.go
```diff
@@ -176,3 +197,76 @@ func readSnapshotSource(input []byte) (app.SnapshotSpec, error) {
...
+		if err != nil {
+			allErrors = multierror.Append(allErrors, fmt.Errorf("unable to fetch index manifest for container image %s: %w", component.ContainerImage, err))
+			continue
+		}
+
+		// The image is an image index and accessible so remove the image index itself and add index manifests
+		components = components[:len(components)-1]
+		for _, manifest := range indexManifest.Manifests {
+			archComponent := component
+			archComponent.Name = fmt.Sprintf("%s-%s", component.Name, manifest.Platform.Architecture)
```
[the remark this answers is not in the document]
yashvardhannanavati: Would it work if I add `variant-architecture` as a suffix?

cmd/validate/image_test.go
```diff
@@ -1025,6 +1096,74 @@ func Test_WarningOutput(t *testing.T) {
...
+		"registry/image:tag",
+		"--policy",
+		fmt.Sprintf(`{"publicKey": %s}`, utils.TestPublicKeyJSON),
+		"--effective-time",
+		effectiveTimeTest,
+		"--strict",
+		"false",
+		"--ignore-rekor",
+		"true",
+	}...))
```
zregvart, line 1136: (nitpick) No need for the interim slice

```suggestion
	cmd.SetArgs(append(rootArgs,
		"--image",
		"registry/image:tag",
		"--policy",
		fmt.Sprintf(`{"publicKey": %s}`, utils.TestPublicKeyJSON),
		"--effective-time",
		effectiveTimeTest,
		"--strict",
		"false",
		"--ignore-rekor",
		"true"))
```

cmd/validate/image_test.go
```diff
@@ -1025,6 +1096,74 @@ func Test_WarningOutput(t *testing.T) {
...
+
+	var out bytes.Buffer
+	cmd.SetOut(&out)
+
+	utils.SetTestRekorPublicKey(t)
+
+	err := cmd.Execute()
+	assert.Error(t, err)
```
zregvart, line 1144: Should the content of the error be asserted?
```suggestion
	assert.Error(t, err)
	assert.EqualError(t, , err, "...")
```

cmd/validate/image_test.go
```diff
@@ -1025,6 +1096,74 @@ func Test_WarningOutput(t *testing.T) {
...
 
+func Test_FailureImageAccessibilityNonStrict(t *testing.T) {
+	validate := func(_ context.Context, component app.SnapshotComponent, _ policy.Policy, _ bool) (*output.Output, error) {
+		return &output.Output{
+			ImageSignatureCheck: output.VerificationStatus{
+				Passed: true,
+			},
+			ImageAccessibleCheck: output.VerificationStatus{
+				Passed: false,
+				Result: &evaluator.Result{Message: "Image URL is not accessible: HEAD http://${REGISTRY}/v2/acceptance/does-not-exist/manifests/latest: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details)"},
```
zregvart, line 1107: Would be better to match the value passed via `--image`
```suggestion
				Result: &evaluator.Result{Message: "Image URL is not accessible: HEAD registry/image:tag: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details)"},
```

internal/applicationsnapshot/input_test.go
```diff
@@ -223,3 +301,135 @@ func TestReadSnapshotFile(t *testing.T) {
...
+	mockRemoteClient := new(MockRemoteClient)
+	mockImageIndex := new(mockIndex)
+	mockRemoteClient.
+		On("Get", mock.Anything).Return(&remote.Descriptor{Descriptor: v1.Descriptor{MediaType: "application/vnd.oci.image.index.v1+json"}}, nil)
+	mockRemoteClient.
+		On("Index", mock.Anything).Return(mockImageIndex, nil)
```
zregvart, line 311: Would be better not to allow for anything (`mock.Anything`) but assert concrete values

zregvart (approved):
Some nitpicks, we can followup on this easily in a separate PR. I'll leave this open for others to have a look, for me :+1: 
Thanks!

yashvardhannanavati:
Please do not merge this PR until the config change for `enable-deep-inspect` is deployed to Production

cmd/validate/image_test.go
simonbaird: Done afaict (line 1144).

internal/applicationsnapshot/input_test.go
simonbaird: Is it worth adding a todo or filing a story for later re that suggestion?

simonbaird (commented):
Lgtm.

simonbaird (approved):

internal/applicationsnapshot/input_test.go
yashvardhannanavati: I can add a todo for that and file an issue. Thanks!

zregvart:
> Please do not merge this PR until the config change for `enable-deep-inspect` is deployed to Production

Not sure what is referenced here, is there a change that needs to happen in some other non-enterprise-contract repository that is a prerequisite of this? For Konflux I would expect that the e2e tests are or will updated to check this, so when we try to deliver this as an update via pull request such as [this](https://github.com/redhat-appstudio/infra-deployments/pull/3558) any issues will be caught.

From the Enterprise Contract side, this and the pull request in [ec-policies](https://github.com/enterprise-contract/ec-policies/pull/952) should be all that is needed.

internal/applicationsnapshot/input_test.go
yashvardhannanavati: Added todo in a couple of places and fixed the rest

PR #1494, opened 2024-04-04
When using --freshen the commit lookup is failing

joejstuart:
The commit lookup with --freshen uses the file path to find relevant commits. When @revision was appended to the path it would fail. This change will only use `freshen` if revision is left out and if both `revision` and `freshen` is specified, it will use `revision`.


zregvart (approved):
Thanks!

PR #1495, opened 2024-04-04
Add policy.yaml for ec verifications in Konflux

simonbaird:
We were previously using the standard redhat-non-hermetic config directly from https://github.com/enterprise-contract/config/blob/main/redhat-no-hermetic/policy.yaml

Now we have a need to use a customized policy so let's create one. I figured we could keep it here in this repo under .ec, in the style of .tekton, .github, etc.

As mentioned in the comments, ec by convention will look for a .ec/policy.yaml file at the git url specified.

Ref: [EC-520](https://issues.redhat.com/browse/EC-520)

### DIFF
--- a/hack/cut-release.sh
+++ b/hack/cut-release.sh
@@ -97,7 +97,7 @@ When it's done you can merge. (Continue to next section while you're waiting...)
 Go to the integration tests at ${KONFLUX_APPS_URL}/${KONFLUX_APPLICATION_NAME}/integrationtests
 Edit ${KONFLUX_APPLICATION_NAME}-enterprise-contract and add a parameter as follows:
   Name: POLICY_CONFIGURATION
-  Value: github.com/enterprise-contract/config//redhat-no-hermetic
+  Value: github.com/enterprise-contract/ec-cli
 Save changes
 
 # Apply cli pipeline modifications
--- a/.ec/policy.yaml
+++ b/.ec/policy.yaml
@@ -1,0 +1,45 @@
+# Copyright The Enterprise Contract Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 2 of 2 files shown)


hack/cut-release.sh
```diff
@@ -97,7 +97,7 @@ When it's done you can merge. (Continue to next section while you're waiting...)
 Go to the integration tests at ${KONFLUX_APPS_URL}/${KONFLUX_APPLICATION_NAME}/integrationtests
 Edit ${KONFLUX_APPLICATION_NAME}-enterprise-contract and add a parameter as follows:
   Name: POLICY_CONFIGURATION
-  Value: github.com/enterprise-contract/config//redhat-no-hermetic
+  Value: github.com/enterprise-contract/ec-cli
```
lcarva, line 100: Hmm if this is meant for a release branch, we should probably use the same exact policy as it is used by the release pipeline.

simonbaird: It effectively is the same now I think. But note that (IIUC) we can't access it from gitlab directly, so either way we would need to make a copy of it.

simonbaird:
I tested this early by tweaking the POLICY_CONFIGURATION parameter in the IntegrationTestScenario, and it looks like it's working as expected.

See https://console.redhat.com/preview/application-pipeline/workspaces/rhtap-contract/applications/ec-main-ci/pipelineruns/ec-main-ci-sxwqf-q8jmc for example.

zregvart:
I don't think this is the right approach. We have two issues here:
- Clair CVE scan is flagging unfixable CVEs as violations, I don't know what the policy is exactly but that seems wrong
- We have an issue unrelated to the PR (#1491) failing the check on the PR

I think we should find out what the policy is, and how are we or the Clair scan supposed to handle it. And I think the IntegrationTestScenario should be flagged as optional, with the agreement to file issues (as bugs?) for followup so we do not prevent merging a change unrelated to the found issue.

simonbaird:
This might not be needed in the short term. Moving to draft for now.

hack/cut-release.sh
lcarva: It depends on who's we. Red Hatter's? Sure we can. It is also synced to the cluster so Konflux users also have access. I'm actually quite confused about what this script is doing. It looks like it's about configuring Konflux for a new release branch. If so, I strongly recommend using the same exact release policy for integration tests. Otherwise, we are bound to find surprises at release time.

simonbaird: Script doesn't do anything by itself, it's really just a piece of interactive documentation.

simonbaird:
I'm still interested in the idea of maintaining an ec config in the repo here explicitly for use with building ec. E.g. imagine if we wanted to add some additional policies not included in the standard redhat/Konflux configs.

But right now we don't need it, so this is now abandoned.

PR #1497, opened 2024-04-05
Set container image name label explicitly

simonbaird:
Avoid an EC warning that I assume might become a violation some day: 'The "name" label should not be inherited from the parent image'

Before this the name inherited from the base image was 'ubi9-minimal'.

Will add this change to main branch later.

Motivated largely by wanting to trigger another Konflux build...


simonbaird:
I'll merge this without review if it goes green.

zregvart (commented):
👍

PR #1506, opened 2024-04-09
Use `GITHUB_REF` as default

zregvart:
When `GITHUB_BASE_REF` is not defined, i.e. the workflow is not running for a pull request, this uses `GITHUB_REF` instead.


zregvart:
Need to merge this so build-definitions picks up a newer release. We seem to be missing #1494 in the snapshot, so older version is used in the appstudio-utils image.

PR #1510, opened 2024-04-09
Use the existing cache for multi platform release with Podman update

joejstuart:
Use the existing cache for multi platform release. With this we can reuse the existing caches for releases for all os/arch  combinations. This should save a bit of time, but also help with disk space.


zregvart (approved):
Let’s give this a shot

PR #1529, opened 2024-04-15
Set name label in Dockerfile.dist

lcarva:
Images built in Konflux are required to have the `name` label set to a value different than the value of the same label from the parent image.

(I'm pretty sure this is to satisfy some requirement in the legacy build pipeline and is not actually used by anything in Konflux.)

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -73,6 +73,7 @@ ARG TARGETOS
 ARG TARGETARCH
 
 LABEL \
+  name="ec-cli" \
   description="Enterprise Contract verifies and checks supply chain artifacts to ensure they meet security and business policies." \
   io.k8s.description="Enterprise Contract verifies and checks supply chain artifacts to ensure they meet security and business policies." \
   summary="Provides the binaries for downloading the EC CLI. Also used as a Tekton task runner image for EC tasks." \


simonbaird:
FYI, I picked a different name in #1497 .

simonbaird (approved):

lcarva:
> FYI, I picked a different name in #1497 .

Cool. Updated to use that name instead.

Dockerfile.dist
```diff
@@ -73,12 +73,13 @@ ARG TARGETOS
...
 
 LABEL \
+  name="ec-cli" \
   description="Enterprise Contract verifies and checks supply chain artifacts to ensure they meet security and business policies." \
   io.k8s.description="Enterprise Contract verifies and checks supply chain artifacts to ensure they meet security and business policies." \
   summary="Provides the binaries for downloading the EC CLI. Also used as a Tekton task runner image for EC tasks." \
   io.k8s.display-name="Enterprise Contract for Red Hat Trusted Artifact Signer" \
   io.openshift.tags="rhtas rhtap trusted-artifact-signer trusted-application-pipeline enterprise-contract ec opa cosign sigstore" \
-  com.redhat.component="ec-cli"
+  com.redhat.component="ec-cli" \
```
simonbaird, line 82: One too many \ chares

lcarva: thanks!

PR #1530, opened 2024-04-16
Correctly compute the diff patch for git tracker

zregvart:
The `from` and `to` were substituted because the patch was computed commit to parent, instead of parent to commit.


lcarva (approved):
:tada:

PR #1532, opened 2024-04-16
Ensure evaluators are not destroyed if reused

zregvart:
We changed the life cycle of evaluators to the top of the command, and mistakenly left the destruction of the evaluator tied to the lifecycle of validating a single component.
This causes files created for one evalator and shared between component validations or evaluators inaccessible. An error such as:

```
get compiler options: capabilities not opened: open /tmp/ec-work-299517100/capabilities.json: no such file or directory
```

would be reported.

This changes the destruction of the evaluator to be in line with the new lifecycle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ clean: ## Delete build output
 .PHONY: test
 test: ## Run all unit tests
 	@echo "Unit tests:"
-	@go test -race -covermode=atomic -coverprofile=coverage-unit.out -timeout 500ms -tags=unit ./...
+	@go test -race -covermode=atomic -coverprofile=coverage-unit.out -timeout 1s -tags=unit ./...
 	@echo "Integration tests:"
 	@go test -race -covermode=atomic -coverprofile=coverage-integration.out -timeout 15s -tags=integration ./... | grep -v '\[no test files\]'
 # Given the nature of generative tests the test timeout is increased from 500ms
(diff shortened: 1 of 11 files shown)


robnester-rh (approved):
LGTM

simonbaird (approved):

zregvart:
The acceptance tests were tweaked, they did not fail when I removed the change from the production code because the evaluator was executing fast enough for the issue not to manifest itself. Now each component's image is unique, so evaluation takes a bit more, but enough to trigger the issue.

lcarva:
Seems like failure is due to a random Tekton resolver error:
```
Pipeline rhtap-contract-tenant/cli-main-ci-on-pull-request-xxpzv can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "": retryable error validating referenced object deprecated-image-check: Internal error occurred: failed calling webhook "webhook.pipeline.tekton.dev": failed to call webhook: Post "https://tekton-pipelines-webhook.openshift-pipelines.svc:443/defaulting?timeout=10s": context deadline exceeded
```

lcarva:
The `Red Hat Konflux / ec-main-ci-enterprise-contract / cli-main-ci` check is tracking the PipelineRun [ec-main-ci-j46zp-vjp6n](https://console.redhat.com/preview/application-pipeline/workspaces/rhtap-contract/applications/ec-main-ci/pipelineruns/ec-main-ci-j46zp-vjp6n/security). That has already completed successfully, but something prevented the results from getting reported here. Let's go ahead and merge this.

PR #1533, opened 2024-04-17
Bump version

simonbaird:


simonbaird:
Oops I clicked too fast. This was supposed to merge to simonbaird:main , not enterprise-contract:main. I think it's fairly harmless though. The impact is that the snapshot builds will start being 0.4.z instead of 0.3.z.

PR #1534, opened 2024-04-17
Allow duplicate values for slsa_provenance_v0.2 schema

yashvardhannanavati:
When processing multi-arch images, it is possible for the image index and the index manifests to have the same name. This change enables that. It is also in compliance with
https://slsa.dev/spec/v0.2/provenance#schema

### DIFF
--- a/pkg/schema/slsa_provenance_v0.2.json
+++ b/pkg/schema/slsa_provenance_v0.2.json
@@ -61,10 +61,7 @@
           "name",
           "digest"
         ]
-      },
-      "uniqueKeys": [
-        "/name"
-      ]
+      }
     },
     "predicateType": {
       "const": "https://slsa.dev/provenance/v0.2"
--- a/pkg/schema/slsa_provenance_v0.2_test.go
+++ b/pkg/schema/slsa_provenance_v0.2_test.go
@@ -91,7 +91,6 @@ func TestSubjectMustBeProvided(t *testing.T) {
 		`{"subject": [{"name": "a", "digest": {"foo": "abcdef0123456789"}}]}`,
 		`{"subject": [{"name": "a", "digest": {"sha256": ""}}]}`,
 		`{"subject": [{"name": "a", "digest": {"sha256": "g%-A"}}]}`,
-		`{"subject": [{"name": "x", "digest": {"sha256": "abcdef"}}, {"name": "x", "digest": {"sha256": "fedcba"}}]}`,
 	)
 }
 


lcarva (approved):
LGTM!

@zregvart, could you chime in here as you were pretty involved in adding this schema.

lcarva:
As an example, Chains created a SLSA Provenance with this for subject: 

```json
[
  {
    "name": "quay.io/redhat-user-workloads-stage/ynanavat-tenant/bootc-image-builder/bootc-image-builder",
    "digest": {
      "sha256": "df93e3976eb26199f3eaaa06cc4cd83f9d6663491aabaf73f67018cbf094dbc8"
    }
  },
  {
    "name": "quay.io/redhat-user-workloads-stage/ynanavat-tenant/bootc-image-builder/bootc-image-builder",
    "digest": {
      "sha256": "91a66493a17b922553468c24ce4d719fa22db3cabac3064c91f8ad9774ff0e91"
    }
  },
  {
    "name": "quay.io/redhat-user-workloads-stage/ynanavat-tenant/bootc-image-builder/bootc-image-builder",
    "digest": {
      "sha256": "929e3d5f989c811078e38d38ccd0708cdc6c18d06064d2a23bddb61b861e3ec2"
    }
  }
]
```

There's also nothing in the spec that mentions the `name` should be unique. In fact, it states this:

>   // Output file; name is "_" to indicate "not important".
>   "subject": [{"name": "_", "digest": {"sha256": "5678..."}}],

That implies the name could be repeated, e.g. `_`, if there are multiple subjects where the name is not important.

simonbaird (approved):

zregvart:
This tells me that whatever is generating the SLSA provenance is not doing it in conformance with [the specification](https://github.com/in-toto/attestation/blob/main/spec/v0.1.0/README.md#statement):

> subject[*].name string, required
> 
>     Identifier to distinguish this artifact from others within the subject.
> 
>     The semantics are up to the producer and consumer. Because consumers evaluate the name against a policy, the name SHOULD be stable between attestations. If the name is not meaningful, use "_". For example, a [SLSA Provenance](https://slsa.dev/provenance) attestation might use the name to specify output filename, expecting the consumer to only considers entries with a particular name. Alternatively, a vulnerability scan attestation might use the name "_" because the results apply regardless of what the artifact is named.
> 
>     MUST be non-empty and unique within subject.

lcarva:
> This tells me that whatever is generating the SLSA provenance is not doing it in conformance with [the specification](https://github.com/in-toto/attestation/blob/main/spec/v0.1.0/README.md#statement):
> 
> > subject[*].name string, required
> > ```
> > Identifier to distinguish this artifact from others within the subject.
> > 
> > The semantics are up to the producer and consumer. Because consumers evaluate the name against a policy, the name SHOULD be stable between attestations. If the name is not meaningful, use "_". For example, a [SLSA Provenance](https://slsa.dev/provenance) attestation might use the name to specify output filename, expecting the consumer to only considers entries with a particular name. Alternatively, a vulnerability scan attestation might use the name "_" because the results apply regardless of what the artifact is named.
> > 
> > MUST be non-empty and unique within subject.
> > ```

Ah!  I didn't check the in-toto spec which actually better defines that field. Well the culprit here is Chains. I also think that a unique name is somewhat shortsighted when it comes to images. Filed an issue on the spec: https://github.com/in-toto/attestation/issues/348

Regardless of the outcome there, I think we should merge this. We don't get any value from enforcing this restricting. Arguably, the uniqueness check also doesn't capture the special meaning of `_`.

lcarva:
Filed EC-558 for the long-term solution to this problem.

lcarva (approved):

PR #1535, opened 2024-04-17
Revert version snafu

simonbaird:
PR #1533 was merged in the wrong place at the wrong time. Revert it.


simonbaird:
```
git diff 3a9ba06a..eecf77f9
...nothing...
```

So I think it's safe to instamerge.

PR #1537, opened 2024-04-18
Adds full blown examples in the configuration docs

### LINKED ISSUE #1473: Include full configuration examples rather than snippets
In https://enterprisecontract.dev/docs/ec-cli/main/configuration.html#_examples the examples are snippets of YAML, yet the example above is an inline JSON. That can be a bit confusing. We should have whole examples and in single format to make the configuration clearer.

zregvart:
Adds full YAML and JSON examples in the configuration docs.

Requires tabs from the website repository[1]

Resolves #1473

[1] https://github.com/enterprise-contract/enterprise-contract.github.io/pull/251


simonbaird (approved):
Awesome.

PR #1538, opened 2024-04-18
Fix manual triggered release workflow logic

simonbaird:
The intention here iiuc is to always do a release if the release workflow was triggered manually. This wasn't working, but should be working now.

Noticed this while testing some wip for EC-522 in my fork.

Ref: [EC-522](https://issues.redhat.com/browse/EC-522)


simonbaird:
https://console.redhat.com/preview/application-pipeline/workspaces/rhtap-contract/applications/ec-main-ci/pipelineruns/ec-main-ci-z2qn9-4wplm/detail is green, so not sure why it's still yellow and spinning above.

The "checks" failure is something about uploading codecov results, so should be safe to ignore.

simonbaird:
This is pretty safe to merge without review. See notes above about the failure.

PR #1539, opened 2024-04-19
Unit tests no network

zregvart:
Modifies the unit tests not to reach out over the network.


cmd/validate/image_test.go
```diff
@@ -1194,6 +1194,7 @@ func TestEvaluatorLifecycle(t *testing.T) {
 		fmt.Sprintf(`{"publicKey": %s, "sources": [%s]}`, utils.TestPublicKeyJSON, strings.Join(sources, ", ")),
 		"--effective-time",
 		effectiveTimeTest,
+		"--ignore-rekor",
```
lcarva, line 1197: FYI: This is fine, of course. An alternative is to call `utils.SetTestRekorPublicKey(t)`.

PR #1542, opened 2024-04-19
Remove rebasing logic from uncommitted check

zregvart:
I think we might not need this, if a file was changed on main and the PR touches it it will show up in the diff or the PR will be marked conflicting. If we need this merge queues is probably a better approach.


lcarva (approved):
:heart:

PR #1547, opened 2024-04-22
Do not fetch the inlined rule data multiple times

lcarva:


robnester-rh (approved):
LGTM

internal/policy/source/source_test.go
```diff
@@ -103,19 +104,19 @@ func TestInlineDataSource(t *testing.T) {
...
+	require.FileExists(t, file)
 
 	data, err := os.ReadFile(file)
-	assert.NoError(t, err)
-	assert.Equal(t, []byte("some data"), data)
+	require.NoError(t, err)
+	require.Equal(t, []byte("some data"), data)
 
-	assert.Equal(t, "data:application/json;base64,c29tZSBkYXRh", s.PolicyUrl())
+	require.Equal(t, "data:application/json;base64,c29tZSBkYXRh", s.PolicyUrl())
```
simonbaird, line 119: What's the difference here?

simonbaird: I googled it: https://www.yellowduck.be/posts/assert-vs-require-in-testify

robnester-rh (commented):
LGTM

internal/policy/source/source.go
```diff
@@ -144,13 +144,21 @@ func InlineData(source []byte) PolicySource {
 }
 
 func (s inlineData) GetPolicy(ctx context.Context, workDir string, showMsg bool) (string, error) {
-	dest := uniqueDestination(workDir, s.Subdir(), fmt.Sprintf("%x", (sha256.Sum256(s.source))))
+	dest := path.Join(workDir, s.Subdir())
```
zregvart, line 147: With the use of `downloadCache` the previous `dest` using `uniqueDestination` can be used

lcarva: Just tried. It doesn't work because it makes the key unique on every invocation of `GetPolicy`.

PR #1578, opened 2024-05-01
Publish versioned release after snapshot release

simonbaird:
There's a short window of time between removing the snapshot build and creating a new one, where the snapshot release doesn't exist. IIUC this change should make that window shorter because we're publishing the snapshot release before the versioned release instead of after.

It looks like the time taken for a release publish is 6 seconds, so this potentially reduces the time we have no snapshot release available from 12 seconds down to 6 seconds, give or take.

There might be a way to eliminate the 6 second gap entirely and also avoid the scenario where a broken release builds means there is no snapshot available at all, but I'll file a separate story for that and we can worry about it later.

Followup patch for:

Ref: https://issues.redhat.com/browse/EC-602

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -164,14 +164,6 @@ jobs:
 
           git push -f --tags
 
-      - name: Versioned release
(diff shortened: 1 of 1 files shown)


simonbaird:
Trivial change so I think I'll merge without review.

PR #1579, opened 2024-05-02
Make it so snapshot release is always "latest"

simonbaird:
In commit 33af1620 PR #1578 there was an unintentional side-effect of changing which release was designated as "latest" in GitHub. Let's undo the unintentional change.

Followup to a followup of...

Ref: https://issues.redhat.com/browse/EC-602


robnester-rh (approved):
LGTM

PR #1587, opened 2024-05-06
Renovate configuration for updating .tool-versions

### LINKED ISSUE #1089: Use renovatebot to keep .tools-versions up to date
We added the [.tools-versions](.tools-versions) file to this repo in order to better support usage of [asdf](https://asdf-vm.com/). The go version listed is in the format MAJOR.MINOR.PATCH.  Although updates to MAJOR and MINOR can be done manually, we should ensure the latest value for PATCH is used since it may contain important bug/security fixes.

Let's use renovatebot to automate this process, since it already [supports asdf](https://docs.renovatebot.com/modules/manager/asdf/), and the ability to [restrict updates to only patch updates](https://docs.renovatebot.com/configuration-options/#matchupdatetypes).

Let's also be careful to ensure renovatebot doesn't interfere with dependabot already setup in this repo. This is likely possible by simply omitting the `extends` directive.

zregvart:
Fixes #1089

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,9 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "timezone": "America/New_York",
+  "schedule": ["before 10pm"],
+  "automerge": true,
+  "automergeType": "pr",
+  "platformAutomerge": true,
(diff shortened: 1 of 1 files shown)


robnester-rh (approved):
LGTM

PR #1596, opened 2024-05-07
Upgrade to UBI 9.4

zregvart:
This also removes the need for using openshift-golang-builder from brew as UBI 9.4 ships golang 1.21.

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -26,11 +26,13 @@
 # See https://source.redhat.com/groups/public/teamnado/wiki/brew_registry for how to get your own pull secret
 # See our Konflux pull secret at https://console.redhat.com/preview/application-pipeline/secrets
 
-FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_1.21@sha256:4fe910174caaaae09ff75b6f1b1c2f4460fd9acfe38ec778a818a54de7f31afc AS build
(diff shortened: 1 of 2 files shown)


Dockerfile.dist
```diff
@@ -26,12 +26,14 @@
 # See https://source.redhat.com/groups/public/teamnado/wiki/brew_registry for how to get your own pull secret
 # See our Konflux pull secret at https://console.redhat.com/preview/application-pipeline/secrets
 
-FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_1.21@sha256:4fe910174caaaae09ff75b6f1b1c2f4460fd9acfe38ec778a818a54de7f31afc AS build
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4@sha256:2636170dc55a0931d013014a72ae26c0c2521d4b61a28354b3e2e5369fa335a3 AS build
```
simonbaird, line 29: IIUC there's some fancy stuff in `openshift-golang-builder` or maybe also in go-toolset (not sure) that maybe we prefer to use over plain old ubi9 with golang. Not saying it's a blocker, but maybe something to consider.

zregvart: The compliance shim? I think that [this](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/golang_builder_FIPS_wrapper.sh) is the script that is producing those messages. To me it looks like it is mostly focused on enabling FIPS, not sure if we need that today or not.

simonbaird (approved):

PR #1603, opened 2024-05-07
Add +redhat to version for Konflux builds

simonbaird:
The motivation is so there's a way to distinguish a Red Hat Konflux build of ec from an GitHub upstream build.

After this patch, `ec version` will show something like `v0.4.1+redhat` for a Konflux build, and `v0.4.1` for the same build if it was done in GitHub.

Ref: [EC-573](https://issues.redhat.com/browse/EC-573)

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -28,6 +28,7 @@
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4@sha256:2636170dc55a0931d013014a72ae26c0c2521d4b61a28354b3e2e5369fa335a3 AS build
 
+ARG BUILD_SUFFIX="redhat"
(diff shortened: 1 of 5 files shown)


robnester-rh (approved):
LGTM

Dockerfile.dist
```diff
@@ -41,7 +41,7 @@ RUN go mod download
 # Now copy everything including .git
 COPY . .
 
-RUN /build/build.sh "${BUILD_LIST}"
+RUN /build/build.sh "${BUILD_LIST}" redhat
```
zregvart, line 44: Would it make sense to make this `"redhat"` an ARG?

zregvart (approved):

Dockerfile.dist
simonbaird: Maybe not needed right now, but I like it. Added to new revision.

simonbaird:
New revision includes Zoran's suggestion, and I added some semi-related improvements for local build testing with `make dist-container`.

PR #1604, opened 2024-05-07
Expose --extra-rule-data in v-e-c task

mbestavros:
Adds a new param called `EXTRA_RULE_DATA` in the `verify-enterprise-contract` task that maps to the new `--extra-rule-data` flag. Supports multiple key/value pairs with `key1=value1,key2=value2` syntax.

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -110,6 +110,10 @@ spec:
       type: string
       description: Run policy checks with the provided time.
       default: "now"
+    - name: EXTRA_RULE_DATA
+      type: string
+      description: Merge additional Rego variables into the policy data. Use syntax "key=value,key2=value2..."
+      default: ""
 
   workspaces:
     - name: data
@@ -170,6 +174,7 @@ spec:
         - "--strict=false"
         - "--show-successes"
         - "--effective-time=$(params.EFFECTIVE_TIME)"
+        - "--extra-rule-data=$(params.EXTRA_RULE_DATA)"
         - "--output"
         - "yaml=$(params.HOMEDIR)/report.yaml"
         - "--output"
--- a/docs/modules/ROOT/pages/verify-enterprise-contract.adoc
+++ b/docs/modules/ROOT/pages/verify-enterprise-contract.adoc
@@ -57,6 +57,7 @@ paths can be provided by using the `:` separator.
 *EFFECTIVE_TIME* (`string`):: Run policy checks with the provided time.
 +
 *Default*: `now`
+*EXTRA_RULE_DATA* (`string`):: Merge additional Rego variables into the policy data. Use syntax "key=value,key2=value2..."
 
 == Results
 
--- a/features/__snapshots__/task_validate_image.snap
+++ b/features/__snapshots__/task_validate_image.snap
@@ -274,6 +274,100 @@ success: true
 
 ---
 
+[Extra rule data provided to task:report - 1]
+components:
+- attestations:
+  - predicateBuildType: tekton.dev/v1beta1/TaskRun
+    predicateType: https://slsa.dev/provenance/v0.2
+    signatures:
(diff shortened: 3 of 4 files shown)


tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -110,6 +110,10 @@ spec:
       type: string
       description: Run policy checks with the provided time.
       default: "now"
+    - name: EXTRA_RULE_DATA
+      type: string
+      description: Merge additional Rego variables into the policy data. Use syntax "key=value,key2=value2..."
+      default: "key=value"
```
simonbaird, line 116: Probably needs to be blank by default, which would solve the snapshot failures in the acceptance tests. But I think maybe there'll be some extra work to avoid giving ec `--extra-rule-data=` with no values where there are no values.

mbestavros: Per offline discussion, changed this to empty string. Should still work, and will fix acceptance tests.

mbestavros:
@simonbaird I added a new acceptance test and everything looks green - ready to go?

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -170,6 +174,7 @@ spec:
         - "--strict=false"
         - "--show-successes"
         - "--effective-time=$(params.EFFECTIVE_TIME)"
+        - "--extra-rule-data=$(params.EXTRA_RULE_DATA)"
```
simonbaird, line 177:
```suggestion
        - "--extra-rule-data='$(params.EXTRA_RULE_DATA)'"
```

simonbaird: I think this will help support a key or value with a space, e.g. `--extra-rule-data='foo=hi there,bar=123'` might work, but `--extra-rule-data=foo=hi there,bar=123` will throw errors.

simonbaird (commented):
Need some quotes, but otherwise lgtm.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
simonbaird: Okay, I guess I don't know how to quote these things..

Acceptance test has this:

```
- Snapshot - 2
+ Received + 2

@@ -85,7 +85,7 @@

      ruleData:
-       key1: value1
-       key2: value2
+       '''key1': value1
+       key2: value2'
```

mbestavros: @simonbaird Is that acceptable behavior? Feels like it could cause issues if that quoting makes the key/value different than intended. Or it could just be an anomaly in the report only?

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -170,6 +174,7 @@ spec:
         - "--strict=false"
         - "--show-successes"
         - "--effective-time=$(params.EFFECTIVE_TIME)"
+        - "--extra-rule-data='$(params.EXTRA_RULE_DATA)'"
```
zregvart, line 177: The `args` will be taken as strings and provided as is in the `argv` array to the executable in `command`, this is not subject to shell words, so no need for quotes here. The quotes that remain here are for YAML to force `args` to be an array of strings, but most likely we do not need them as well as they could default to being strings. Good to be safe though...

```suggestion
        - "--extra-rule-data=$(params.EXTRA_RULE_DATA)"
```

mbestavros: OK, I reverted the quotes, and it looks like acceptance tests are green again :+1: fyi @simonbaird

simonbaird: Ooh now I get it, thanks..!

PR #1607, opened 2024-05-08
Use go-toolset golang base image for redhat builds

simonbaird:
The --chown=1001:0 was required to avoid an error like this:

  fatal: detected dubious ownership in repository at '/build'
  To add an exception for this directory, call:
          git config --global --add safe.directory /build

I'm not exactly sure the reason why, but I believe it's because the new base image comes with a user called 'default' with uid 1001. When we copy the source code into the build container it comes in owned by root, and that causes git to be suspicious when it runs as user default.

Ref: https://issues.redhat.com/browse/EC-625

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -19,26 +19,22 @@
 # This works fine but will produce an EC violation
 # FROM docker.io/library/golang:1.21 AS build
 
-# This currently has go version 1.20 but we need version 1.21
-#FROM registry.access.redhat.com/ubi9/go-toolset:latest AS build
-
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4@sha256:2636170dc55a0931d013014a72ae26c0c2521d4b61a28354b3e2e5369fa335a3 AS build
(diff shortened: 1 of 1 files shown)


simonbaird:
Draft because I don't know if this is something we want or not. Happy to abandon unless we can decide there's some value in using `ubi9/go-toolset` instead of plain `ubi9/ubi-minimal`.

zregvart:
I've seen RPM repositories fail with 500/404, perhaps just for Fedora mirrors, not sure. So the benefit of this could be that we depend on only registry.access.redhat.com rather than that and the RPM repository as well...

simonbaird:
[The relevant Dockerfile iiuc](https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?architecture=amd64&image=6633d168b5bb692a240e57b3&container-tabs=dockerfile).

simonbaird:
> Hmm yeah, it also means we don't need to figure out the cachi2 prefetch for those rpms.

Not true, since we also microdnf install jq later on.

simonbaird:
Merged #1608 and abandoned this one.

PR #1608, opened 2024-05-09
Use go-toolset golang base image for redhat builds

simonbaird:
Use root user instead of the 'default' user from go-toolset to avoid an error like this:

  fatal: detected dubious ownership in repository at '/build'
  To add an exception for this directory, call:
          git config --global --add safe.directory /build

Could also have solved the problem by doing a --chown=default:root when copying in the files.

Ref: https://issues.redhat.com/browse/EC-625

### DIFF
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -19,17 +19,15 @@
 # This works fine but will produce an EC violation
 # FROM docker.io/library/golang:1.21 AS build
 
-# This currently has go version 1.20 but we need version 1.21
-#FROM registry.access.redhat.com/ubi9/go-toolset:latest AS build
-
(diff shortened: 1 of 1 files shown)


robnester-rh:
Does this leave the working user as `root` and does that present any security concerns?

simonbaird:
> Does this leave the working user as `root` and does that present any security concerns?

No, because it's just the build step. In the final image we have `USER 1001`.

robnester-rh (approved):
LGTM

robnester-rh:
> > Does this leave the working user as `root` and does that present any security concerns?
> 
> No, because it's just the build step. In the final image we have `USER 1001`.

Ack.

zregvart (approved):

PR #1627, opened 2024-05-16
Fix logic for no matching rule

lcarva:
If a source group produces no warnings, failures, or successes, we want to raise an error as that is a no-op and likely indicates a typo in the policy configuration.

This commit fix the check by ensuring the total number of matching rules is computed *after* the exclude/include from the policy config is applied.

Ref: EC-583

### DIFF
--- a/acceptance/examples/rule_data.rego
+++ b/acceptance/examples/rule_data.rego
@@ -1,15 +1,32 @@
 package main
 
-deny[result] {
-	not data.rule_data__configuration__
+import rego.v1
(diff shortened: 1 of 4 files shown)


robnester-rh (approved):

simonbaird (approved):
Nice. Approve assuming you'll figure out the test failures.

lcarva:
Had to update one of the acceptance tests after all since it was actually relying on the bug to succeed.

PR #1630, opened 2024-05-16
Switch to using OCI Trusted Artifacts Pipeline

lcarva:
Ref: EC-556

### DIFF
--- a/.tekton/cli-main-ci-pull-request.yaml
+++ b/.tekton/cli-main-ci-pull-request.yaml
@@ -51,28 +51,6 @@ spec:
         - name: kind
           value: task
         resolver: bundles
-    - name: show-summary
-      params:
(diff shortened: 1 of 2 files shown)


.tekton/cli-main-ci-pull-request.yaml
```diff
@@ -51,28 +51,6 @@ spec:
         - name: kind
           value: task
         resolver: bundles
-    - name: show-summary
```
lcarva, line 54: This Task is not really useful and we may just drop it altogether. See more context in the comments of EC-643.

lcarva:
EC failure is due to the `*-oci-ta` Task variants not being listed as viable alternatives to the required Tasks. Working on that.

simonbaird (approved):

PR #1631, opened 2024-05-16
Human readable output

simonbaird:
Cleaned up and ready for serious review now.

I think this could be merged as is, but it does need some extra work on the following:
* Instead of top level grouping by component, it should be top level grouping by violation/warning/success.
* There should be a way to disable/enable colors, ideally with auto detection of terminal color support (as @zregvart suggested in the review here).

Those could be in followup PRs, or could be included in this one.

Ref: [EC-563](https://issues.redhat.com/browse/EC-563).

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -69,6 +69,8 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		spec                        *app.SnapshotSpec
 		strict                      bool
 		images                      string
+		noColor                     bool
+		forceColor                  bool
 	}{
 		strict: true,
 	}
@@ -402,6 +404,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 				return err
 			}
 			p := format.NewTargetParser(applicationsnapshot.JSON, cmd.OutOrStdout(), utils.FS(cmd.Context()))
+			utils.SetColorEnabled(data.noColor, data.forceColor)
 			if err := report.WriteAll(data.output, p); err != nil {
 				return err
 			}
@@ -486,6 +489,12 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		Include additional information on the failures. For instance for policy
 		violations, include the title and the description of the failed policy
 		rule.`))
+
+	cmd.Flags().BoolVar(&data.noColor, "no-color", data.info, hd.Doc(`
+		Disable color when using text output even when the current terminal supports it`))
+
+	cmd.Flags().BoolVar(&data.forceColor, "color", data.info, hd.Doc(`
+		Enable color when using text output even when the current terminal does not support it`))
 
 	if len(data.input) > 0 || len(data.filePath) > 0 || len(data.images) > 0 {
 		if err := cmd.MarkFlagRequired("image"); err != nil {
--- a/internal/applicationsnapshot/report.go
+++ b/internal/applicationsnapshot/report.go
@@ -18,6 +18,8 @@ package applicationsnapshot
 
 import (
 	"bytes"
+	"embed"
+	_ "embed"
 	"encoding/json"
 	"encoding/xml"
 	"fmt"
@@ -33,6 +35,7 @@ import (
 	"github.com/enterprise-contract/ec-cli/internal/format"
 	"github.com/enterprise-contract/ec-cli/internal/policy"
 	"github.com/enterprise-contract/ec-cli/internal/signature"
+	"github.com/enterprise-contract/ec-cli/internal/utils"
 	"github.com/enterprise-contract/ec-cli/internal/version"
 )
 
@@ -96,6 +99,7 @@ type TestReport struct {
 const (
 	JSON            = "json"
 	YAML            = "yaml"
+	Text            = "text"
 	AppStudio       = "appstudio"
 	Summary         = "summary"
 	SummaryMarkdown = "summary-markdown"
@@ -111,6 +115,7 @@ const (
 var OutputFormats = []string{
 	JSON,
 	YAML,
+	Text,
 	AppStudio,
 	Summary,
 	SummaryMarkdown,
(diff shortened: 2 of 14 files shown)


internal/applicationsnapshot/report.go
```diff
@@ -94,20 +97,35 @@ type TestReport struct {
 
 // Possible formats the report can be written as.
 const (
-	SummaryMarkdown = "summary-markdown"
 	JSON            = "json"
 	YAML            = "yaml"
-	APPSTUDIO       = "appstudio"
+	Text            = "text"
```
zregvart, line 102: As you're refactoring, we could define a `type OutputFormatType string` and have these typed like that...

internal/applicationsnapshot/report.go
```diff
@@ -289,6 +309,67 @@ func generateMarkdownSummary(r *Report) ([]byte, error) {
 	return markdownBuffer.Bytes(), nil
 }
 
+func generateTextReport(r *Report) ([]byte, error) {
+
+	mainTemplate := hd.Doc(`
```
zregvart, line 314: Perhaps use [embed](https://pkg.go.dev/embed) and have the template in a separate file?

internal/applicationsnapshot/report.go
```diff
@@ -289,6 +309,67 @@ func generateMarkdownSummary(r *Report) ([]byte, error) {
...
+		{{ end }}
+		{{ if and (gt .SuccessCount 0) (.Successes) -}}
+		Successes: {{ template "results" .Successes }}
+		{{ end }}
+		{{ end }}{{/* range .Components */}}
+	`)
+
+	partials := map[string]string{
+		// For convenience since juggling the {{- and -}} gets tricky
+		"NL": "\n",
```
zregvart, line 336: Yes, though much less so once you get into the rhythm of using only/mostly `{{-`, see the [templates for the docs](https://github.com/enterprise-contract/ec-cli/tree/main/internal/documentation/asciidoc) for inspiration

internal/utils/helpers.go
```diff
@@ -155,3 +155,19 @@ func HasSuffix(str string, extensions []string) bool {
 func HasJsonOrYamlExt(str string) bool {
 	return HasSuffix(str, []string{".json", ".yaml", ".yml"})
 }
+
+func ansiColorText(str string, color int) string {
```
zregvart, line 159: We use https://pkg.go.dev/github.com/doiit/picocolors (that no longer seems to exist) in the acceptance tests. We might want to think about disabling colors based on the environment, picocolors does something like this:

```
func isColorSupported() bool {
        noColor := os.Getenv("NO_COLOR") != "" || contains(os.Args, "--no-color")
        forceColor := os.Getenv("FORCE_COLOR") != "" || os.Getenv("CLICOLOR_FORCE") != "" || contains(os.Args, "--color")
        terminal := isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) || os.Getenv("TERM") == "dumb"
        ci := os.Getenv("CI") != ""

        return !noColor && (forceColor || terminal || ci)
}
```

There are [terminfo](https://pkg.go.dev/search?q=terminfo&m=) packages we could also use to se if the terminal has color support (think CI)

internal/applicationsnapshot/report.go
```diff
@@ -289,6 +309,67 @@ func generateMarkdownSummary(r *Report) ([]byte, error) {
...
+		"green":  utils.GreenText,
+		"yellow": utils.YellowText,
+	}
+
+	t.Funcs(helpers)
+
+	t = template.Must(t.Parse(mainTemplate))
+
+	for k, v := range partials {
+		t = template.Must(t.Parse(fmt.Sprintf("{{- define \"%s\" -}}\n%s\n{{- end -}}", k, hd.Doc(v))))
```
zregvart, line 365: Could put the nested templates in that same file that gets embedded, or use https://pkg.go.dev/text/template#ParseFS with embed to keep them in separate files

cmd/validate/image.go
```diff
@@ -72,6 +72,9 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 	}{
 		strict: true,
 	}
+
+	validOutputFormats := applicationsnapshot.OutputFormats
```
zregvart, line 76: :+1:

internal/applicationsnapshot/report.go
simonbaird: Yeah I was thinking about that. Maybe in another commit since I merged #1640 already.

internal/applicationsnapshot/report.go
simonbaird: Yep, that's done in the latest revision.

internal/applicationsnapshot/report.go
simonbaird: Actually I didn't end up using it :sweat_smile:

internal/utils/helpers.go
simonbaird: Yeah good idea.

cmd/validate/image.go
```diff
@@ -487,6 +490,12 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 		violations, include the title and the description of the failed policy
 		rule.`))
 
+	cmd.Flags().BoolVar(&data.noColor, "no-color", data.info, hd.Doc(`
+		Disable color when using text output even when the current terminal supports it`))
+
+	cmd.Flags().BoolVar(&data.forceColor, "force-color", data.info, hd.Doc(`
```
zregvart, line 496: Perhaps just `--color`?
```suggestion
	cmd.Flags().BoolVar(&data.forceColor, "color", data.info, hd.Doc(`
```

zregvart (approved):

internal/utils/helpers.go
simonbaird: Third commit has this now.

PR #1643, opened 2024-05-22
Big format update from lsp/gopls via LazyVim

simonbaird:
It all seems pretty reasonable. Let's consider.


zregvart (approved):
I like your editor friend

lcarva:
Does `go fmt` agree?

simonbaird:
Not sure how to enforce this, maybe we can figure it out and make a CI format checker.

simonbaird:
Revision is just a new commit message.

simonbaird:
Oh I cancelled the Konflux build to save some CPU cycles, but I forgot it's a hard requirement now.

PR #1644, opened 2024-05-23
Use new template utils for inspect policy also

simonbaird:
The new templating helpers were introduced in #1631. I want to use the same technique in some older templating.

This required a little refactor of the helpers because we want to be able to use io.Writer directly rather than write to a buffer.

Also add some extra unit tests.


robnester-rh (approved):
LGTM

zregvart (approved):

PR #1656, opened 2024-05-30
Rework validate image human readable text output

simonbaird:
- Show a summary at the beginning
- Show the component details before the results
- Group by the result type instead of by component
- Increase text wrap width
- Show the image ref with each result
- General tweaks and improvements

Ref: https://issues.redhat.com/browse/EC-657

PR #1663, opened 2024-06-04
Don't wrap image ref in text output

simonbaird:
The wrapping only wraps on whitespace, so before this patch we'd end up with output like this for long image refs, which IMO looks untidy. Let's just have an over-long line rather than wrapping it like that.

```
› [Warning] cve.unpatched_cve_warnings
  ImageRef:
  quay.io/redhat-user-workloads/somelong-app-name/some-long-component-name@etc...
```

With the agressive space eating done by -}} and {{- it's easier to keep using the indentWrap helper func rather than try to add the two space indent some other way.

This is a followup tweak for the functionality introduced in #1656.

### DIFF
--- a/internal/applicationsnapshot/templates/_results.tmpl
+++ b/internal/applicationsnapshot/templates/_results.tmpl
@@ -16,7 +16,7 @@
     {{- colorIndicator $type }} {{ colorText $type (printf "[%s] %s" $type .Metadata.code) }}{{ nl -}}
 
     {{- if $imageRef -}}
-      {{- indentWrap $indent $wrap (printf "ImageRef: %s" $imageRef ) }}{{ nl -}}
+      {{- indent $indent (printf "ImageRef: %s" $imageRef ) }}{{ nl -}}
     {{- end -}}
 
     {{/* For a success the message is generally just "Pass" so don't show it */}}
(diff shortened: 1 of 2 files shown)


internal/applicationsnapshot/templates/_results.tmpl
```diff
@@ -16,7 +16,8 @@
     {{- colorIndicator $type }} {{ colorText $type (printf "[%s] %s" $type .Metadata.code) }}{{ nl -}}
 
     {{- if $imageRef -}}
-      {{- indentWrap $indent $wrap (printf "ImageRef: %s" $imageRef ) }}{{ nl -}}
+      {{/* Keep the indent but don't actually wrap the text for this one */}}
+      {{- indentWrap $indent 1000 (printf "ImageRef: %s" $imageRef ) }}{{ nl -}}
```
zregvart, line 20: Would this work?
```suggestion
      ImageRef: {{  $imageRef }}{{ nl -}}
```

zregvart (approved):

internal/applicationsnapshot/templates/_results.tmpl
simonbaird: Actually it's not quite so simple since the space before gets eaten up. I tried a few variations, decide to add an indent helper that just indents. :shrug:

simonbaird: Doesn't matter much, but I was thinking about the $indent var and avoiding hard coded indenting...

simonbaird:
I'll merge this when/if it goes green, don't want to invest more review time since it's so trivial.

PR #1664, opened 2024-06-04
Use .template extension for all go templates

simonbaird:
There were some templates used in the text output with extension .tmpl. Rename those to be consistent.


zregvart (approved):

simonbaird:
Let's go for #1665 instead.

PR #1673, opened 2024-06-07
Update to non-deprecated go-jose dependency

simonbaird:
The square/go-jose module was deprecated and archived in Feb 2023. Use the new version from go-jose/go-jose.


simonbaird:
This is trivial and affects the test suite only, so I feel okay to merge without waiting for a review.

PR #1695, opened 2024-06-12
Teach EC CLI to initialize a TUF root

lcarva:
Ref: EC-584


zregvart (approved):

simonbaird (approved):
Nice!

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -147,7 +147,7 @@ spec:
         fi
 
         echo 'Initializing TUF root...'
-        cosign initialize --mirror "${TUF_MIRROR}" --root "${TUF_MIRROR}/root.json"
+        ec sigstore initialize --mirror "${TUF_MIRROR}" --root "${TUF_MIRROR}/root.json"
```
simonbaird, line 150: Just brainstorming, but we could consider calling it `ec sigstore cosign-initialize`  - it would leave the command namespace more open in case we want to add other sigstore stuff and make it more clear what this is equivalent to.

robnester-rh (approved):

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
lcarva: So the initialize also works for `gitsign`. I think `sigstore` is the right level.

PR #1697, opened 2024-06-13
Fix Konflux build platform arch (release-v0.4 branch)

simonbaird:
The GOOS and GOARCH env vars need to be set explicitly, otherwise we just build the linux x86 binary over and over. Looks like this regression was introduced in commit b8db99c2.

I think restoring the export on those two vars would also have worked.

Ref: https://issues.redhat.com/browse/SECURESIGN-1054

(cherry picked to release-v0.4 branch from main branch commit d62ea87fc5bbf9ff8367faf6429ee9a1045fc5f2)

This is the release branch version of #1696.


simonbaird:
The pre-merge build only creates one binary, so we'll need to wait for the post-merge build to double check it's doing the right thing.

But I did test it locally and it seems to be working as expected.

PR #1706, opened 2024-06-21
Add support for JSON and YAML files in --extra-rule-data

yashvardhannanavati:
This commit adds support for specifying filename as value of a key in --extra-rule-data which can be later used in policy checks.

resolves: CVP-4147

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -227,12 +227,12 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 					policySpec := p.Spec()
 					sources := policySpec.Sources
 					for i := range sources {
-						source := sources[i]
+						src := sources[i]
 						var rule_data_raw []byte
 						unmarshaled := make(map[string]interface{})
 
-						if source.RuleData != nil {
-							rule_data_raw, err = source.RuleData.MarshalJSON()
+						if src.RuleData != nil {
+							rule_data_raw, err = src.RuleData.MarshalJSON()
 							if err != nil {
 								log.Errorf("Unable to parse ruledata to raw data")
 							}
@@ -249,7 +249,11 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 							if len(parts) < 2 {
 								log.Errorf("Incorrect syntax for --extra-rule-data")
 							}
-							unmarshaled[parts[0]] = parts[1]
+							extraRuleDataPolicyConfig, err := validate_utils.GetPolicyConfig(ctx, parts[1])
+							if err != nil {
+								log.Errorf("Unable to load data from extraRuleData: %s", err.Error())
+							}
+							unmarshaled[parts[0]] = extraRuleDataPolicyConfig
 						}
 						rule_data_raw, err = json.Marshal(unmarshaled)
 						if err != nil {
--- a/cmd/validate/image_test.go
+++ b/cmd/validate/image_test.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -31,6 +32,8 @@ import (
 	"github.com/gkampitakis/go-snaps/snaps"
 	app "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/sigstore/cosign/v2/pkg/cosign"
+	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 
@@ -747,6 +750,10 @@ func Test_ValidateImageCommandExtraData(t *testing.T) {
       - "registry/policy:latest"
     data:
       - "registry/policy-data:latest"
+    ruleData:
+      custom_rule_data:
+        prefix_data:
+          - registry1
 configuration:
   collections:
     - minimal
@@ -759,6 +766,22 @@ configuration:
 		panic(err)
 	}
 
+	testExtraRuleDataYAML := `---
+kind: ReleasePlanAdmission
+spec:
+  application: [some-app]
+  data:
+    mapping:
+      components:
+        - name: some-name
+          repository: quay.io/some-namespace/msd
+`
+
+	err = afero.WriteFile(fs, "/value.yaml", []byte(testExtraRuleDataYAML), 0644)
+	if err != nil {
+		panic(err)
+	}
+
 	cmd.SetArgs(append(rootArgs, []string{
 		"--image",
 		"registry/image:tag",
@@ -767,7 +790,7 @@ configuration:
 		"--policy",
 		"/policy.json",
 		"--extra-rule-data",
-		"key=value",
+		"key=/value.yaml,key2=value2",
 	}...))
 
 	var out bytes.Buffer
@@ -794,7 +817,9 @@ configuration:
 			"registry/policy:latest"
 		],
 		"ruleData": {
-			"key":"value"
+			"custom_rule_data":{"prefix_data":["registry1"]},
+			"key": "---\nkind: ReleasePlanAdmission\nspec:\n  application: [some-app]\n  data:\n    mapping:\n      components:\n        - name: some-name\n          repository: quay.io/some-namespace/msd\n",
+			"key2": "value2"
 		}
 	  }`, string(sourceSampleMarshaled))
 }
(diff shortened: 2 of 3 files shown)


cmd/validate/image_test.go
```diff
@@ -759,6 +759,22 @@ configuration:
...
 
+	testExtraRuleDataYAML := `---
+kind: ReleasePlanAdmission
+spec:
+  application: [some-app]
+  data:
+    mapping:
+	  components:
+	    - name: some-name
+		  repository: quay.io/some-namespace/msd
```
simonbaird, line 770: Formatting nitpick: Probably doens't matter it looks like inconsistent indenting, maybe check the tabs/spaces inside the string.

Ps, in some places we use github.com/MakeNowJust/heredoc.Doc (sometimes as hd.Doc)  to make these kinds of strings look tidier. There's no firm policy about it, but you could use it here if you want.

cmd/validate/image_test.go
```diff
@@ -794,7 +810,8 @@ configuration:
 			"registry/policy:latest"
 		],
 		"ruleData": {
-			"key":"value"
+			"key": "---\nkind: ReleasePlanAdmission\nspec:\n  application: [some-app]\n  data:\n    mapping:\n\t  components:\n\t    - name: some-name\n\t\t  repository: quay.io/some-namespace/msd\n",
```
simonbaird, line 813: Did you consider unmarshalling it? My thinking is it's more flexible not to, but since you're only doing this for a .json or .yaml maybe it makes sense.

Related question, I wonder if a user would ever want to slurp in some text from a .txt file..?

internal/validate/helpers.go
```diff
@@ -50,32 +50,33 @@ func GetPolicyConfig(ctx context.Context, policyConfiguration string) (string, e
 		if err != nil {
 			return "", err
 		}
-		return readPolicyConfigurationFile(ctx, configFile)
+		log.Debugf("Loading %s as policy configuration", policyConfiguration)
```
simonbaird, line 53: Should it be:

```suggestion
		log.Debugf("Loading %s as policy configuration", configFile)
```

internal/validate/helpers.go
```diff
@@ -50,32 +50,33 @@ func GetPolicyConfig(ctx context.Context, policyConfiguration string) (string, e
...
 	return policyConfiguration, nil
 }
 
-// Read policyConfiguration file and return its contents.
-func readPolicyConfigurationFile(ctx context.Context, policyConfiguration string) (string, error) {
-	// Check if policyConfig is a file path. If so, try to read it.
-	// If successful we write that into the data.policyConfiguration var.
+// Read file from the workspace and return its contents.
+func ReadWorkspaceFile(ctx context.Context, fileName string) (string, error) {
+	// Check if fileName is a file path. If so, try to read it.
```
simonbaird, line 69: I know it's not exactly part of your patch here, but is this comment wrong now? Seems like we're just assuming it's a file path.

simonbaird (commented):
I'm just wondering about the marshalling idea, not suggesting we should do it.

Lgtm apart from the one debug var name fix.

cmd/validate/image_test.go
simonbaird: Actually for the sake of the test here, the string doesn't matter, you could set it to "bacon" if you want. :relaxed:

simonbaird: ..though it would matter if we wanted to unmarshal

internal/validate/helpers.go
yashvardhannanavati: makes sense. will get rid of it

cmd/validate/image_test.go
yashvardhannanavati: nice catch! I fixed the indentation on the string :)

cmd/validate/image_test.go
yashvardhannanavati: hm, please correct me if I am missing something but it does look like it gets unmarshalled later on here: https://github.com/enterprise-contract/ec-cli/pull/1706/files#diff-7f11215ba891ff8199fa09a412849a5414e416050d163c813f6499009bbe0b96R271 (Line 271)

I did not consider .txt file in the scope of this change as we don't have a use case for it. Do you think we should add it now?

internal/validate/helpers.go
yashvardhannanavati: good eye! Fixed it

yashvardhannanavati:
> ## [Codecov](https://app.codecov.io/gh/enterprise-contract/ec-cli/pull/1706?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=enterprise-contract) Report
> Attention: Patch coverage is `77.77778%` with `4 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 86.74%. Comparing base [(`ab24b6a`)](https://app.codecov.io/gh/enterprise-contract/ec-cli/commit/ab24b6a76f7f0d7bb41a2b1d57e69c932a87866f?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=enterprise-contract) to head [(`bf3a66a`)](https://app.codecov.io/gh/enterprise-contract/ec-cli/commit/bf3a66a8332cef4d485c06963a4df33e0761f561?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=enterprise-contract).
> 
> > ❗ **Current head [bf3a66a](https://github.com/enterprise-contract/ec-cli/commit/bf3a66a8332cef4d485c06963a4df33e0761f561) differs from pull request most recent head [dadfa45](https://github.com/enterprise-contract/ec-cli/commit/dadfa454e854a6fcd71d02a560a69f5b130d7cf9)**
> > Please [upload](https://docs.codecov.com/docs/codecov-uploader) reports for the commit [dadfa45](https://github.com/enterprise-contract/ec-cli/commit/dadfa454e854a6fcd71d02a560a69f5b130d7cf9) to get more accurate results.
> 
> Additional details and impacted files

Added another test to cover the missing Error log. 

For future reference, there are multiple error log lines that are not covered which are not a part of this PR.

cmd/validate/image.go
```diff
@@ -37,7 +37,7 @@ import (
 	"github.com/enterprise-contract/ec-cli/internal/format"
 	"github.com/enterprise-contract/ec-cli/internal/output"
 	"github.com/enterprise-contract/ec-cli/internal/policy"
-	"github.com/enterprise-contract/ec-cli/internal/policy/source"
+	source_utils "github.com/enterprise-contract/ec-cli/internal/policy/source"
```
zregvart, line 40: I would not make this change, if the shadowing of `source` is an issue at all, I'd rename the variable from `source` to `src` or just `s`. e.g. [here](https://github.com/yashvardhannanavati/ec-cli/blob/dadfa454e854a6fcd71d02a560a69f5b130d7cf9/cmd/validate/image.go#L230)

cmd/validate/image.go
```diff
@@ -249,7 +249,15 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 							if len(parts) < 2 {
 								log.Errorf("Incorrect syntax for --extra-rule-data")
 							}
-							unmarshaled[parts[0]] = parts[1]
+							if source_utils.SourceIsFile(parts[1]) && utils.HasJsonOrYamlExt(parts[1]) {
```
zregvart, line 252: It might make sense use `--extra-rule-data=file:...` syntax, so we can pull in from other kinds of services, not just from file system, in that case relying on `github.com/enterprise-contract/ec-cli/internal/policy/source.PolicySource` might be a good solution here

zregvart, line 252: To me it looks like we would overwrite any `sources[i].RuleData` that was already set on the policy here with `--extra-rule-data` flag (that might be a bug prior to this change). I don't see any prior test coverage for when both `ruleData` in policy and `--extra-rule-data` is provided.

internal/validate/helpers.go
```diff
@@ -50,32 +50,32 @@ func GetPolicyConfig(ctx context.Context, policyConfiguration string) (string, e
...
 	// we assume it's a string and return it as is.
 	return policyConfiguration, nil
 }
 
-// Read policyConfiguration file and return its contents.
-func readPolicyConfigurationFile(ctx context.Context, policyConfiguration string) (string, error) {
-	// Check if policyConfig is a file path. If so, try to read it.
-	// If successful we write that into the data.policyConfiguration var.
+// Read file from the workspace and return its contents.
+func ReadWorkspaceFile(ctx context.Context, fileName string) (string, error) {
```
zregvart, line 68: I would not mention the Workspace, here, that is a Tekton term, not a concern of the CLI, not sure about converting to `string` it, might be a good opportunity to reevaluate that

```suggestion
func ReadFile(ctx context.Context, fileName string) ([]byte, error) {
```

cmd/validate/image.go
yashvardhannanavati: makes sense. Done

cmd/validate/image.go
yashvardhannanavati: I can confirm it wasn't a bug. I added `ruleData` to my unit test along with `--extra-rule-data` and it seemed to work fine. Please let me know what you think

internal/validate/helpers.go
yashvardhannanavati: Removed `Workspace` from the function name. 

I think the reason behind returning `string` instead of `[]byte` would have been returning the string data as is in-case the input is a `string` and not a file in `GetPolicyConfig`. Do you still think we should change that?

zregvart (approved):
LGTM

PR #1719, opened 2024-06-27
Use forked version of go-containerregistry

zregvart:
This version supports credential lookup according to[1]

Reference: EC-664

[1] https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md#format

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,6 +49,10 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+// use forked version until we can get a fix merged
+// see https://github.com/google/go-containerregistry/pull/1966
+replace github.com/google/go-containerregistry => github.com/zregvart/go-containerregistry v0.0.0-20240627132555-cec55a14ea32
+
 require (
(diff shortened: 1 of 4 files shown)


simonbaird (approved):

go.mod
```diff
@@ -49,6 +49,10 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+// use forked version until we can get a fix merged
+// see https://github.com/google/go-containerregistry/pull/1966
+replace github.com/google/go-containerregistry => github.com/zregvart/go-containerregistry v0.0.0-20240627132555-cec55a14ea32
```
simonbaird, line 54: `github.com/zregvart/go-containerregistry` feels like it implies main branch. Is there a way to indicate it's another branch? I know there's a git sha in the version, so the branch doesn't particularly matter, but I'm just curious about it.

zregvart: Golang doesn't version branches, this is a [pseudo version](https://go.dev/doc/modules/version-numbers#pseudo-version-number) for untagged commit

PR #1722, opened 2024-06-28
Enable go-gather functionality via env var

robnester-rh:
This commit allows for using the go-gather library in lieu of the conftest-download functionality. In order to utilize go-gather, set the USEGOGATHER environmental variable to 1, i.e.

```
USEGOGATHER=1 ec validate input [...]
```


internal/downloader/downloader.go
```diff
@@ -53,6 +57,23 @@ func Download(ctx context.Context, destDir string, sourceUrl string, showMsg boo
...
+	// - if USEGOGATHER=0, use the Conftest downloader
+	// - if USEGOGATHER=1, use the go-gather downloader
+	// - if USEGOGATHER is not set, use the Conftest downloader
+	// - if USEGOGATHER is set to any other value, use the Conftest downloader
+	// This is a temporary solution to allow for testing the go-gather downloader
+	// without breaking the existing functionality.
+	// Once the go-gather downloader is stable, the USEGOGATHER env variable
+	// will be removed and the go-gather downloader will be used by default.
+	// The USEGOGATHER env variable will be removed in a future release.
+	if experimental := os.Getenv("USEGOGATHER"); experimental != "" {
```
zregvart, line 70: Perhaps it is not necessary to read the environment variable twice? Perhaps something like:

```
	dl := func(ctx context.Context, dest, src string) error {
		// conftest's Download function leverages oras under the hood to fetch from OCI. It uses the
		// global oras client and sets the user agent to "conftest". This is not a thread safe
		// operation. Here we get around this limitation by ensuring a single download happens at a
		// time.
		dlMutex.Lock()
		defer dlMutex.Unlock()
		return downloader.Download(ctx, dest, []string{src})
	}
	if os.Getenv("USEGOGATHER") == "1" {
		dl = func(ctx context.Context, dest, src string) error {
			_, err := gather.Gather(ctx, src, dest)
			return err
		}
	} else if d, ok := ctx.Value(downloadImplKey).(downloadImpl); ok {
		dl = func(ctx context.Context, dest, src string) error {
			return d.Download(ctx, dest, []string{src})
		}
	}

	err = dl(ctx, destDir, []string{sourceUrl})

File: internal/downloader/downloader_test.go [hunk 2]
```diff
@@ -61,20 +65,52 @@ func TestDownloader_Download(t *testing.T) {
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			d := mockDownloader{}
 			ctx := WithDownloadImpl(context.TODO(), &d)
-			d.On("Download", ctx, tt.dest, []string{tt.source}).Return(tt.err)
+
+			originalGatherFunction := gatherFunc
+
+			if tt.useGoGather == "1" {
+				os.Setenv("USEGOGATHER", "1")
```
zregvart, line 91: use [`testing.Setenv`](https://pkg.go.dev/testing#T.Setenv) instead

internal/downloader/downloader.go
```diff
@@ -62,13 +74,19 @@ func Download(ctx context.Context, destDir string, sourceUrl string, showMsg boo
...
-		// global oras client and sets the user agent to "conftest". This is not a thread safe
-		// operation. Here we get around this limitation by ensuring a single download happens at a
-		// time.
-		dlMutex.Lock()
-		err = downloader.Download(ctx, destDir, []string{sourceUrl})
-		dlMutex.Unlock()
+
+		// similar as above with respect to USEGOGATHER
+		if os.Getenv("USEGOGATHER") == "1" {
+			_, err = gatherFunc(ctx, sourceUrl, destDir)
```
simonbaird, line 80: The commentary below says conftest download is not thread safe, and we need the lock/unlock. Are we thinking that gatherFunc *is* thread safe here?

go.mod
```diff
@@ -126,7 +128,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/chrismellard/docker-credential-acr-env v0.0.0-20230304212654-82a0ddb27589 // indirect
 	github.com/clbanning/mxj/v2 v2.7.0 // indirect
-	github.com/cloudflare/circl v1.3.7 // indirect
+	github.com/cloudflare/circl v1.3.8 // indirect
```
simonbaird, line 131: Do we really need the (I guess) unrelated dependency bumps like this to be included in this PR?

internal/downloader/downloader.go
```diff
@@ -53,6 +57,14 @@ func Download(ctx context.Context, destDir string, sourceUrl string, showMsg boo
 		return fmt.Errorf("attempting to download from insecure source: %s", sourceUrl)
 	}
 
+	// read the USEGOGATHER env variable to determine which downloader to use:
+	// - if USEGOGATHER=1, use the go-gather downloader
+	// - if USEGOGATHER is not set, or set to any other value, use the Conftest downloader
+	if os.Getenv("USEGOGATHER") == "1" {
```
simonbaird, line 63: You could make it a little tidier by moving this to `internal/utils/helpers.go`, similar to `utils.Experimental()` defined in there.

simonbaird: Actually I'm confused about why lines 60-63 are needed. Wouldn't it be better to keep going to line 80 before checking the env var? That way the `WithDownloadImpl` used in testing to substitute in a mock downloader could still work.

Or, if not, then how would we ever get to line 80?

simonbaird: Also, we'd want the logging on line 68-72 to be done either way I think.

internal/downloader/downloader.go
simonbaird: Not sure, but we might want to use the mocked downloader for unit tests even when USEGOGATHER is set. So I think the `if d, ok := ctx.Value(downloadImplKey).(downloadImpl); ok` part should be the first check instead of the last.

In pseudo code:
```
  if downloader impl found in context:
    // must be in unit test, downloads are mocked probably
  else if go gather env flag is set:
    // use go gather
  else:
    // good old conftest download
  fi
```

simonbaird (changes requested):
Some tweaks needed I think.

internal/downloader/downloader.go
robnester-rh: This was tidied up in a subsequent push.

internal/downloader/downloader.go
robnester-rh: gatherFunc _should_ be thread safe here.

go.mod
robnester-rh: These are brought in with executing `go mod tidy`.

internal/downloader/downloader.go
robnester-rh: Being that the intention is to move to `go-gather` in lieu of the `conftest-downloader`, I'm not sure this is really worth much in the current time frame. We can update this when / if needed when we transition to using `go-gather` exclusively. At that time we'll do away with looking at the `USEGOGATHER` env var anyway.

zregvart (approved):
LGTM

go.mod
simonbaird: But why? Are they go-gather sub-dependencies?

Ideally this PR has just the minimum diff to main branch.

internal/downloader/downloader_test.go
```diff
@@ -61,20 +65,52 @@ func TestDownloader_Download(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
 			d := mockDownloader{}
 			ctx := WithDownloadImpl(context.TODO(), &d)
-			d.On("Download", ctx, tt.dest, []string{tt.source}).Return(tt.err)
+
+			originalGatherFunction := gatherFunc
+
+			if tt.useGoGather == "1" {
+				t.Setenv("USEGOGATHER", "1")
+				gatherFunc = func(_ context.Context, _ string, _ string) (metadata.Metadata, error) {
```
simonbaird, line 92: I might be missing something, but I'm confused about how it downloads from `example.com/repo.git` if it's not using `mockDownloader{}`

internal/utils/helpers.go
```diff
@@ -124,6 +124,10 @@ func Experimental() bool {
 	return os.Getenv("EC_EXPERIMENTAL") == "1"
 }
 
+func UseGoGather() bool {
+	return os.Getenv("USEGOGATHER") == "1"
+}
```
simonbaird, line 129: I like it.

internal/downloader/downloader_test.go
```diff
@@ -61,20 +65,52 @@ func TestDownloader_Download(t *testing.T) {
 			source: "http://example.com",
 			err:    errors.New("attempting to download from insecure source: http://example.com"),
 		},
+		{
+			name:        "experimental download",
+			dest:        "dir",
+			source:      "example.com/repo.git",
+			useGoGather: "1",
```
simonbaird, line 72: nitpick: I'd probably make that one a boolean

internal/downloader/downloader_test.go
simonbaird: Oh, is gatherFunc here being set to a mocked function, but different to mock downloader?

robnester-rh: yes

internal/downloader/downloader_test.go
```diff
@@ -67,14 +71,33 @@ func TestDownloader_Download(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			d := mockDownloader{}
```
simonbaird, line 72: otherD = ...

internal/downloader/downloader.go
```diff
@@ -59,17 +64,27 @@ func Download(ctx context.Context, destDir string, sourceUrl string, showMsg boo
...
 		// operation. Here we get around this limitation by ensuring a single download happens at a
 		// time.
 		dlMutex.Lock()
-		err = downloader.Download(ctx, destDir, []string{sourceUrl})
-		dlMutex.Unlock()
+		defer dlMutex.Unlock()
+		return downloader.Download(ctx, destDir, []string{sourceUrl})
+	}
+
+	if d, ok := ctx.Value(downloadImplKey).(downloadImpl); ok {
```
simonbaird, line 77: If I remembered right, we have to move this back how it was, because we want to not use this code for the go-gather branch of the unit test, (unless you have a plan to untangle the logic).

robnester-rh: this has been moved back.

simonbaird (approved):

PR #1742, opened 2024-07-03
Use source code reference in simple demo script

simonbaird:
There is a slsa check that produces violations if the source reference is not provided. Let's update the simple demo script to include the git sha and url so this check can pass.

This gets things passing again with the default golden container image and the default ECP.

### DIFF
--- a/hack/simple-demo.sh
+++ b/hack/simple-demo.sh
@@ -19,11 +19,28 @@
 # See https://github.com/enterprise-contract/golden-container
 IMAGE=${IMAGE:-"quay.io/redhat-appstudio/ec-golden-image:latest"}
 
+# Assume the latest image was pushed already
+GIT_REPO=enterprise-contract/golden-container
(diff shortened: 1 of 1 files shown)


hack/simple-demo.sh
```diff
@@ -19,11 +19,28 @@
 # See https://github.com/enterprise-contract/golden-container
 IMAGE=${IMAGE:-"quay.io/redhat-appstudio/ec-golden-image:latest"}
 
+# Assume the latest image was pushed already
+GIT_REPO=enterprise-contract/golden-container
+GIT_SHA=$(curl -s "https://api.github.com/repos/${GIT_REPO}/commits?per_page=1" | jq -r '.[0].sha')
```
zregvart, line 24: Could read the `vcs-ref` label of the image in case the latest commit hasn't been pushed:

```suggestion
GIT_SHA=$(skopeo inspect docker://quay.io/redhat-appstudio/ec-golden-image:latest --format '{{index .Labels "vcs-ref"}}')
```

zregvart (approved):

PR #1752, opened 2024-07-08
Re-enable stats and make collection script more robust

mbestavros:
Ref: https://issues.redhat.com/browse/EC-576 (tl;dr summary: release stats on the website got pulled because it was breaking releases. We should re-enable them and fix the root issue.)

I added some shell hackery to "continue-on-fail" for the `gh` call, which was causing the Actions failures in the linked issue. Now, the rest of the setup should work as expected, so I'm re-enabling it.

### DIFF
--- a/hack/stats.sh
+++ b/hack/stats.sh
@@ -50,5 +50,5 @@ mkdir -p stats
         "downloads": .downloadCount
         }
     ]
-    }'
+    }' || true
 } > stats/stats.json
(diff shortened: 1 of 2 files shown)


zregvart (approved):
LGTM

hack/stats.sh
```diff
@@ -50,5 +50,7 @@ mkdir -p stats
         "downloads": .downloadCount
         }
     ]
-    }'
+    }' || {
+        :
+    }
```
zregvart, line 55: Actually this could be a bit simpler
```suggestion
    }' || true
```

mbestavros: Good call, added!

zregvart (approved):

PR #1762, opened 2024-07-16
Configure hermetic build

zregvart:
The `hack/update-rpm-lock.sh` script will run the `rpm-lockfile-prototype`[1] tool configured with the `Dockerfile.dist` and `rpms.in.yaml` files to generate the `rpms.lock.yaml` file. The lock file can be read by cachi2[2] to precache RPM packages with pinned versions. And that in turn enables hermetic builds.

The order of build stages in `Dockerfile.dist` was reversed as the current version of `rpm-lockfile-prototype` reads the first `FROM` in a Dockerfile to determine the base image.

`hack/update-rpm-lock.sh` script helps by installing the needed tools and dependencies in a container image that is later run to execute the `rpm-lockfile-prototype` tools. The script also extracts any `.repo` files in `/etc/yum.repos.d` which can be referenced in the `rpms.in.yaml` file.

Pulls the common parts of the PipelineRuns into a Pipeline definition that is resolved from the PipelineRuns. This way changes are confined to a single file for push and pull request PipelineRuns.

Adds RPM precache support and enables the hermetic build mode.

Reference: EC-360

[1] https://github.com/konflux-ci/rpm-lockfile-prototype
[2] https://github.com/containerbuildsystem/cachi2/


zregvart:
The switch of the `FROM` order in the `Dockefile.dist` broke the ec-cli image.

zregvart:
Should be fixed now.

zregvart:
Hint: when reviewing look at the individual commits, there is a `yq . .tekton/*.yaml` that reformats the files in [b5f6a71](https://github.com/enterprise-contract/ec-cli/pull/1762/commits/b5f6a71cf1253a19223bb5392873ff8be46d68e4), this was needed as the next commit bumps the Task bundle versions needed to get a successful PipelineRun.

zregvart:
EC with the `@redhat` policy and without any exclusions now succeeds:

```
$ ec validate image --images <(echo '{
  "application": "test",
  "components": [
    {
      "name": "test",
      "containerImage": "quay.io/redhat-user-workloads/rhtap-contract-tenant/ec-main-ci/cli-main-ci:on-pr-0716a612ff8bf7653d29005a4d5dd39737846d9f",
      "source": {
        "git": {
          "url": "https://github.com/zregvart/ec-cli.git",
          "revision": "0716a612ff8bf7653d29005a4d5dd39737846d9f"
        }
      }
    }
  ]
}') --ignore-rekor --policy '{
  "apiVersion": "appstudio.redhat.com/v1alpha1",
  "kind": "EnterpriseContractPolicy",
  "metadata": {
    "name": "default",
    "namespace": "enterprise-contract-service"
  },
  "spec": {
    "description": "Includes rules for levels 1, 2 & 3 of SLSA v0.1. This is the default config used for new RHTAP applications. Available collections are defined in https://redhat-appstudio.github.io/docs.stonesoup.io/ec-policies/release_policy.html#_available_rule_collections. If a different policy configuration is desired, this resource can serve as a starting point. See the docs on how to include and exclude rules https://redhat-appstudio.github.io/docs.stonesoup.io/ec-policies/policy_configuration.html#_including_and_excluding_rules.",
    "name": "Default",
    "publicKey": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEZP/0htjhVt2y0ohjgtIIgICOtQtA\nnaYJRuLprwIv6FDhZ5yFjYUEtsmoNcW7rx2KM6FOXGsCX3BNc7qhHELT+g==\n-----END PUBLIC KEY-----\n",
    "sources": [
      {
        "config": {
          "include": [
            "@redhat"
          ]
        },
        "data": [
          "oci::quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest",
          "github.com/release-engineering/rhtap-ec-policy//data"
        ],
        "name": "Default",
        "policy": [
          "github.com/enterprise-contract/ec-policies//policy/lib",
          "github.com/enterprise-contract/ec-policies//policy/release"
        ]
      }
    ]
  }
}' --output yaml
```

hack/update-rpm-lock.sh
```diff
@@ -0,0 +1,119 @@
...
+set -o pipefail
+set -o nounset
+
+root_dir=$(git rev-parse --show-toplevel)
+
+latest_release=$(gh api '/repos/konflux-ci/rpm-lockfile-prototype/tags?per_page=1' --jq '.[0].name')
+
+# build the image for running the RPM lock tool
+echo Building RPM lock tooling image...
+image=$(podman build --quiet --file <(cat <<DOCKERFILE
```
simonbaird, line 30: I never seen an inline Dockerfile build like this. I like it!

hack/update-rpm-lock.sh
```diff
@@ -0,0 +1,119 @@
...
+# Python version needs to match whatever version of Python dnf itself depends on
+FROM registry.access.redhat.com/ubi9/python-39:latest
+
+USER 0
```
simonbaird, line 34: Let's put a comment here to explain the reason at least.

hack/update-rpm-lock.sh
```diff
@@ -0,0 +1,119 @@
...
+)
+
+# copy the base image to temporary directory
+base_img_dir=$(mktemp -d --tmpdir)
+skopeo copy --quiet "docker://${base_img/:!(:)@/@}" "dir:/${base_img_dir}"
+
+# extract all /etc/yum.repos.d/* files from the base image
+for l in $(jq -r '\''.layers[].digest | sub("sha256:"; "")'\'' "${base_img_dir}/manifest.json"); do
+    tar --dir "${base_img_dir}" --extract --ignore-zeros 'etc/yum.repos.d/*' -f "${base_img_dir}/${l}"
+done
```
simonbaird, line 76: Would `oc extract` be suitable and nicer here?

hack/update-rpm-lock.sh
```diff
@@ -0,0 +1,119 @@
...
+podman run \
+    --rm \
+    --mount type=bind,source="${root_dir}/Dockerfile.dist",destination=/opt/app-root/src/Dockerfile \
+    --mount type=bind,source="${root_dir}/rpms.in.yaml",destination=/opt/app-root/src/rpms.in.yaml \
+    --mount type=bind,source="${root_dir}/rpms.lock.yaml",destination=/opt/app-root/src/rpms.lock.yaml \
+    "${image}" \
+    bash -c "${script}"
+
+# shellcheck disable=SC2094
+cat <<< "$(cat <<EOF
```
simonbaird, line 99: Maybe adding `rpms.lock.yaml` to `LICENSE_IGNORE` in `Makefile` would be acceptable rather than doing this hoop jumping to get the preamble in the generated file.

.tekton/cli-build.yaml
```diff
@@ -0,0 +1,391 @@
+apiVersion: tekton.dev/v1
```
simonbaird, line 1: I like this but would have been nice to review it in its own PR.

.tekton/cli-main-ci-pull-request.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
       == "main"
-  creationTimestamp: null
```
simonbaird, line 12: Are we confident that removing `creationTimestamp: null` is fine?

hack/bump-tekton-bundles.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Updates the Pipeline definitions in .tekton to the latest versions of Task bundles
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+root_dir=$(git rev-parse --show-toplevel)
+
+bundles="$(crane blob quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles@"$(crane manifest quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest | jq -r '.layers[0].digest')")"
```
simonbaird, line 26: I have a script like this where I used `ec inspect policy-data --source oci://${desired_prefix}/data-acceptable-bundles:latest` but crane is fine.

hack/bump-tekton-bundles.sh
```diff
@@ -0,0 +1,40 @@
...
+  cat <<< "$(yq eval-all '
+    select(fileIndex == 1).trusted_tasks as $t |
+    with(select(fileIndex == 0).spec.tasks[].taskRef.params[];
+      with(select(.name == "bundle");
+        (.value | sub("@.*", "")) as $r |
+        .value |= $r + "@" + $t["oci://" + $r][0].ref
+      )
+    ) |
+    select(fileIndex == 0)
+  ' "$f" <(echo "${bundles}"))" > "$f"
```
simonbaird, line 39: This might be a "yq magic of the month" award nominee. :grin:

simonbaird (approved):
I had a few suggestions and nitpicks, but no serious blockers.

hack/update-rpm-lock.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Updates the rpms.lock.yaml file
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+root_dir=$(git rev-parse --show-toplevel)
+
+latest_release=$(gh api '/repos/konflux-ci/rpm-lockfile-prototype/tags?per_page=1' --jq '.[0].name')
```
lcarva, line 26: Use [/releases/latest](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release) ?

hack/update-rpm-lock.sh
lcarva: +1 that would be simpler to maintain IMO. It would also ensure correctness as a layer could remove a file as well. `oc image extract` will take all of this into account.

(Maybe that approach wasn't taken due to limitations in getting the `oc` binary into the image.)

hack/update-rpm-lock.sh
```diff
@@ -0,0 +1,119 @@
...
+done
+
+cp "${base_img_dir}/etc/yum.repos.d"/*.repo /opt/app-root/src/
+
+# generate/update the RPM lock file
+/opt/app-root/bin/rpm-lockfile-prototype --outfile rpms.lock.yaml --image "${base_img}" rpms.in.yaml
+'
+
+echo Running RPM lock tooling...
+podman run \
```
lcarva, line 90: Note: You could use a `--volume` during the build process and simply execute this script from the Dockerfile, dumping the rpmlock file into a host volume. This would also avoid having to extract the yum repos from the base image directly (I think). 

(The approach you have is also perfectly fine.)

.tekton/cli-main-ci-pull-request.yaml
lcarva: I've done it in other places without any issues.

lcarva (approved):

zregvart:
Thanks for the review & suggestions. In the interest of getting the Pipeline to succeed I'll merge this now and I'll create followup PRs to include all suggestions here.

hack/update-rpm-lock.sh
zregvart: They don't use releases, only tags, the one release that is there is a very old one

hack/update-rpm-lock.sh
zregvart: :exploding_head:

.tekton/cli-build.yaml
zregvart: Yeah, I agree, I wanted to get this done...

hack/bump-tekton-bundles.sh
zregvart: Neat, I'll do that

hack/bump-tekton-bundles.sh
zregvart: It was one of those I can do this manually in 15min or write a script to do it in 2h

hack/update-rpm-lock.sh
zregvart: I didn't want to do that, there is a bit that parses the base image from the Dockerfile and a bit of fiddling of the `.repo` files. I wanted those to be in the script and variable based on the inputs (namely the `Dockerfile.dist`) and the tool image to be stable and not needing rebuilding.

PR #1764, opened 2024-07-17
Remove unused tests

zregvart:
We don't run these, I don't think we even maintain them, so let's remove them.


zregvart:
There are Task bundle updates needed to get a successful PipelineRun on #1762, so that would need to go in first...

lcarva:
> There are Task bundle updates needed to get a successful PipelineRun on #1762, so that would need to go in first...

I see that is merged. Rebase?

PR #1769, opened 2024-07-19
Humanize v-e-c Task logs

zregvart:
The top of the logfile for a TaskRun of verify-enterprise-contract Task should be the text (human readable) report. Following it are the YAML and JSON logs and preceded by a break `----- DEBUG OUTPUT -----` are the debug logs and version information.

This is to make the TaskRun's logs easier to understand by humans.

Also adds the `--logfile` parameter to redirect the logging to a file.

Reference: EC-564

### DIFF
--- a/cmd/root/root_cmd.go
+++ b/cmd/root/root_cmd.go
@@ -18,9 +18,11 @@ package root
 
 import (
 	"context"
+	"io"
 	"time"
 
 	hd "github.com/MakeNowJust/heredoc"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/enterprise-contract/ec-cli/internal/kubernetes"
@@ -35,6 +37,7 @@ var (
 	debug         bool = false
 	trace         bool = false
 	globalTimeout      = 5 * time.Minute
+	logfile       string
 )
 
 func NewRootCmd() *cobra.Command {
@@ -51,7 +54,7 @@ func NewRootCmd() *cobra.Command {
 		SilenceUsage: true,
 
 		PersistentPreRun: func(cmd *cobra.Command, _ []string) {
-			logging.InitLogging(verbose, quiet, debug, trace)
+			logging.InitLogging(verbose, quiet, debug, trace, logfile)
 
 			// Create a new context now that flags have been parsed so a custom timeout can be used.
 			ctx := cmd.Context()
@@ -60,6 +63,9 @@ func NewRootCmd() *cobra.Command {
 		},
 
 		PersistentPostRun: func(cmd *cobra.Command, _ []string) {
+			if f, ok := logrus.StandardLogger().Out.(io.Closer); ok {
+				f.Close()
+			}
 			if cancel != nil {
 				cancel()
 			}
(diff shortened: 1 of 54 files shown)


tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -160,6 +155,7 @@ spec:
 
     - name: validate
       image: quay.io/enterprise-contract/ec-cli:snapshot
+      onError: continue
```
lcarva, line 158: Will the Task still fail if there's an error, e.g. invalid policy config is provided? I think so.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -211,22 +210,36 @@ spec:
 
     - name: report
       image: quay.io/enterprise-contract/ec-cli:snapshot
-      command: [cat]
```
lcarva, line 214: Could've used `onError: continue` here and the other steps as well. Maybe a little bit cleaner?

lcarva (approved):

cmd/root/root_cmd.go
```diff
@@ -77,5 +83,6 @@ func setFlags(rootCmd *cobra.Command) {
 	rootCmd.PersistentFlags().BoolVar(&debug, "debug", debug, "same as verbose but also show function names and line numbers")
 	rootCmd.PersistentFlags().BoolVar(&trace, "trace", trace, "enable trace logging")
 	rootCmd.PersistentFlags().DurationVar(&globalTimeout, "timeout", globalTimeout, "max overall execution duration")
+	rootCmd.PersistentFlags().StringVar(&logfile, "logfile", "", "file to write the logging output")
```
simonbaird, line 86: Maybe: "file to write the logging output. If not specified logging output will be written to stderr"

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
simonbaird: What's the reason for adding this line? Maybe a comment is worthwhile.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -211,22 +210,36 @@ spec:
...
       args:
-        - "."
-        - "$(results.TEST_OUTPUT.path)"
+        - "----- DEBUG OUTPUT -----\n"
+
+    - name: version
+      image: quay.io/enterprise-contract/ec-cli:snapshot
+      command: [ec]
+      args:
+        - version
```
simonbaird, line 236: Maybe some extra debug:

```
      - name: images
        ...
          echo "$(params.IMAGES)"

      - name: config-source
        ...
           echo "($params.POLICY_CONFIGURATION)"

      - name: config
         ...
            jq .config < "$(params.HOMEDIR)/report-json.json"

Comment (lcarva) [reply tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml:236]: FYI, every step we add means a new container is created.

File: tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml (hunk 1 above)
Comment (zregvart) [reply tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml]: Added a comment, `ec` could fail, for example because bad input was provided and the rest of the steps would be skipped, including the output of debug logs so we would loose valuable clues as to how to debug the issue

**Review by lcarva — APPROVED** [review]

File: internal/applicationsnapshot/report.go [hunk 5]
```diff
@@ -159,6 +160,7 @@ func NewReport(snapshot string, components []Component, policy policy.Policy, da
 		Data:          data,
 		PolicyInput:   policyInput,
 		EffectiveTime: policy.EffectiveTime().UTC(),
+		ShowSuccesses: showSuccesses,
```
simonbaird, line 162: As discussed, should also have this one based on the --info flag:

```diff
+                      ShowInfo: showInfo,
```

Could be done later.

simonbaird (approved):

internal/applicationsnapshot/report.go
zregvart: Looking at the code that might be a bit more work, so I'll file a separate story about this

PR #1770, opened 2024-07-19
Inherit Konflux renovate config

lcarva:
This commit updates the renovate.json config to extend the renovate config provided by Konflux via mintmaker:
https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json

This feature is described here:
https://docs.renovatebot.com/config-presets/#github-hosted-presets

It also adds `tekton` to the list of enabled managers because the config in this repo overwrites that key in order to enable the `asdf` manager.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["github>konflux-ci/mintmaker//config/renovate/renovate.json"],
   "timezone": "America/New_York",
   "schedule": ["before 10pm"],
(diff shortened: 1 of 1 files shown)


lcarva:
I used `renovate --print-config true --platform local` to spot check that some of the configs from mintmaker are indeed taken into account. I also tried adding a bogus value to `extends` and verified that renovate complains about it.

simonbaird (approved):
Lgtm, let's try it.

lcarva:
I believe the Konflux renovate instance runs every two hours. So if there are new changes, we should see a new PR within that period.

PR #1814, opened 2024-07-31
feat(RHTAPWATCH-1039): mount custom certificate

yftacherzog:
Introducing optional parameters for referencing a config map containing CA bundle. The certificate is mounted to a location within the path the OS is taking certificates from.

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -91,6 +91,16 @@ spec:
         paths can be provided by using the `:` separator.
       default: ""
 
+    - name: CA_TRUST_CONFIGMAP_NAME
+      type: string
+      description: The name of the ConfigMap to read CA bundle data from.
+      default: trusted-ca
+
+    - name: CA_TRUST_CONFIG_MAP_KEY
+      type: string
+      description: The name of the key in the ConfigMap that contains the CA bundle data.
+      default: ca-bundle.crt
+
     - name: INFO
       type: string
       description: Include rule titles and descriptions in the output. Set to `"false"` to disable it.
@@ -207,6 +217,11 @@ spec:
           memory: 2Gi
         limits:
           memory: 2Gi
+      volumeMounts:
+        - name: trusted-ca
+          mountPath: /etc/pki/tls/certs/ca-custom-bundle.crt
+          subPath: ca-bundle.crt
+          readOnly: true
 
     - name: report
       image: quay.io/enterprise-contract/ec-cli:snapshot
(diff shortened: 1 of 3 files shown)


yftacherzog:
@zregvart hi,
Can you give me a hint here what this failure might be about?
In the logs I see it fails for:
```
Error: File was modified in build
Error: Process completed with exit code 1.
```

lcarva:
@yftacherzog, run `make generate` and add the files under `docs` that were updated to your commit.

yftacherzog:
> @yftacherzog, run `make generate` and add the files under `docs` that were updated to your commit.

Thanks!

I was running into the error below when running that command locally, but for now I was able to work around it by running this inside a container.

```
→ make generate 
❱ generate
go generate ./...
verifying github.com/enterprise-contract/go-gather/gather@v0.0.2: checksum mismatch
	downloaded: h1:wdg7DKN8sxYJeQhUZD/gZQTe6Dff4NIfQN8dZG2r7d8=
	go.sum:     h1:ndkHJHJemunUF6Ik8XG2u3sPHHOm8nNIiS2hj6jFPLA=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
make: *** [Makefile:45: generate] Error 1
```

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -259,3 +274,12 @@ spec:
         - >
           .result == "SUCCESS" or .result == "WARNING" or ($strict | not)
         - "$(results.TEST_OUTPUT.path)"
+
+  volumes:
```
lcarva, line 278: Any particular reason a volume, instead of a workspace, was used?

yftacherzog: No particular reason. Any best practices around this? I'm not sure what's the impact of using one vs. the other.

lcarva: I usually prefer a workspace because it provide more flexibility. While a volume is tied to a particular "storage" type, e.g. ConfigMap, a workspace is not. A workspace can be from a Secret, ConfigMap, PVC, emptyDir, etc. Although most of those would not be applicable in this case, I can see either Secret or ConfigMap being used.

More info here: https://tekton.dev/docs/pipelines/workspaces/

yftacherzog: Thanks for the explanation! I'll keep that in mind going forward.

PR #1815, opened 2024-07-31
fix: allow validate input to handle multiple source groups

mbestavros:
Resolves https://github.com/enterprise-contract/ec-cli/issues/1688.

Turns out that the [`NewInput`](https://github.com/enterprise-contract/ec-cli/blob/main/internal/evaluation_target/input/input.go#L38) function only returned the final source group in the policy. This patch makes the function return ALL source groups and changes the caller to evaluate each one, then append all to the output.

### DIFF
--- a/internal/input/validate.go
+++ b/internal/input/validate.go
@@ -46,14 +46,28 @@ func ValidateInput(ctx context.Context, fpath string, policy policy.Policy, deta
 		return nil, err
 	}
 
-	results, _, err := p.Evaluator.Evaluate(ctx, evaluator.EvaluationTarget{Inputs: inputFiles})
-	if err != nil {
-		log.Debug("Problem running conftest policy check!")
-		return nil, err
+	var allResults []evaluator.Outcome
+	for _, e := range p.Evaluators {
+		results, _, err := e.Evaluate(ctx, evaluator.EvaluationTarget{Inputs: inputFiles})
+		if err != nil {
+			log.Debug("Problem running conftest policy check!")
+			return nil, err
+		}
+		log.Debug("\n\nRunning conftest policy check\n\n")
+
+		if err != nil {
+			log.Debug("Problem running conftest policy check!")
+			return nil, err
+		}
+		allResults = append(allResults, results...)
 	}
 
 	log.Debug("Conftest policy check complete")
-	return &output.Output{PolicyCheck: results, Detailed: detailed}, nil
+
+	out := output.Output{Detailed: detailed}
+	out.SetPolicyCheck(allResults)
+
+	return &out, nil
 }
 
 // detect if a file or directory was passed. if a directory, gather all files in it
--- a/internal/input/validate_test.go
+++ b/internal/input/validate_test.go
@@ -64,13 +64,13 @@ func (e badMockEvaluator) CapabilitiesPath() string {
 
 func mockNewPipelineDefinitionFile(ctx context.Context, fpath []string, policy policy.Policy) (*input.Input, error) {
 	return &input.Input{
-		Evaluator: mockEvaluator{},
+		Evaluators: []evaluator.Evaluator{mockEvaluator{}},
 	}, nil
 }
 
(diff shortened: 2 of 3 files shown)


internal/input/validate.go
```diff
@@ -46,14 +46,28 @@ func ValidateInput(ctx context.Context, fpath string, policy policy.Policy, deta
...
 
-	results, _, err := p.Evaluator.Evaluate(ctx, evaluator.EvaluationTarget{Inputs: inputFiles})
-	if err != nil {
-		log.Debug("Problem running conftest policy check!")
-		return nil, err
+	var allResults []evaluator.Outcome
+	for _, e := range p.Evaluators {
+		results, _, err := e.Evaluate(ctx, evaluator.EvaluationTarget{Inputs: inputFiles})
+		if err != nil {
+			log.Debug("Problem running conftest policy check!")
```
zregvart, line 53: Could we also log the error?

internal/input/validate.go
```diff
@@ -46,14 +46,28 @@ func ValidateInput(ctx context.Context, fpath string, policy policy.Policy, deta
...
+			return nil, err
+		}
+		log.Debug("\n\nRunning conftest policy check\n\n")
+
+		if err != nil {
+			log.Debug("Problem running conftest policy check!")
```
zregvart, line 59: Let's log the error as well

zregvart (commented):
Could add unit/acceptance tests to add regression tests for this

lcarva (commented):
Fix looks reasonable. Let's make sure we add some tests (unit or acceptance) to prevent the behavior from breaking in the future.

internal/input/validate.go
simonbaird: +1

internal/input/validate.go
```diff
@@ -46,14 +46,28 @@ func ValidateInput(ctx context.Context, fpath string, policy policy.Policy, deta
...
-		log.Debug("Problem running conftest policy check!")
-		return nil, err
+	var allResults []evaluator.Outcome
+	for _, e := range p.Evaluators {
+		results, _, err := e.Evaluate(ctx, evaluator.EvaluationTarget{Inputs: inputFiles})
+		if err != nil {
+			log.Debug("Problem running conftest policy check!")
+			return nil, err
+		}
+		log.Debug("\n\nRunning conftest policy check\n\n")
```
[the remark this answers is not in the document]
simonbaird: Yeah, let's remove the extra blanks lines.

internal/input/validate_test.go
```diff
@@ -119,14 +119,14 @@ func Test_ValidatePipeline(t *testing.T) {
...
-			output:  &output.Output{PolicyCheck: []evaluator.Outcome{}},
+			output:  &output.Output{PolicyCheck: []evaluator.Outcome(nil)},
 			defFunc: mockNewPipelineDefinitionFile,
 		},
 		{
 			name:    "validation succeeds with yaml input",
 			fpath:   "kind: task",
 			err:     nil,
-			output:  &output.Output{PolicyCheck: []evaluator.Outcome{}},
+			output:  &output.Output{PolicyCheck: []evaluator.Outcome(nil)},
```
simonbaird, line 129: I'm curious about this diff, is it just a style change?

lcarva:
I created https://github.com/enterprise-contract/ec-cli/pull/1858 which includes the changes from the PR and addresses the comments.

internal/input/validate_test.go
mbestavros: It was necessary to allow tests to pass. AFAICT it's just a different type of "nothing" since before it was just an empty list.

internal/input/validate.go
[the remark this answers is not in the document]
mbestavros: I agree, but just for the sake of completeness, I pulled some of this code from `ec validate image` and that logging was there too.

mbestavros:
@lcarva's patch (#1858) is the way to go here, but I added some follow-up comments to clarify a few things just for the sake of completeness.

lcarva:
https://github.com/enterprise-contract/ec-cli/pull/1858 was merged. Let's close this.

PR #1817, opened 2024-08-01
KFLUXBUGS-1505 Fix EC OOM by specifying workers

robnester-rh:
This commit attempts to resolve an issue resulting in ec OOM when attempting to validate images with large SBOMs by providing a flag to limit the number of parallel works to reduce memory usage.


zregvart:
We were discussing a way for this to be 1 by default so we don't require a change in the release service, perhaps the Task's default can be 1?

lcarva:
> We were discussing a way for this to be 1 by default so we don't require a change in the release service, perhaps the Task's default can be 1?

Yes. Forgot about this. I think setting 1 in the Task (which is different than the default on the CLI, 5) makes sense given the resource constraints we set on the Task.

lcarva (approved):

docs/modules/ROOT/pages/ec_validate_image.adoc
```diff
@@ -150,6 +150,7 @@ mark (?) sign, for example: --output text=output.txt?show-successes=false
 --snapshot:: Provide the AppStudio Snapshot as a source of the images to validate, as inline
 JSON of the "spec" or a reference to a Kubernetes object [<namespace>/]<name>
 -s, --strict:: Return non-zero status on non-successful validation. Defaults to true. Use --strict=false to return a zero status code. (Default: true)
+--workers:: Number of workers to use for validation. Defaults to 5. (Default: 5)
```
simonbaird, line 153: What's the default though? :rofl:

lcarva: The first 5 is in Portuguese while the second 5 is in English.

PR #1843, opened 2024-08-09
Revert "chore(deps): update dependency golang to v1.22.6"

zregvart:
Reverts enterprise-contract/ec-cli#1837


zregvart:
I don't think we want to update golang to 1.22

lcarva:
I think this only impacts asdf, not the actual builds. But probably good to stay consistent.

PR #1873, opened 2024-08-19
Remove usage of cosign binary [0.4]

lcarva:


zregvart (approved):

lcarva:
There are concerns that some folks may have a copy of the Task definition which would still require the cosign binary in the image. We'll use EC-817 to come up with a different approach.

PR #1883, opened 2024-08-22
Add ec.oci.image_files rego function

### LINKED ISSUE #1788: Custom Rego function to fetch files from images
In policy we evaluate content of image files. For example the OLM bundle manifests or the CycloneDX SBOM files. For this we rely on [OCI file fetchers](https://github.com/enterprise-contract/ec-cli/tree/main/internal/fetchers/oci/files). For greater flexibility, that is, moving the logic that determines when and what files are extracted and provided as input, a custom Rego function (e.g. `ec.oci.file`) would be better. This way the logic of when to extract a file and what files to extract can be governed by the policy implementation.

A/C:
* Custom rego function exists to fetch files from images
* Image ref requires full digest (i.e. it's pinned)
* Sufficient test coverage
* The usual documentation is generated

lcarva:
Resolves #1788

### DIFF
--- a/acceptance/examples/oci_image_files.rego
+++ b/acceptance/examples/oci_image_files.rego
@@ -1,0 +1,24 @@
+package files
+
+import rego.v1
+
+
+# METADATA
+# custom:
+#   short_name: match
+deny contains result if {
+    files := ec.oci.image_files(input.image.ref, ["manifests"])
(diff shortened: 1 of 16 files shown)


zregvart (approved):
Nice!

simonbaird (approved):

PR #1889, opened 2024-08-26
Use source.config instead of configuration

lcarva:
The "global" `configuration` is deprecated. Instead, the `config` attribute for a certain policy source should be used.

This commit updates acceptance tests and docs to use the new format.

Ref: EC-688

### DIFF
--- a/cmd/validate/__snapshots__/image_test.snap
+++ b/cmd/validate/__snapshots__/image_test.snap
@@ -13,20 +13,20 @@
  "effective-time": "1970-01-01T00:00:00Z",
  "key": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAECBtqKHcvxYkGx7ZXqps3nrYS+ZSA\nmh3m1MZfTGlnr2oN0z+sBWEC23s4RkVSXkEydI6SLYatUtJK8OmiBRS+Xw==\n-----END PUBLIC KEY-----\n",
  "policy": {
-  "configuration": {
-   "exclude": [
-    "not_useful",
-    "test:conftest-clair"
-   ],
-   "include": [
-    "always_checked",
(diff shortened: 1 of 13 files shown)


cmd/validate/__snapshots__/image_test.snap
```diff
@@ -13,20 +13,20 @@
...
   "sources": [
    {
+    "config": {
+     "exclude": [
+      "not_useful",
+      "test:conftest-clair"
+     ],
+     "include": [
+      "always_checked",
+      "@salsa_one_collection"
```
simonbaird, line 27: Out of scope, but I wondered - is this an intentional or unintentional comedy misspelling?

cmd/validate/image_test.go
```diff
@@ -754,12 +752,10 @@ func Test_ValidateImageCommandExtraData(t *testing.T) {
...
           - registry1
-configuration:
-  collections:
-    - minimal
-  include:
-    - "*"
-  exclude: []
+    config:
+      include:
+        - '@minimal'
```
simonbaird, line 757: Not sure what the behavior is when both legacy "collections" and "include" are set, but should this be as follows?

```suggestion
        - '*'
```

simonbaird: Maybe it's inconsequential.

simonbaird (approved):
Thanks, that was a lot!

cmd/validate/__snapshots__/image_test.snap
lcarva: It looks like @zregvart added it in https://github.com/enterprise-contract/ec-cli/pull/1076 probably for the lolz.

cmd/validate/image_test.go
lcarva: That's a good question. Arguably, even the old example was a bit odd. I think this probably better captures the intent.

PR #1892, opened 2024-08-27
Escape curly braces in docs

lcarva:
Otherwise, asciidoctor assumes it's meant to be a reference and throws a warning:

> skipping reference to missing attribute

The warning causes the website build to halt.


lcarva:
@robnester-rh, ptal :pray:

lcarva:
Merging this as the website build is currently broken.

PR #1893, opened 2024-08-28
Change default validate image output to text

simonbaird:
The previous default was json.

There was one change in the unit tests required to make the tests pass. I'm not 100% sure why that was needed (or rather why it wasn't needed before), but my understanding is that the fs and the ctx was being reused between two test cases and it was causing problems somehow.

I did look at the task definition also, but the output is specified explicitly, so should be no change needed.

Ref: https://issues.redhat.com/browse/EC-758

### DIFF
--- a/internal/applicationsnapshot/report.go
+++ b/internal/applicationsnapshot/report.go
@@ -166,7 +166,7 @@ func NewReport(snapshot string, components []Component, policy policy.Policy, da
 // WriteAll writes the report to all the given targets.
 func (r Report) WriteAll(targets []string, p format.TargetParser) (allErrors error) {
 	if len(targets) == 0 {
-		targets = append(targets, JSON)
+		targets = append(targets, Text)
 	}
 	for _, targetName := range targets {
 		target, err := p.Parse(targetName)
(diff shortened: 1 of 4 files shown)


simonbaird:
For once I think I will pay attention to the CodeCov complaint.

simonbaird:
I fixed a few annoyances and added some more test coverage. Will take out of draft when it goes green.

simonbaird:
Now I'm pretty sure the CodeCov complaint is wrong. :thinking: :shrug:

internal/applicationsnapshot/report.go
```diff
@@ -166,7 +166,7 @@ func NewReport(snapshot string, components []Component, policy policy.Policy, da
 // WriteAll writes the report to all the given targets.
 func (r Report) WriteAll(targets []string, p format.TargetParser) (allErrors error) {
 	if len(targets) == 0 {
-		targets = append(targets, JSON)
+		targets = append(targets, Text)
```
simonbaird, line 169: The significant functional change.

lcarva (approved):

PR #1896, opened 2024-08-28
feat(KONFLUX-4136): add new reduce step

scoheb:
- Introducing **Single Component Mode**
  - See KONFLUX-4059

- This step, if activated by the SINGLE_COMPONENT parameter and a valid pipelinerun name, will attempt to reduce the incoming snapshot to contain a single component. That component being the one that caused the snapshot to create for.
- This is determined by examining the labels of the currently running pipelinerun:

```
appstudio.openshift.io/component: acb-service
test.appstudio.openshift.io/type: component
```

- This tells you that the underlying snapshot for this Test was created by a `component` that was recently built called `abc-service`

NOTE: This requires https://github.com/redhat-appstudio/infra-deployments/pull/4413

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -135,6 +135,16 @@ spec:
       description: Number of parallel workers to use for policy evaluation.
       default: "1"
 
+    - name: SINGLE_COMPONENT
+      description: Reduce the Snapshot to only the component whose build caused the Snapshot to be created
+      type: string
+      default: "false"
+
+    - name: PIPELINERUN_ID
+      description: Name of current PipelineRun.
+      type: string
+      default: "unknown"
+
   workspaces:
     - name: data
       description: The workspace where the snapshot spec json file resides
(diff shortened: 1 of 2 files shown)


simonbaird:
Not a review of this patch, (which looks pretty solid), but more about the approach taken:

When the (I presume) integration controller starts this pipeline and decides, based some some logic, to set the "SINGLE_COMPONENT" to "true", could it not also do the filtering of the IMAGES param then, and just pass the filtered list in the IMAGES pipeline param?

scoheb:
> Not a review of this patch, (which looks pretty solid), but more about the approach taken:
> 
> When the (I presume) integration controller starts this pipeline and decides, based some some logic, to set the "SINGLE_COMPONENT" to "true", could it not also do the filtering of the IMAGES param then, and just pass the filtered list in the IMAGES pipeline param?

There is a desire to not involve the controllers at this point due to time constraints.

I do agree that having the data model dictate this mode would be the best way forward. 

We may be going in that way in any event with recent talks in the Arch call...

simonbaird (approved):

simonbaird:
The acceptance tests exercise the task definition, so we can be confident there are no regressions.

That said, we might want to write some test coverage to confirm the single component does what we expect it to, but I think it can be done in a future PR.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -167,7 +184,28 @@ spec:
...
+    - name: reduce
+      params:
+        - name: IMAGES
+          value: $(params.IMAGES)
+        - name: SINGLE_COMPONENT
+          value: $(params.SINGLE_COMPONENT)
+        - name: CR_TO_QUERY
+          value: $(params.CR_TO_QUERY)
+#        - name: HOMEDIR
+#          value: $(params.HOMEDIR)
```
zregvart, line 199:
```suggestion
```

scoheb:
obsolete.

PR #1897, opened 2024-08-28
Replace cosign binary with bash wrapper [0.4]

simonbaird:
Same as #1873 but with an extra hack-cherry on top.

Ref: https://issues.redhat.com/browse/EC-817


lcarva:
Ha! I love it. If we're doing that approach, let's revert the change to the Task. This will tell us for user if the wrapper is working as intended.

lcarva (approved):

simonbaird:
I want to merge this a little after merging #2000 due to https://issues.redhat.com/browse/CLOUDWF-10542 .

PR #1899, opened 2024-08-29
No cosign with wrapper v02

simonbaird:
The v0.2 version of #1897.

Ref: https://issues.redhat.com/browse/EC-817


simonbaird:
Oops.

PR #1902, opened 2024-08-30
Update to go 1.22

lcarva:
Ref: EC-598

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.21.9
+golang 1.22.5
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@
 
 ## Build
 
-FROM docker.io/library/golang:1.21 AS build
+FROM docker.io/library/golang:1.22 AS build
 
 ARG TARGETOS
 ARG TARGETARCH
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -16,10 +16,9 @@
 
 ## Build
 
-# This works fine but will produce an EC violation
-# FROM docker.io/library/golang:1.21 AS build
-
-FROM registry.access.redhat.com/ubi9/go-toolset:1.21@sha256:ae17d73e70a966f39ef4dfca74241e3ca4374cd1198b02c30ea0748b8dcc83a6 AS build
+# Ideally, use the official image from Red Hat, e.g. registry.access.redhat.com/ubi9/go-toolset,
+# but a 1.22 release does not yet exist.
+FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:v1.22@sha256:859b38e6f53f49350402ca0a8dc342402f0afd8434c2f8984a7a3e714173cdca AS build
 
 ARG TARGETOS
 ARG TARGETARCH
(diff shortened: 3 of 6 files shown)


robnester-rh (approved):

.tool-versions
```diff
@@ -1 +1 @@
-golang 1.21.9
+golang 1.22.5
```
zregvart, line 1: ~~1.22.6 is newer, perhaps use that?~~ I see the `brew.registry.redhat.io/rh-osbs/openshift-golang-builder:v1.22` is at `go version go1.22.5 (Red Hat 1.22.5-1.el9) linux/amd64`, makes sense to be on the same patch version.

I guess we could enforce that all versions used here are matched in CI somehow? Would that be something of value?

Dockerfile
```diff
@@ -16,7 +16,7 @@
 
 ## Build
 
-FROM docker.io/library/golang:1.21 AS build
+FROM docker.io/library/golang:1.22 AS build
```
zregvart, line 19: Not sure if we want to, but we could pin to the patch version here...

```suggestion
FROM docker.io/library/golang:1.22.5 AS build
```

Dockerfile.dist
```diff
@@ -17,9 +17,11 @@
 ## Build
 
 # This works fine but will produce an EC violation
-# FROM docker.io/library/golang:1.21 AS build
+# FROM docker.io/library/golang:1.22 AS build
```
zregvart, line 20: Do we need these comments, perhaps its a good opportunity to remove them?

Dockerfile.dist
```diff
@@ -17,9 +17,11 @@
...
 
-FROM registry.access.redhat.com/ubi9/go-toolset:1.21@sha256:ae17d73e70a966f39ef4dfca74241e3ca4374cd1198b02c30ea0748b8dcc83a6 AS build
+# Ideally, use the official image from Red Hat, but a 1.22 release does not yet exist.
+# FROM registry.access.redhat.com/ubi9/go-toolset:1.21@sha256:ae17d73e70a966f39ef4dfca74241e3ca4374cd1198b02c30ea0748b8dcc83a6 AS build
```
zregvart, line 23: I'd remove this it is pointing to 1.21...

```suggestion
```

zregvart (approved):
:rocket:

Dockerfile
lcarva: I think it's ok to just use `1.22`. Go is usually pretty stable between patch releases.

.tool-versions
lcarva: Yeah. Not sure if we can build the project if the version on the builder image is older (even just at the patch version). Need to play with it further.

Dockerfile.dist
lcarva: Fair. I re-wrote the comment and remove that specific image ref.

Dockerfile.dist
lcarva: Good idea.

.tool-versions
lcarva: Following up in https://github.com/enterprise-contract/ec-cli/pull/1917.

> I guess we could enforce that all versions used here are matched in CI somehow? Would that be something of value?

Maybe let's see if not doing that causes any issues?

PR #1917, opened 2024-09-03
Bump all dependencies

lcarva:
Now that EC cli supports go 1.22, I ran `go get -u ./...` on all the modules in this repo. This is the result.


lcarva:
The linter faillure seems to be bogus... AFAICT the code is doing exactly what the docs are recommending.

simonbaird:
Commit message typo fyi: TerminiationReason

simonbaird:
You could be a little more verbose about the G115 since it's security related, but otherwise lgtm.

simonbaird (approved):

PR #1918, opened 2024-09-03
Add kubectl to the ec-cli image

scoheb:
- kubectl is needed in order to inspect labels on resources

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ ARG TARGETOS
 ARG TARGETARCH
 ARG BUILD_SUFFIX=""
 ARG BUILD_LIST="${TARGETOS}_${TARGETARCH}"
+ARG KUBECTL_VERSION=1.27.2
 
 # Avoid safe directory git failures building with default user from go-toolset
 USER root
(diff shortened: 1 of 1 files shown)


zregvart:
I don't think we can do this. The image built here is multiarch, and we support a bunch of these (see https://github.com/enterprise-contract/ec-cli/blob/d12edd162458b07ecbab38fb1ba3e08d248fc46f/Dockerfile.dist#L26) while this downloads kubectl for Linux on amd64 in all cases. I see there is no kubectl for most of these platforms.

Ideally we'd like to only do a `COPY` in the last stage, curl should be invoked in the build stage. We'd also like to control the version of kubectl via tools/go.mod (as we do for other tools that we use), this way we get version bumps there via dependabot/renovate bot.

Let me fiddle with this for a bit and try to come up with an alternative.

zregvart:
> Building kubectl from source is probably the best approach for an open source project. Not sure how expensive it is to build it though.

I have a thing, doesn't seem too expensive, let me get the my ducks in a row, I'll create a PR shortly...

scoheb:
closed in favour of #1922

PR #1919, opened 2024-09-04
EC-806 Ensure ec uses go-gather by default

robnester-rh:
This commit removes the check for the env var `USEGOGATHER` and instead has ec use the go-gather downloader. As we've previously verified that the data downloaded by go-gather matches the information used via the prior downloader implementation, the corresponding downloader code has been removed in `internal/downloader/downloader.go`. The associated tests in `internal/downloader/downloader_test.go` has been updated. Additionally, since we're no longer using branching logic to determine if we should use go-gather for downloading, the `UseGoGather` function in `internal/utils/helpers.go` has been removed.

### DIFF
--- a/internal/downloader/downloader.go
+++ b/internal/downloader/downloader.go
@@ -23,14 +23,10 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
-	"sync"
 
 	"github.com/enterprise-contract/go-gather/gather"
 	"github.com/enterprise-contract/go-gather/metadata"
-	"github.com/open-policy-agent/conftest/downloader"
 	log "github.com/sirupsen/logrus"
-
-	"github.com/enterprise-contract/ec-cli/internal/utils"
 )
 
 type key int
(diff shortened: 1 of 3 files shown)


internal/downloader/downloader.go
```diff
@@ -66,31 +60,11 @@ func Download(ctx context.Context, destDir string, sourceUrl string, showMsg boo
 	}
 
 	dl := func(ctx context.Context, sourceUrl, destDir string) (metadata.Metadata, error) {
-		// conftest's Download function leverages oras under the hood to fetch from OCI. It uses the
-		// global oras client and sets the user agent to "conftest". This is not a thread safe
-		// operation. Here we get around this limitation by ensuring a single download happens at a
```
lcarva, line 71: Wow I had forgotten about this...

internal/downloader/downloader.go
```diff
@@ -66,31 +60,11 @@ func Download(ctx context.Context, destDir string, sourceUrl string, showMsg boo
...
-			}
-			return nil, err
+		m, err := gatherFunc(ctx, sourceUrl, destDir)
+		if err != nil {
+			log.Debug("Download failed!")
 		}
+		return m, err
 	}
 
 	m, err := dl(ctx, sourceUrl, destDir)
```
lcarva, line 70: We probably don't need to create a `dl` wrapper function and can just call `gatherFunc` directly here.

lcarva (approved):

internal/downloader/downloader.go
simonbaird: +1, wdyt Rob.

zregvart (approved):
LGTM, but pay heed to the suggestion from Luiz

lcarva (approved):

PR #1922, opened 2024-09-04
Add kubectl tool

zregvart:


zregvart:
We might consider something akin to this in lieu of #1918. cc @scoheb

lcarva:
Nice!

Dockerfile
```diff
@@ -32,6 +32,10 @@ WORKDIR /build
 COPY go.mod go.sum ./
 RUN go mod download
 
+# Copy the tools mod file for better layer caching when building locally
+COPY tools/go.mod tools/go.sum ./tools/
+RUN cd tools && go mod download
```
lcarva, line 37: To make this happen we need an additional entry here: https://github.com/enterprise-contract/ec-cli/blob/d12edd162458b07ecbab38fb1ba3e08d248fc46f/.tekton/cli-main-ci-pull-request.yaml#L32-L33

Maybe we need another sub-module, or just add this to the one in `/`? Not sure if we want the SBOM of our image to include everything that is in under `tools`.

scoheb:
looks good to me!

Dockerfile
zregvart: Opted to do a separate golang module. TIL: we already have `tools/go.mod` and `acceptance/go.mod` in the SBOM, see #1924

zregvart:
Ready for review now...

Makefile
```diff
@@ -1,4 +1,4 @@
-MAKEFLAGS+=-j --no-print-directory
+MAKEFLAGS+=--no-print-directory
```
lcarva, line 1: What's this for?

lcarva (approved):

PR #1923, opened 2024-09-04
Don't produce error output for invalid purl checks

simonbaird:
I noticed the error output in when running the unit test in https://github.com/enterprise-contract/ec-policies/pull/1126

It seems like this should not be an error. Since the purpose of this method is to decide if the purl is valid, quietly returning false if it is invalid would be all that it should do. Logging an error seems unexpected.

(For the parse functon I didn't change it. My reasoning is that if you call parse then you're expecting it to be parseable. If it fails at that point then it's unexpected, so logging an error does seem reasonable in that case.)

Tangentially related to...
Ref: https://issues.redhat.com/browse/EC-847

### DIFF
--- a/acceptance/examples/purl.rego
+++ b/acceptance/examples/purl.rego
@@ -18,6 +18,18 @@ deny contains result if {
 
 # METADATA
 # custom:
+#   short_name: parse_fail
+deny contains result if {
+    # This will emit a failure too but this time an error is logged
+    not ec.purl.parse(_bad)
+    result := {
+        "code": "purl.parse_fail",
+        "msg": sprintf("PURL can't be parsed %q", [_bad])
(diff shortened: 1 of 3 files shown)


lcarva (approved):
Using debug sounds right. It would be great to ensure the log message still appears though. I want CI to fail if we stop logging for some reason since the errors are always silenced for the rego functions.

zregvart (approved):

simonbaird:
I think acceptance tests are passing locally for me, so failure was unexpected.

lcarva:
Could be a fluke. Clicked on re-run.

simonbaird:
Added one more commit for a little extra test coverage for the remaining error log as suggested.

PR #1924, opened 2024-09-05
Configure Syft to exclude testing and tooling deps

zregvart:


.syft.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright The Enterprise Contract Contributors
```
lcarva, line 1: Technically, this should go after the `---` otherwise yaml parsers will think that this section is its own `null` yaml document.

.syft.yaml
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+---
+exclude:
+  - ./acceptance/**
+  - ./internal/documentation/**
```
lcarva, line 20: Not sure if we should be this selective or if it'll matter. Because we feed `/go.mod` to cachi2, it's just going to record all the deps from it anyways. Maybe we just exclude modules completely instead?

.syft.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - ./internal/mocks/**
+  - ./tools/*
```
lcarva, line 22: Need to exclude `/tools/kubectl`?

zregvart: That should be in SBOM, its in the image, right?

.syft.yaml
zregvart: Huh, not sure, there is SBOM generated by cachi2 and that is based on the configuration, the Syft is instructed to scan the image and source; all of these are merged afterwards and we end up with the SBOM containing the acceptace/tools stuff

.syft.yaml
lcarva: I meant that we should exclude that from the exclude list. IOW, we should make sure `/tools/kubectl` is included in the SBOM. :rofl:

zregvart: Yeah, `/usr/local/bin/kubectl` is in there (I guess from the image), but not `tools/kubectl/go.mod`

zregvart:
Trims down the packages found from the source from 1323 to 452, and includes only the dependencies found in `/go.mod` and `/tools/kubectl/go.mod`.

lcarva (approved):

PR #1968, opened 2024-09-17
feat(KONFLUX-4136): new reduce-snapshot script

scoheb:
- add new script to ec-cli image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,6 +72,9 @@ COPY --from=build "/build/dist/ec_${TARGETOS}_${TARGETARCH}" /usr/local/bin/ec
 # Copy the one kubectl binary that can run in this container
 COPY --from=build "/build/dist/kubectl_${TARGETOS}_${TARGETARCH}" /usr/local/bin/kubectl
 
+# Copt reduce-snapshot script needed for single component mode
+COPY hack/reduce-snapshot.sh /usr/local/bin
+
 # OpenShift preflight check requires a license
 COPY --from=build /build/LICENSE /licenses/LICENSE
 
--- a/hack/reduce-snapshot.sh
+++ b/hack/reduce-snapshot.sh
@@ -64,8 +64,6 @@ if [ "${SINGLE_COMPONENT}" == "true" ]; then
     fi
 
     echo "Reducing Snapshot to:"
-    echo "$REDUCED_SNAPSHOT" | jq .
-    SNAPSHOT=$(echo "$REDUCED_SNAPSHOT" | tr -d ' ' | tr -d '\n')
-    echo "$SNAPSHOT" | jq -c . > "${SNAPSHOT_PATH}"
+    echo "$REDUCED_SNAPSHOT" | jq '.' | tee "${SNAPSHOT_PATH}"
   fi
 fi
--- a/hack/reduce-snapshot.sh
+++ b/hack/reduce-snapshot.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+# Copyright The Enterprise Contract Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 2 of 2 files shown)


hack/reduce-snapshot/reduce-snapshot.sh
```diff
@@ -0,0 +1,54 @@
...
+      set +e
+      jq -e < "${SNAPSHOT}" > /dev/null 2>&1
+      CHECK_JSON_FILE=$?
+      set -e
+      if [ "${CHECK_JSON_FILE}" != "0" ]; then
+        echo "Cannot load snapshot from json string or file"
+        exit 1
+      fi
+      SNAPSHOT=$(cat "${SNAPSHOT}")
+    fi
```
zregvart, line 36: I think the `jq` commands here lack the filter. I think something simpler can be done here instead:

```suggestion
    SNAPSHOT="$(cat "${SNAPSHOT}" 2> /dev/null || echo "${SNAPSHOT}")"
    [[ ! "${SNAPSHOT}" =~ ^\s*\{ ]] && echo 'Error: Cannot load snapshot from JSON string or file' && exit 1
```

hack/reduce-snapshot/reduce-snapshot.sh
```diff
@@ -0,0 +1,54 @@
...
+    if [ "${COMPONENT_COUNT}" != "1" ] ; then
+      echo "Error: Reduced Snapshot has ${COMPONENT_COUNT} components. It should contain 1"
+      echo "       Verify that the Snapshot contains the built component: ${SNAPSHOT_CREATION_COMPONENT}"
+      exit 1
+    fi
+
+    echo "Reducing Snapshot to:"
+    echo "$REDUCED_SNAPSHOT" | jq .
+    SNAPSHOT=$(echo "$REDUCED_SNAPSHOT" | tr -d ' ' | tr -d '\n')
+    echo "$SNAPSHOT" > "${SNAPSHOT_PATH}"
```
zregvart, line 52: Might be dangerous if any of the keys or values in the JSON contain the space character, why not something like this:

```suggestion
    echo "$SNAPSHOT" | jq -c . > "${SNAPSHOT_PATH}"
```

scoheb: fixed. thanks

hack/reduce-snapshot/reduce-snapshot.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+set -eux
```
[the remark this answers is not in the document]
scoheb: fixed. thanks

hack/reduce-snapshot/reduce-snapshot.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
scoheb: done. thanks

hack/reduce-snapshot/reduce-snapshot.sh
lcarva: I see this now:
```
echo "$REDUCED_SNAPSHOT" | jq .
SNAPSHOT=$(echo "$REDUCED_SNAPSHOT" | tr -d ' ' | tr -d '\n')
echo "$SNAPSHOT" | jq -c . > "${SNAPSHOT_PATH}"
```

I think @zregvart meant this:

```
# for displaying
echo "$REDUCED_SNAPSHOT" | jq .
# for saving
echo "$REDUCED_SNAPSHOT" | jq -c . > "${SNAPSHOT_PATH}"
```

Honestly, I would just do something like this:

```
echo "$REDUCED_SNAPSHOT" | jq '.' | tee "${SNAPSHOT_PATH}"
```

I don't really think we need to save the reduced snapshot as a compacted json.

scoheb: ok adjusted. but I seem to remember an issue whereby I needed to remove the line feeds...

lcarva (approved):

zregvart (approved):

Dockerfile
```diff
@@ -72,6 +72,9 @@ COPY --from=build "/build/dist/ec_${TARGETOS}_${TARGETARCH}" /usr/local/bin/ec
 # Copy the one kubectl binary that can run in this container
 COPY --from=build "/build/dist/kubectl_${TARGETOS}_${TARGETARCH}" /usr/local/bin/kubectl
 
+# Copt reduce-snapshot script needed for single component mode
+COPY hack/reduce-snapshot.sh /usr/local/bin
```
simonbaird, line 76: For tidiness I'd probably suggest:

```
COPY --from=build /build/hack/reduce-snapshot.sh /usr/local/bin
```

simonbaird: (Not super important, also I realize I'm too slow and its merged already.)

PR #1969, opened 2024-09-17
feat(KONFLUX-4136): add new reduce step

scoheb:
- add new step to reduce snapshot for single component mode
- requires #1968

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -135,6 +135,25 @@ spec:
       description: Number of parallel workers to use for policy evaluation.
       default: "1"
 
+    - name: SINGLE_COMPONENT
+      description: Reduce the Snapshot to only the component whose build caused the Snapshot to be created
+      type: string
+      default: "false"
+
+    - name: SINGLE_COMPONENT_CUSTOM_RESOURCE
+      description: >
+        Name, including kind, of the Kubernetes resource to query for labels when single
+        component mode is enabled, e.g. pr/somepipeline.
+      type: string
+      default: "unknown"
+
+    - name: SINGLE_COMPONENT_CUSTOM_RESOURCE_NS
+      description: >
+        Kubernetes namespace where the SINGLE_COMPONENT_NAME is found. Only used
+        when single component mode is enabled.
+      type: string
+      default: ""
+
   workspaces:
     - name: data
       description: The workspace where the snapshot spec json file resides
@@ -168,6 +187,21 @@ spec:
         - name: TUF_MIRROR
           value: "$(params.TUF_MIRROR)"
 
+    - name: reduce
+      env:
+        - name: SNAPSHOT
+          value: $(params.IMAGES)
+        - name: SINGLE_COMPONENT
+          value: $(params.SINGLE_COMPONENT)
+        - name: CUSTOM_RESOURCE
+          value: $(params.SINGLE_COMPONENT_CUSTOM_RESOURCE)
+        - name: CUSTOM_RESOURCE_NAMESPACE
+          value: $(params.SINGLE_COMPONENT_CUSTOM_RESOURCE_NS)
+        - name: SNAPSHOT_PATH
+          value: $(params.HOMEDIR)/snapshot.json
+      image: quay.io/enterprise-contract/ec-cli:snapshot
+      onError: continue # progress even if the step fails so we can see the debug logs
+      command: [reduce-snapshot.sh]
     - name: validate
       image: quay.io/enterprise-contract/ec-cli:snapshot
       onError: continue # progress even if the step fails so we can see the debug logs
@@ -177,7 +211,7 @@ spec:
         - image
         - "--verbose"
         - "--images"
-        - "$(params.IMAGES)"
+        - "/tekton/home/snapshot.json"
         - "--policy"
         - "$(params.POLICY_CONFIGURATION)"
         - "--public-key"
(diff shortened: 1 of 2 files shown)


tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -177,7 +212,7 @@ spec:
         - image
         - "--verbose"
         - "--images"
-        - "$(params.IMAGES)"
+        - "$(params.HOMEDIR)/snapshot.json"
```
lcarva, line 215: Should this be `$(params.SNAPSHOT_PATH)`?

Also, not sure if we want to expose this as a parameter to the user. It seems more like an implementation detail.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -135,6 +135,26 @@ spec:
       description: Number of parallel workers to use for policy evaluation.
       default: "1"
 
+    - name: SINGLE_COMPONENT
+      description: Reduce the Snapshot to only the component whose build caused the Snapshot to be created
+      type: string
+      default: "false"
+
+    - name: CUSTOM_RESOURCE
```
lcarva, line 143: The name and description of this parameter, and the one below, should reflect that is associated with `SINGLE_COMPONENT` mode. Consider:

```yaml
- name: SINGLE_COMPONENT_NAME
  description: >
    Name, including kind, of the Kubernetes resource to query for labels when single
    component mode is enabled, e.g. pr/somepipeline.
- name: SINGLE_COMPONENT_NAMESPACE
  description: >
    Kubernetes namespace where the SINGLE_COMPONENT_NAME is found. Only used 
    when single component mode is enabled.
```

simonbaird: Agreed.

simonbaird: Maybe `SINGLE_COMPONENT_CR` and `SINGLE_COMPONENT_NS`.

simonbaird: Is it ever not a pipeline run? Maybe SINGLE_COMPONENT_PR ?

simonbaird: Since you need to specify the kind, the implication is it might be something other than a PipelineRun. Is that the intention?

scoheb: If we have to change it, then I would prefer `SINGLE_COMPONENT_CUSTOM_RESOURCE` and `SINGLE_COMPONENT_CUSTOM_RESOURCE_NAMESPACE` 

> Is it ever not a pipeline run? Maybe SINGLE_COMPONENT_PR ?

Yes in Release Pipelines, it is a `Snapshot`

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
scoheb: for `Integration Test Scenarios`, SNAPSHOT is the json representation of the snapshot.
for `Release Pipelines`, SNAPSHOT will be the path to the json file.

With my latest PR #1972 , I am always outputting a JSON file that will either contain the **complete** `Snapshot` or the **reduced** one.

so yes, once that PR merges, I can use `$(params.SNAPSHOT_PATH)`.

> It seems more like an implementation detail

Are you meaning to say it should not a param, and I am simply hard-code `/tekton/home/snapshot.json`?

simonbaird:
I think the acceptance test failure is legit: `unable to parse Snapshot specification from /tekton/home/snapshot.json: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type v1alpha1.SnapshotSpec`

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
simonbaird: > I would prefer `SINGLE_COMPONENT_CUSTOM_RESOURCE` and `SINGLE_COMPONENT_CUSTOM_RESOURCE_NAMESPACE`

Works for me.

simonbaird: It's quite long I guess. `SINGLE_COMPONENT_CUSTOM_RESOURCE_NS` maybe?

scoheb: I am ok with `SINGLE_COMPONENT_CUSTOM_RESOURCE` and `SINGLE_COMPONENT_CUSTOM_RESOURCE_NS`

scoheb:
> I think the acceptance test failure is legit: `unable to parse Snapshot specification from /tekton/home/snapshot.json: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type v1alpha1.SnapshotSpec`

yes. hence #1972

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
lcarva: > Are you meaning to say it should not a param, and I am simply hard-code /tekton/home/snapshot.json?

Yeah, exactly.

scoheb: removed parameter and using the already defined `HOME` parameter in the arg

lcarva (approved):

lcarva:
Let's just get the commits organized to better represent the change before merging.

simonbaird (approved):
Lgtm, but please do a squash as suggested elsewhere.

PR #1980, opened 2024-09-18
WIP: EC-802

robnester-rh:
This is a WIP


robnester-rh:
After a conversation with @lcarva this will be abandoned and an alternate approach taken.

PR #1993, opened 2024-09-23
Common tracing and retry/backoff support for go-gather

zregvart:
More details in the commit messages. Adds a set of common variables used to configure the HTTP clients used by go-containerregistry and go-gather HTTP and OCI support via ORAS.

Requires https://github.com/enterprise-contract/go-gather/pull/114 and https://github.com/enterprise-contract/go-gather/pull/113 to be merged and versions to be bumped here.


simonbaird (commented):
I gave it a quick read. Lgtm, but maybe seek another review from a golang expert.

internal/downloader/downloader.go
```diff
@@ -37,7 +43,41 @@ type downloadImpl interface {
...
+var log = logrus.StandardLogger()
+
+var gatherFunc = func(ctx context.Context, source, destination string) (metadata.Metadata, error) {
+	initialize()
+	return gather.Gather(ctx, source, destination)
+}
+
+var _initialize = func() {
+	if log.IsLevelEnabled(logrus.TraceLevel) {
+		goci.Transport = http.NewTracingRoundTripperWithLogger(goci.Transport, log)
```
lcarva, line 55: As a follow up, we should use a solution that doesn't rely on global states. Besides being a safer approach it would also remove the need to use `sync`. I understand that that would require changes in go-gather itself. Could you file a follow up issue?

internal/downloader/downloader.go
```diff
@@ -37,7 +43,41 @@ type downloadImpl interface {
...
 
-var gatherFunc = gather.Gather
+var log = logrus.StandardLogger()
+
+var gatherFunc = func(ctx context.Context, source, destination string) (metadata.Metadata, error) {
+	initialize()
+	return gather.Gather(ctx, source, destination)
+}
+
+var _initialize = func() {
```
lcarva, line 53: Could this be done during `init` to avoid having to use `sync.OnceFunc`?

lcarva (approved):

internal/downloader/downloader.go
zregvart: Could not control it in tests then

PR #2005, opened 2024-09-23
Keep go-containerregistry up to date

lcarva:
Ref: EC-778

### DIFF
--- a/.github/workflows/update-go-containerregistry.yaml
+++ b/.github/workflows/update-go-containerregistry.yaml
@@ -1,0 +1,71 @@
+---
+# Copyright The Enterprise Contract Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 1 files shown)


simonbaird (approved):
I didn't look too closely, but should be fine to merge early and often.

lcarva:
That didn't quite work. But at least now I can try things out [here](https://github.com/enterprise-contract/ec-cli/compare/fix-app-id?expand=1).

Current failure: https://github.com/enterprise-contract/ec-cli/actions/runs/11002493835/job/30549573454

I thought that was due to a bad app ID, but AFAICT, I am now providing the right one. I'll keep digging.

lcarva:
It works!

https://github.com/enterprise-contract/ec-cli/pull/2007

https://github.com/enterprise-contract/ec-cli/actions/runs/11002747801/job/30550367225

But! It didn't trigger any of the checks... It's probably missing some permissions somewhere.

.github/workflows/update-go-containerregistry.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: update
+        run: |
+          go mod edit -replace github.com/google/go-containerregistry=github.com/enterprise-contract/go-containerregistry@main
+          go mod tidy
+        env:
+          # Hack to ensure this repo is always updated, see:
+          #   https://github.com/golang/go/issues/45413
+          GOPRIVATE: github.com/enterprise-contract/go-containerregistry
+
+      - uses: actions/create-github-app-token@5d869da34e18e7287c1daad50e0b8ea0f506ce69 # v1.11.0
```
zregvart, line 54: TIL

lcarva:
> But! It didn't trigger any of the checks... It's probably missing some permissions somewhere.

Yeah because I ran the workflow from a branch (fix-app-id) so the PR was created targeting that branch. Since our CI mostly targets the `main` branch, there was nothing to run.

PR #2006, opened 2024-09-23
Revise the "cut release" doc script

simonbaird:
- Remove scripts to apply patches from main branch pipelines to the new pipelines.
- Describe the new plan, which is to just use the main branch pipelines instead of merging in the pipelines from the PR that Konflux creates.
- Describe how to make a PR in tenants-config to create the application, component, etc. (Note that we've never done it this way before, so it's a little speculative.)
- Add a couple of steps that were previously not listed.
- Various other changes along those lines.

Ref: https://issues.redhat.com/browse/EC-819


simonbaird:
This is a little scrappy, but should be good enough for the `hacks` directory. :grin: .

zregvart (approved):

PR #2008, opened 2024-09-24
Use errors in place of hashicorp/go-multierror

### LINKED ISSUE #1971: Replace the use of github.com/hashicorp/go-multierror with errors.Join
We don't really need github.com/hashicorp/go-multierror any more, from golang 1.20 we can use [errors.Join](https://pkg.go.dev/errors#Join) instead. One less dependency to worry about...

zregvart:
Replaces the use of hashicorp/go-multierror with errors.Join. The formatting changes aside it is very straightforward change.

Fixes #1971

### DIFF
--- a/features/__snapshots__/validate_image.snap
+++ b/features/__snapshots__/validate_image.snap
@@ -1276,10 +1276,7 @@ Error: success criteria not met
 ---
 
 [happy day with missing git config:stderr - 1]
-Error: 1 error occurred:
-    * no suitable config file found at git::https://${GITHOST}/git/happy-config.git
-
-
+Error: no suitable config file found at git::https://${GITHOST}/git/happy-config.git
 
 ---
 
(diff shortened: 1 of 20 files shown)


simonbaird (approved):
Lgtm.

features/__snapshots__/validate_image.snap
```diff
@@ -1276,10 +1276,7 @@ Error: success criteria not met
 ---
 
 [happy day with missing git config:stderr - 1]
-Error: 1 error occurred:
```
lcarva, line 1279: What does it look like if there is more than one error? Are they joined by `\n`?

lcarva (approved):

features/__snapshots__/validate_image.snap
zregvart: Yeah, I missed some failing tests that had more than one error added in [`2bd4399` (#2008)](https://github.com/enterprise-contract/ec-cli/pull/2008/commits/2bd43998bac0891cd6dbad0a44876276f0f67050). For example this: [`2bd4399` (#2008)](https://github.com/enterprise-contract/ec-cli/pull/2008/commits/2bd43998bac0891cd6dbad0a44876276f0f67050#diff-e24f93dbfa8e364e10dfe82831576305eb4a9e9c83eb621e06c7fbf489de695fR2042-R2044) snapshot has three errors joined

lcarva (approved):

PR #2017, opened 2024-09-26
Use leaf of package name

lcarva:
The idea is that, for the most part, we only care about the last segment of a package name. For example, if the rego file defines:

```rego
package policy.release.foo
```

`foo` is the only meaningful name when it comes to inclusion and exclusions. `policy.release` is ignored in that context. As such, we should strip that prefix everywhere to avoid confusion.


lcarva:
This is in draft because the implications are not very clear to me yet. Also not sure if we should go all in regarding manipulating the package name, or maybe take the opposite direction of simply using the full package name everywhere.

What is clear to me is that using a mix of both is not a good way forward.

zregvart:
I think we should be using full package names, there might be a case where the shortened package name overlaps in two or more packages, fictional example: `policy.image.signature` and `policy.attestation.signature`. That we do any shortening at all became of the way we structured the policy rules in ec-policies, which we might/will restructure anyhow.

lcarva:
It's a good point. Any shortening of the package name increases the risk of unintentional overlap. Name collisions can still happen when the full package name is used though.

However, we should be consistent. We either shorten the package name everywhere, or we use it as is everywhere. Any sort of in-between approach (which is what we currently have) is bound to cause confusion and unintended consequences.

The only exception is the `data.` prefix added by rego. That should not be included in the name of the package.

One of the implications of using the full package name is how we handle sub-packages. For example, will I be able to simply include `policy` to mean `policy.image.signature` and `policy.attestation.signature`? What if there are rules defined directly at `policy` and I want to include just those? I think we'd have to introduce some sort of explicit wildcard where `policy` is a subset of `policy.*`. We'd also have to tweak the specificity code handle this, e.g. `policy.foo.*` is more specific than `policy.*`.

On the other hand, we have written a lot of policy rules and have never come across a case where using just the last segment of the full package name has caused a problem. (Granted that we do have a [check](https://github.com/enterprise-contract/ec-policies/blob/1c21485ac210058932cea9e8bc66588c9d4a9340/checks/annotations.rego#L115-L131) to ensure this doesn't happen.) With policy source groups the risk of overlap is even lower as I wouldn't mix different sets of policy rules from different sources in the same group.

I think there's an argument somewhere that a short package name is better for policy config authors and users. Conversely, there's probably also an argument that using the full package name is the most correct form so we should just adopt that.

lcarva:
Closing it out. This was a fun experiment. We have decided to go with using the full package name going forward.

PR #2023, opened 2024-09-26
Remove restriction when fetching blobs

lcarva:
The custom rego function `ec.oci.blob` limits downloads to 10MB. This is often used to fetch SBOMs which could be larger. This commit removes the restrictions altogether and instead relies on the constraints set on whichever environment is running EC.

Ref: KFLUXBUGS-1667

### DIFF
--- a/internal/rego/oci/oci.go
+++ b/internal/rego/oci/oci.go
@@ -177,8 +177,6 @@ func registerOCIImageFiles() {
 	rego.RegisterBuiltin2(&decl, ociImageFiles)
 }
 
-const maxBytes = 10 * 1024 * 1024 // 10 MB
-
 func ociBlob(bctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
 	uri, ok := a.Value.(ast.String)
 	if !ok {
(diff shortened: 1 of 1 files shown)


internal/rego/oci/oci.go
```diff
@@ -177,8 +177,6 @@ func registerOCIImageFiles() {
 	rego.RegisterBuiltin2(&decl, ociImageFiles)
 }
 
-const maxBytes = 10 * 1024 * 1024 // 10 MB
```
nickboldt, line 180: or just bump this to 30 MB?

lcarva: That seem arbitrary as well.

I wonder if making this configurable, e.g. via env var, is a better approach.

simonbaird: What was the thinking behind adding a limit initially? Maybe there's still some value in it, even if it's bumped to 100MB or something.

simonbaird (approved):

zregvart (approved):

internal/rego/oci/oci.go
lcarva: Let's try out without any restrictions. We can always go back later on if we find a good reason.

PR #2024, opened 2024-09-26
Few cleanups on go-containerregistry workflow

lcarva:
- Use a variable for the EC App ID
- Add documentation on why a GH App is used


zregvart (approved):

zregvart:
Make sure to add the `EC_AUTOMATION_APP_ID` to the dev secrets document

lcarva:
> Make sure to add the `EC_AUTOMATION_APP_ID` to the dev secrets document

It's not in a secret because it's not sensitive information. Should it still go there?
(You can read the value of the var from the settings page.)

PR #2038, opened 2024-10-01
Remove `validate definition` command

lcarva:
Ref: EC-896


simonbaird:
Impressive:  `+0 −1,090`

simonbaird:
I was surprised to see we had both `internal/input/report.go` and `internal/definition/report.go` since I thought they were the same thing.

simonbaird (approved):

PR #2047, opened 2024-10-03
Switch to "unversioned" in the antora config

simonbaird:
There will be an impact on the urls. I'll aim to check for broken links later and maybe try to set up a redirect.

See also
https://docs.antora.org/antora/latest/component-with-no-version/

Ref: https://issues.redhat.com/browse/EC-903

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,4 +52,4 @@ $ echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo
 
 
 [pol]: https://github.com/enterprise-contract/ec-policies/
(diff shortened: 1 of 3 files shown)


zregvart:
I haven't found a way to do redirects on GitHub pages, if we want the redirects we'll need to setup something similar to https://github.com/enterprise-contract/enterprise-contract.github.io/pull/327

simonbaird:
> I haven't found a way to do redirects on GitHub pages, if we want the redirects we'll need to setup something similar to [enterprise-contract/enterprise-contract.github.io#327](https://github.com/enterprise-contract/enterprise-contract.github.io/pull/327)

I'm thinking let's not worry about it.

Issue #2025, opened 2024-09-26
Allow violations to downgrade themselves to a warning


lcarva:
It may be desirable for a policy rule to dynamically downgrade itself from a violation to a warning. One of the use cases is to allow certain schema violations, e.g. additionalProperties, to emit a warning instead of a violation.

This behavior is somewhat similar to how `effective_on` works today. In fact, we could achieve the same result by using a date that is always in the future.

**Acceptance Criteria**

* Come up with an annotation that can be emitted by policy rules to trigger the behavior described in this issue, e.g. `downgrade: true`.
* Implement functionality to downgrade such violations to warnings.

zregvart:
Perhaps we could introduce generic rules in addition to `deny` and `warn` rules, they will need to provide all the necessary data in their result, including for example a `severity` field. The balance would need to be struck on what fields must remain static, e.g. `short_name`

Something like:

```
rule contains result if {
    result := {
        code: ...,
        msg: ...,
        effective_on: ...,
        severity: ...
    }
}
```

lcarva:
That will likely require some invasive changes in the evaluator code. Maybe something we can eventually do once EC-533 is done?

lcarva:
Added a comment to EC-533 for us to consider introducing that idea.

PR #2049, opened 2024-10-04
Add support for setting severity dynamically

lcarva:
This commit adds support for a new custom rego annotation, called `severity`. When emitted, this can turn a warning into a violation, and a violation to a warning.

(It also includes a commit about removing an unused parameter of the internal function `computeSuccesses`.)

Fixes: #2025
Ref: EC-894

### DIFF
--- a/acceptance/examples/dynamic_severity.rego
+++ b/acceptance/examples/dynamic_severity.rego
@@ -1,0 +1,31 @@
+package main
+
+import rego.v1
+
+deny contains result if {
+    result := {
+        "msg": "Failure to warning",
+        "severity": "warning"
+    }
+}
+
+deny contains result if {
+    result := {
+        "msg": "Failure to failure",
+        "severity": "failure"
+    }
+}
+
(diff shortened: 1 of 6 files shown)


zregvart (approved):
Not too keen on the term "downgrade", also not sure if this works only one way, e.g. could a warning be changed into a failure?
I'd rather go with something like `override_severity` that can have values `failure` or `warning`.
This should work though...

lcarva:
> Not too keen on the term "downgrade", also not sure if this works only one way, e.g. could a warning be changed into a failure? I'd rather go with something like `override_severity` that can have values `failure` or `warning`. This should work though...

Hmm. This seems like a more flexible approach. Let me play with it.

lcarva:
Updated to use the `severity` metadata annotation. I decided to use this instead of `override_severity` as it's a little bit more succinct but still clearly conveys the message.

simonbaird:
I'm too slow, but nice patch.

PR #2059, opened 2024-10-08
Refactor PolicySource from ECP creation

zregvart:
There is no need to return an error from `FetchPolicySources`, and since `FetchPolicySources` is not fetching the policy sources, only constructing a slice of `PolicySource`s let's call it `PolicySourcesFrom` instead.


zregvart:
Pulled out of #2048

PR #2063, opened 2024-10-09
EC-908 Move GetPinnedURL func to go-gather.

robnester-rh:
This commit removes the GetPinnedURL function from ec-cli as it has been relocated to go-gather. This commit also makes the necessary changes to test files based on modifications of the GetPinnedURL function in go-gather.


zregvart (approved):

simonbaird:
I think it's probably unblocked now since I changed the POLICY_CONFIGURATION param in the ITS.

robnester-rh:
> I think it's probably unblocked now since I changed the POLICY_CONFIGURATION param in the ITS.

@simonbaird, seems to be still be blocked.

simonbaird:
My fault, I mistyped the param.

simonbaird:
:tada:

PR #2073, opened 2024-10-14
Allow `GIT_REPO` to be parameterized

zregvart:
Helpful when overriding the `IMAGE` to a non-golden image.


robnester-rh (commented):
LGTM

PR #2075, opened 2024-10-14
Update to go 1.22.7

lcarva:


lcarva:
oops. Let's follow up on #2074 then.

PR #2087, opened 2024-10-16
Do not create subsequent symlinks to same source

zregvart:
This prevents creation of symlinks pointing to the same source within a workdir by generating a non-random unique placement directory and checking if the placement directory (or symlink) already exists.

Reference: https://issues.redhat.com/browse/EC-936

### DIFF
--- a/internal/policy/source/source.go
+++ b/internal/policy/source/source.go
@@ -30,7 +30,6 @@ import (
 	"path/filepath"
 	"runtime/trace"
 	"sync"
-	"time"
 
 	ecc "github.com/enterprise-contract/enterprise-contract-controller/api/v1alpha1"
 	"github.com/enterprise-contract/go-gather/metadata"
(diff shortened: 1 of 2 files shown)


zregvart:
Created as draft, looking for feedback if the test accurately checks the issue we're seeing

zregvart:
This is the quickest/easiest attempt at fixing the issue, it might be much nicer to do a more thorough refactor but I left that for later.

zregvart:
To reproduce/test this using `hack/simple-demo.sh` comment out [this line](https://github.com/enterprise-contract/ec-cli/blob/b9827f791a2a011e86291f011eb130c4ac417f5b/hack/simple-demo.sh#L63) and uncomment [this line](https://github.com/enterprise-contract/ec-cli/blob/b9827f791a2a011e86291f011eb130c4ac417f5b/hack/simple-demo.sh#L65). And place the attached [`policy.yaml`](https://github.com/user-attachments/files/17412698/policy.yaml.txt) file in $CWD.

robnester-rh (approved):
LGTM

lcarva (approved):

PR #2103, opened 2024-10-24
Fixes the CLI reference docs template

zregvart:
Resolves: https://issues.redhat.com/browse/EC-940


zregvart:
The change is trivial enough, I'm going to merge this without review. Any followup review suggestions can be incorporated in a separate PR

PR #2128, opened 2024-11-01
Resolve ImageManifests concurrently

joejstuart:
This should speed up resolving manifests when there's a greater number of components.

https://issues.redhat.com/browse/EC-975

### DIFF
--- a/internal/applicationsnapshot/input.go
+++ b/internal/applicationsnapshot/input.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"fmt"
 	"runtime/trace"
+	"sort"
+	"sync"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	app "github.com/konflux-ci/application-api/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


internal/applicationsnapshot/input.go
```diff
@@ -189,59 +240,39 @@ func expandImageIndex(ctx context.Context, snap *app.SnapshotSpec) {
...
-			allErrors = errors.Join(allErrors, fmt.Errorf("unable to parse container image %s: %w", component.ContainerImage, err))
-			continue
-		}
 
-		desc, err := client.Head(ref)
-		if err != nil {
-			allErrors = errors.Join(allErrors, fmt.Errorf("unable to fetch descriptior for container image %s: %w", ref, err))
-			continue
-		}
+	componentChan := make(chan app.SnapshotComponent, len(snap.Components))
```
lcarva, line 246: This will be too high. I tried these changes with 59 components and quay started to report `429 Too Many Requests`.

I think this should match the value from the `--workers` parameter, default is 5.

With the 59 component snapshot, without your changes, it takes 2m52.847s to expand all the images. With your changes (but capping at 5 concurrency) this takes `0m8.583s`. That's pretty darn good.

lcarva: Or maybe it should be a different value. I can see how we may want a high number for validating images, but a lower number for expanding them. For example, 50 is fine for validating, but it's too high for quay.io.

Consider maybe just hard-coding to something like 5 for now.

joejstuart: Ah, Good point. I'll update.

joejstuart: Tried this with the `errgroup` package. https://github.com/enterprise-contract/ec-cli/pull/2128/commits/367518b7ce946d3bd6ce79258cc37282714d5511

joejstuart: I want to try and test this and see if the limit works.

zregvart (commented):
Let's make sure all goroutines terminate early if an error occurs in one of them

internal/applicationsnapshot/input.go
```diff
@@ -181,67 +183,98 @@ func readSnapshotSource(input []byte) (app.SnapshotSpec, error) {
 	return file, nil
 }
 
+// For an image index, remove the original component and replace it with an expanded component with all its image manifests
+// Do not raise an error if the image is inaccessible, it will be handled as a violation when evaluated against the policy
+// This is to retain the original behavior of the `ec validate` command.
+func imageIndexWorker(client oci.Client, component app.SnapshotComponent, componentChan chan<- app.SnapshotComponent, errorsChan chan<- error) {
```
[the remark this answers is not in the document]
joejstuart: If an error is never returned, do we need to worry about this? https://github.com/enterprise-contract/ec-cli/pull/2128/files#diff-5598e35222f987336103db6c58e8f37565a1b7a3cc1a9ccbc31cb33bb24624faR253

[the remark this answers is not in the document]
joejstuart: But if I return an error won't the execution stop? This comment makes me think we still need only collect the errors still. https://github.com/enterprise-contract/ec-cli/pull/2128/files#diff-5598e35222f987336103db6c58e8f37565a1b7a3cc1a9ccbc31cb33bb24624faR187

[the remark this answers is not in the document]
zregvart: We want the execution not to proceed once the first error occurred. Say we have 10 components pointing to image indexes with to 10 images we can't access any of them (the `client.Head` will fail for all of them). We don't want to resolve and fail for each one of those, we want to resolve at most the number of concurrent goroutines (currently 5) times, and skip resolving in any unscheduled work (skip the 95 remaining).

Have a look at the [example](https://github.com/zregvart/er/blob/main/main.go) I created, if run without arguments, 10 goroutines are executed:

```
$go run github.com/zregvart/er@latest        
fail at: 3, handle false
running: 4
running: 0
running: 1
running: 2
running: 3
running: 5
running: 6
running: 7
running: 8
running: 9
Done: fail at: 3
```

If executing with this error handling only 5 will execute:
```
$ go run github.com/zregvart/er@latest -handle
fail at: 3, handle true
running: 4
running: 2
running: 0
running: 1
running: 3
Done: fail at: 3
```

zregvart (approved):
:rocket:

PR #2138, opened 2024-11-05
Add Konflux release Pipelines

lcarva:
This commit adds two Tekton Pipelines. One is meant to be used to release the CLI image via Konflux, the other for releasing the Tekton Task bundle.

Ideally, a single Pipeline would be used. However, due to limitations in Konflux release, this would force us to use the same EC policy config for both images which is currently not supported by EC.

The Pipelines are based on the push-to-external-registry Pipeline provided by Konflux release (link in code). Some of the Tasks require changes for a couple of reasons:

1. Some Tasks require ReleasePlanAdmission and ReleaseConfig resource references. Neither one of those are used/available during tenant releases.

2. Some Tasks assume the image is an OCI image (good thing). However, the `tekton bundle push` command creates an Image Manifest with the docker media type (kind of deprecated).

To make this saner to maintain, `kustomize` is used to compose the release Pipelines. `make generate-pipelines` is added to facilitate running the command.

Additionally, because two EC policies are used, the `.ec` directory didn't make sense anymore. As such, a new top-level `policies` directory is used.

Ref: EC-910

### DIFF
--- a/policies/cli/policy.yaml
+++ b/policies/cli/policy.yaml
@@ -1,3 +1,19 @@
+# Copyright The Enterprise Contract Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
 #
 # If we need to diverge from the "official" policy for pre-merge EC checks,
 # the EC integration test scenario POLICY_CONFIGURATION param can be set to
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,11 @@ build-for-test: dist/ec_$(BUILD_IMG_ARCH)
 clean: ## Delete build output
 	@rm -f dist/*
 
+.PHONY: generate-pipelines
+generate-pipelines: ## Generate release pipelines
+	kustomize build ./release/src/cli --output ./release/cli.yaml
+	kustomize build ./release/src/tekton-task --output ./release/tekton-task.yaml
+
 ##@ Testing
 
 # Declutter the output by grepping out the files where there are no
--- a/policies/tekton-task/policy.yaml
+++ b/policies/tekton-task/policy.yaml
@@ -1,0 +1,32 @@
+# Copyright The Enterprise Contract Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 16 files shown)


lcarva:
/retest

"Couldn't get task"

lcarva:
Making it a draft because I found some updates that address the need of some of the Task customization. Trying it out.

lcarva:
Changed the approach of how to use the modified Tasks.

1. `apply-mapping`: Use a version from the development branch that already contains the fix.
2. `push-snapshot`: Use proposed changes: https://github.com/konflux-ci/release-service-catalog/pull/671
3. `collect-data` and `verify-access-to-resources`: Use proposed changes: https://github.com/konflux-ci/release-service-catalog/pull/672

Once those changes are merged and make their way into the `production` branch, we can update the kustomize bits here to just use that. However, this doesn't have to happen for _this_ PR to be merged. I have tested that the release pipelines work as expected as is.

Moving it out of draft. PTAL! :pray:

policies/cli/policy.yaml
```diff
@@ -1,3 +1,19 @@
+# Copyright The Enterprise Contract Contributors
```
zregvart, line 1: Nitpick, perhaps `.ec/cli-policy.yml`? Reasoning that `.ec` might be a directory use as a convention?

release/push-to-registry-cli.yaml
```diff
@@ -0,0 +1,218 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: tekton.dev/v1
+kind: Pipeline
+metadata:
+  name: push-to-registry-cli
```
zregvart, line 20: Nitpick:
```suggestion
  name: release-cli
```

Thinking we might want to customize this further and perform not just the push to registry

release/push-to-registry-cli.yaml
```diff
@@ -0,0 +1,218 @@
...
+    - name: releasePlan
+      value: $(params.releasePlan)
+    - name: snapshot
+      value: $(params.snapshot)
+    - name: requireInternalServices
+      value: "false"
+    taskRef:
+      params:
+      - name: url
+        value: https://github.com/lcarva/release-service-bundles.git
```
zregvart, line 81: Perhaps add a comment and point to https://github.com/konflux-ci/release-service-catalog/pull/672?

zregvart (approved):
Lots of YAML, so for me it is difficult to review this. LGTM and let's revise/improve as needed in followups.

policies/cli/policy.yaml
lcarva: That doesn't work because go-gather expects it to be a directory when using git gather:

```
Error: error copying directory: /tmp/git-repo-3559490554/.ec/cli-policy.yaml is not a directory
```

And the file has to be named `policy.yaml` as well. We should make this better but I'd rather not do that as part of this PR.

I could rename `policies` to `.ec`?

release/push-to-registry-cli.yaml
lcarva: Done. Also renamed files accordingly.

release/push-to-registry-cli.yaml
lcarva: This yaml is generated by Kustomize so no comments here, but there are comments in the files under `release/src/base/tasks`.

lcarva:
I'm gonna go ahead and merge this so we can carry on. Additional, post-merge, reviews are welcome.

Issue #2107, opened 2024-10-24
[BUG] Investigate and fix the acceptance test failure


zregvart:
Acceptance tests failed in [this run](https://github.com/enterprise-contract/ec-cli/actions/runs/11481710447/job/31953041821#step:6:2387). Investigate and fix the issue. This could possibly be related to #2087

PR #2150, opened 2024-11-08
Use given source URLs as keys in download cache

zregvart:
We ran into an issue when we started mixing given and pinned source URLs in the download cache. This has happened because we cached the given URLs first, mutated the policy to contain the pinned URLs and later tried to download through the cache via pinned URLs.

This changes so we only use the given source URLs in the download cache by allowing the source URLs to mutate the policy URLs after they are downloaded. That is, for each source URL in the policy we have two states: one containing the given value and used in the download cache, and second after the download with the resulting resolved/pinned value.

There are two repercussions of this: any resolving/pinning of URLs is transparent to the callers if they use `source.PolicySourcesFrom` followed by `url.GetPolicy` -- this allows any use of `Policy` to be automatically resolved/pinned (no need for Policy pre-processing); and secondly the Policy is mutated while in execution via `url.GetPolicy` and careful consideration, hopefully/effectively shielded by using `source.PolicySourcesFrom`, of the slice mutation -- making sure that the mutation doesn't happen in the temporary value of the iteration is required.

Resolves: https://issues.redhat.com/browse/EC-963
Fixes: #2107

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -230,63 +230,63 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 				RekorURL:    data.rekorURL,
 			}
 
-			// We're not currently using the policyCache returned from PreProcessPolicy, but we could
-			// use it to cache the policy for future use.
-			if p, _, err := policy.PreProcessPolicy(ctx, policyOptions); err != nil {
+			p, err := policy.NewPolicy(ctx, policyOptions)
+			if err != nil {
 				allErrors = errors.Join(allErrors, err)
-			} else {
-				// inject extra variables into rule data per source
-				if len(data.extraRuleData) > 0 {
-					policySpec := p.Spec()
-					sources := policySpec.Sources
-					for i := range sources {
-						src := sources[i]
-						var rule_data_raw []byte
-						unmarshaled := make(map[string]interface{})
-
-						if src.RuleData != nil {
-							rule_data_raw, err = src.RuleData.MarshalJSON()
-							if err != nil {
-								log.Errorf("Unable to parse ruledata to raw data")
-							}
-							err = json.Unmarshal(rule_data_raw, &unmarshaled)
-							if err != nil {
-								log.Errorf("Unable to parse ruledata into standard JSON object")
-							}
-						} else {
-							sources[i].RuleData = new(extv1.JSON)
-						}
-
-						for j := range data.extraRuleData {
-							parts := strings.SplitN(data.extraRuleData[j], "=", 2)
-							if len(parts) < 2 {
-								log.Errorf("Incorrect syntax for --extra-rule-data")
-							}
-							extraRuleDataPolicyConfig, err := validate_utils.GetPolicyConfig(ctx, parts[1])
-							if err != nil {
-								log.Errorf("Unable to load data from extraRuleData: %s", err.Error())
-							}
-							unmarshaled[parts[0]] = extraRuleDataPolicyConfig
-						}
-						rule_data_raw, err = json.Marshal(unmarshaled)
-						if err != nil {
-							log.Errorf("Unable to parse updated ruledata: %s", err.Error())
-						}
-
-						if rule_data_raw == nil {
-							log.Errorf("Invalid rule data JSON")
-						}
-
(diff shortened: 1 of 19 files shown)


zregvart:
I think I'd like to add more tests for this, not sure if I want to do this in this PR (which is already quite large) or in a followup.

zregvart:
To reproduce the issue reliably add `time.Sleep(2 * time.Seconds)` [here](https://github.com/enterprise-contract/ec-cli/pull/2150/files#diff-afd15fb52ad3aad95f7208a6db2f2fcf6010da9d8fa152b9e60079cd1cf0fe3bL90) (in `source.getPolicyThroughCache`)

features/__snapshots__/validate_input.snap
```diff
@@ -15,7 +15,7 @@
     "sources": [
       {
         "policy": [
-          "git::https://${GITHOST}/git/happy-day-policy.git"
+          "git::${GITHOST}/git/happy-day-policy.git?ref=${LATEST_COMMIT}"
```
zregvart, line 18: `validate input` now supports pinned source URLs

cmd/validate/image.go
```diff
@@ -230,63 +230,63 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
...
+				sources := policySpec.Sources
+				for i := range sources {
+					src := sources[i]
+					var rule_data_raw []byte
+					unmarshaled := make(map[string]interface{})
+
+					if src.RuleData != nil {
+						rule_data_raw, err = src.RuleData.MarshalJSON()
+						if err != nil {
+							log.Errorf("Unable to parse ruledata to raw data")
```
lcarva, line 251: I think we need to append the err to `allErrors` and and call `continue`? Same for the other statements below.

features/validate_image.feature
```diff
@@ -1137,9 +1137,8 @@ Feature: evaluate enterprise contract
       | [{"op": "add", "path": "/sources/8/ruleData", "value": {"mr": "mxyzptlk"}}]    |
       | [{"op": "add", "path": "/sources/9/ruleData", "value": {"more": "data"}}]      |
       And an Snapshot named "multitude" with 10 components signed with "known" key
-    When ec command is run with "validate image --snapshot acceptance/multitude --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --rekor-url ${REKOR} --show-successes --output json"
+    When ec command is run with "validate image --snapshot acceptance/multitude --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --rekor-url ${REKOR} --show-successes --output json --trace"
```
lcarva, line 1140: `--trace` intentional?

features/validate_image.feature
```diff
@@ -1137,9 +1137,8 @@ Feature: evaluate enterprise contract
...
     Then the exit status should be 0
-     And the output should match the snapshot
```
lcarva, line 1142: Intentional?

lcarva:
I really would prefer if we did have a pattern where something transforms an existing policy into a new policy with the pinned references. (I don't think we got that implementation quite right.) I'm hesitant of adding more pointer manipulation in the code base because that is harder to debug and maintain. But this is all internal and we should be able to refactor as needed later on.

cmd/validate/image.go
zregvart: This wasn't there before, not sure if I want to open that can of worms now

zregvart:
> I really would prefer if we did have a pattern where something transforms an existing policy into a new policy with the pinned references. (I don't think we got that implementation quite right.) I'm hesitant of adding more pointer manipulation in the code base because that is harder to debug and maintain. But this is all internal and we should be able to refactor as needed later on.

Me too, would be a much bigger refactor though. The way we have it working now is that the download of sources is initiated from the evaluator, and the previous attempt of preprocessing the policy led us to a myriad of issues. Not saying it can't be done.

cmd/validate/image.go
lcarva: Fair point. I'll file an issue.

lcarva: https://issues.redhat.com/browse/EC-1020

PR #2165, opened 2024-11-15
Don't store data field - POC

joejstuart:
POC of writing results to a file then streaming them to stdout. This only writes the `data` field from the `results` struct.

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"runtime"
 	"runtime/trace"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/validate/image.go
```diff
@@ -400,21 +403,58 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
...
+				tempFile, err := os.CreateTemp("", "batch_*.gob")
+				if err != nil {
+					return fmt.Errorf("failed to create temp file: %w", err)
+				}
+				defer tempFile.Close()
+				tempFiles = append(tempFiles, tempFile.Name())
+
+				encoder := gob.NewEncoder(tempFile)
+				if err := encoder.Encode(manyData); err != nil {
+					fmt.Printf("failed to encode batch: %w", err)
```
robnester-rh, line 447: This should be `fmt.Errorf` if you're doing error wrapping.

PR #2174, opened 2024-11-18
Validate bundle schema

joejstuart:
Run schema validation when creating a new tracker.

https://issues.redhat.com/browse/EC-1016

### DIFF
--- a/internal/tracker/tracker.go
+++ b/internal/tracker/tracker.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/santhosh-tekuri/jsonschema/v5"
 	log "github.com/sirupsen/logrus"
 	"github.com/stuart-warren/yamlfmt"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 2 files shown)


internal/tracker/tracker.go
```diff
@@ -33,6 +34,29 @@ import (
 
 const ociPrefix = "oci://"
 
+const jsonSchema = `{
```
zregvart, line 37: Could use https://pkg.go.dev/embed here

internal/tracker/tracker.go
```diff
@@ -33,6 +34,29 @@ import (
...
+                "type": "object",
+                "properties": {
+                    "effective_on": {"type": "string"},
+                    "expires_on": {"type": "string"},
+                    "ref": {"type": "string"}
+                },
+                "required": ["effective_on", "ref"],
+                "additionalProperties": false
+            },
+            "uniqueItems": true,
```
zregvart, line 53: Not sure what the probability of encountering duplicate value is, if it is near certain that it will happen then we'll hit this and prevent new Tasks from being pushed to quay.

internal/tracker/tracker.go
zregvart, line 37: Should we validate the schema over the whole file?

internal/tracker/tracker.go
```diff
@@ -63,6 +91,28 @@ func newTracker(input []byte) (t Tracker, err error) {
...
+	if err := yaml.Unmarshal(bundleConfig, &t); err != nil {
+		return err
+	}
+	val, ok := t["trusted_tasks"]
+	if !ok {
+		return fmt.Errorf("the data does not contain a trusted_tasks key")
+	}
+
+	if err = sch.Validate(val); err != nil {
+		return err
```
zregvart, line 110: I'm wondering, in lieu of https://github.com/enterprise-contract/ec-policies/pull/1225 if we should warn that there are duplicates?

PR #2177, opened 2024-11-19
Remove dead code

lcarva:
These functions and variables are not used anywhere. Let's remove them to avoid collecting cruft.


zregvart (approved):
Thanks! Linter also suggests removing more dead code

PR #2180, opened 2024-11-20
Filter out duplicate trusted task records

zregvart:
This filters out all duplicate trusted task records based on their reference. Previously we would de-duplicate only if the records were adjacent.

Reference: https://issues.redhat.com/browse/EC-1015

### DIFF
--- a/internal/tracker/tracker.go
+++ b/internal/tracker/tracker.go
@@ -269,10 +269,12 @@ func (t *Tracker) filterBundles(prune bool) {
 func filterRecords(records []taskRecord, prune bool) []taskRecord {
 	now := time.Now().UTC()
 
-	// lastRef tracks the latest ref seen. This is used to remove consecutive entries with the
-	// same digest.
-	var lastRef string
+	// seenRefs tracks the refs that have been encountered. This is used to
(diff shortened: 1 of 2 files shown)


lcarva:
> Previously we would de-duplicate only if the records were adjacent.

This was very much intentional: https://issues.redhat.com/browse/HACBS-2394

zregvart:
It seems that the schema check in ec-policies needs to change to allow duplicate entries then. 

I'm not sure that build-definitions can produce equal digests for two builds, the git references should be different and the images should have different digests as they differ in annotations and in the config/created.

lcarva:
> It seems that the schema check in ec-policies needs to change to allow duplicate entries then.
:+1: 

> I'm not sure that build-definitions can produce equal digests for two builds, the git references should be different and the images should have different digests as they differ in annotations and in the config/created.

I don't remember exactly the scenario in which this may occur, but it may have something to do with the build scripts taking shortcuts to not rebuild an image bundle.

zregvart:
> I don't remember exactly the scenario in which this may occur, but it may have something to do with the build scripts taking shortcuts to not rebuild an image bundle.

Let's not find out, shall we...

PR #2181, opened 2024-11-20
Track the policy data by source group

joejstuart:
Only collect the policy data once for each source group. Before this the data was duplicated for each component.

https://issues.redhat.com/browse/EC-1027

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -301,7 +301,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			type result struct {
 				err         error
 				component   applicationsnapshot.Component
-				data        []evaluator.Data
+				data        map[string]evaluator.Data
 				policyInput []byte
 			}
 
@@ -343,6 +343,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 					}
 
 					log.Debugf("Worker %d got a component %q", id, comp.ContainerImage)
+
 					out, err := validate(ctx, comp, data.spec, data.policy, evaluators, data.info)
 					res := result{
 						err: err,
@@ -366,8 +367,8 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 						res.component.Signatures = out.Signatures
 						res.component.Attestations = out.Attestations
 						res.component.ContainerImage = out.ImageURL
-						res.data = out.Data
 						res.component.Attestations = out.Attestations
+						res.data = out.Data
 						res.policyInput = out.PolicyInput
 					}
 					res.component.Success = err == nil && len(res.component.Violations) == 0
@@ -400,7 +401,9 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			close(jobs)
 
 			var components []applicationsnapshot.Component
-			var manyData [][]evaluator.Data
+			var manyData []evaluator.Data
+			// we don't want to accumulate the data from each source group for each component
+			sgData := make(map[string]evaluator.Data)
 			var manyPolicyInput [][]byte
 			var allErrors error = nil
 			for i := 0; i < numComponents; i++ {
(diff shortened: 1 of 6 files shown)


internal/image/validate.go
```diff
@@ -130,7 +131,7 @@ func ValidateImage(ctx context.Context, comp app.SnapshotComponent, snap *app.Sn
 			return nil, err
 		}
 		allResults = append(allResults, results...)
-		out.Data = append(out.Data, data)
+		out.Data[fmt.Sprintf("%d", idx)] = append(out.Data[fmt.Sprintf("%d", idx)], data)
```
zregvart, line 134: Could be a tad nicer
```suggestion
		key := strconv.FormatInt(idx, 10)
		out.Data[key] = append(out.Data[key], data)
```

zregvart (commented):
I'm expecting snapshots that need updating (haven't looked at the CI), so `UPDATE_SNAPS=1 make ci` should help

zregvart:
The snapshot variable substitution is somewhat rudimentary, I think we either need to indent by 2 or decrease the indent by 2, and the underscores should probably reflect that [here](https://github.com/enterprise-contract/ec-cli/blob/e614d885d00fcdbe44b21d9631c4ec279172a024/acceptance/cli/cli.go#L226)

cmd/validate/image.go
```diff
@@ -409,12 +412,17 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 					e := fmt.Errorf("error validating image %s of component %s: %w", r.component.ContainerImage, r.component.Name, r.err)
 					allErrors = errors.Join(allErrors, e)
 				} else {
+					for key, val := range r.data {
```
lcarva, line 415: I don't think this is quite right. This is merging all the data from different source groups.

Consider something like this:

```go
// Renamed from manyData
var evaluatorData []evaluator.Data
dataCollected := false

for i := 0; i < numComponents; i++ {
  r := <-results
  ...
  if !dataCollected {
    evaluatorData = r.data
    dataCollected = true
  }
}
```

lcarva: or rather:

```go
// Renamed from manyData
var evaluatorData []evaluator.Data

for i := 0; i < numComponents; i++ {
  r := <-results
  ...
  if evaluatorData == nil {
    evaluatorData = r.data
  }
}
```

Anyways, I think you get the idea.

joejstuart: Don't we want that since the data could differ per source group?

lcarva: `r.data` is a list of data.

joejstuart: I changed it to `map[string]evaluator.Data`. My thought was each evaluator maps to a source group, so to just collect data per evaluator.
https://github.com/enterprise-contract/ec-cli/pull/2181/files#diff-7f11215ba891ff8199fa09a412849a5414e416050d163c813f6499009bbe0b96R304
https://github.com/enterprise-contract/ec-cli/pull/2181/files#diff-ed62eb280d370e54af87c7fdb7026890b2efeefd1da359a2b28ecb76904cd86eR136

joejstuart: This makes me think this whole PR could be replaced with 
```diff
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -410,7 +410,9 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
                                        allErrors = errors.Join(allErrors, e)
                                } else {
                                        components = append(components, r.component)
-                                       manyData = append(manyData, r.data)
+                                       if len(manyData) == 0 {
+                                               manyData = append(manyData, r.data)
+                                       }
                                        manyPolicyInput = append(manyPolicyInput, r.policyInput)
                                }
                        }


```

joejstuart:
Closing in favor of this https://github.com/enterprise-contract/ec-cli/pull/2184

PR #2184, opened 2024-11-25
Only store policy data once

joejstuart:
Only collect the policy data once for each source group. Before this the data was duplicated for each component.

https://issues.redhat.com/browse/EC-1027

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -400,7 +400,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			close(jobs)
 
 			var components []applicationsnapshot.Component
-			var manyData [][]evaluator.Data
+			var evaluatorData [][]evaluator.Data
 			var manyPolicyInput [][]byte
 			var allErrors error = nil
 			for i := 0; i < numComponents; i++ {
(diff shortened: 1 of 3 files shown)


cmd/validate/image.go
```diff
@@ -400,7 +400,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			close(jobs)
 
 			var components []applicationsnapshot.Component
-			var manyData [][]evaluator.Data
+			var evaluatorData [][]evaluator.Data
```
lcarva, line 403: Should this be `[]evaludator.Data` instead?

joejstuart: Intuitively, I think so, but I don't know if it adds any value and there's an acceptance test for it even, so I'd prefer to leave it alone.

lcarva (approved):

zregvart (approved):
One followup could be that we only collect the data if `--output=data` or the like was provided

joejstuart:
> One followup could be that we only collect the data if `--output=data` or the like was provided

I think this will do that. https://github.com/enterprise-contract/ec-cli/pull/2184/commits/5464dbfd7a3cf8808e64e49c23badd390b2ff2bb

PR #2197, opened 2024-12-04
Remove deprecated sbom-json-check Task

simonbaird:
The task started failing. Since it's deprecated, rather than figure out why, let's remove it.

See also equivalent main branch commit in sha
a9ec0992d65287e7c13725b03933005685ed03bc7 .

### DIFF
--- a/.tekton/cli-v04-pull-request.yaml
+++ b/.tekton/cli-v04-pull-request.yaml
@@ -379,28 +379,6 @@ spec:
         operator: in
         values:
         - "false"
-    - name: sbom-json-check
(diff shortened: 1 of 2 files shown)


lcarva (approved):

lcarva:
oops. clicked on "merge when ready" which merged it right away because there are no required checks in the target branch :cry:

simonbaird:
> oops. clicked on "merge when ready" which merged it right away because there are no required checks in the target branch 😢

Konflux build and EC check are green (mostly), so should be fine.

PR #2198, opened 2024-12-04
Bump ubi-minimal to 9.5 & update konflux refs

simonbaird:
The motivation is to get a newer version of lib-krb5, specifically 1.21.1-4.el9_5 to fix a vulnerability detected by clair scan.

This is the same base image currently used in main branch, see commit 234d9f40bb87465ada974a098033706466be26b7 .

### DIFF
--- a/.tekton/cli-build.yaml
+++ b/.tekton/cli-build.yaml
@@ -13,7 +13,7 @@ spec:
           - name: name
             value: show-sbom
           - name: bundle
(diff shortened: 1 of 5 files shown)


lcarva:
I think `[rpms.lock.yaml](https://github.com/enterprise-contract/ec-cli/blob/release-v0.5/rpms.lock.yaml)` will also need an update.

simonbaird:
This is failing on trusted task errors, maybe I need to combine it with #2097.

simonbaird:
Not much to review here other than digest changes and package version bumps. Let's merge!

PR #2207, opened 2024-12-08
Only store attestation data that is needed

joejstuart:
The whole attestation is being stored for each component that is evaluated, but in most cases only a subset of the attestation is printed. This change captures only the needed attestation data based on the output selected at runtime.

https://issues.redhat.com/browse/EC-1026

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -364,10 +364,17 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 						}
 
 						res.component.Signatures = out.Signatures
-						res.component.Attestations = out.Attestations
+						// Create a new result object for attestations. The point is to only keep the data that's needed.
+						// For example, the Statement is only needed when the full attestation is printed.
+						for _, att := range out.Attestations {
+							attResult := applicationsnapshot.NewAttestationResult(att)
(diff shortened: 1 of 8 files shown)


simonbaird (approved):
It seems reasonable. Can you see a significant reduction in memory use for a large snapshot?

lcarva:
I'm concerned that this doesn't go far enough. It's still accumulating large amounts of data in memory which we know it's a problem. What about writing the data to files and just track the file paths in memory?

Anyways, I think measuring the impact in memory utilization before and after this change should be done to help decide which path to take.

joejstuart:
| Metric      | After change                  | Before change                |
|-------------|--------------------------------|--------------------------------|
| User time   | 999.78u (999.78 seconds)       | 916.93u (916.93 seconds)       |
| System time | 31.84s (31.84 seconds)         | 30.46s (30.46 seconds)         |
| Real time   | 1232.18r (1232.18 seconds)     | 1342.26r (1342.26 seconds)     |
| Memory      | 777280kB (~759 MB)             | 1147664kB (~1120 MB)           |

simonbaird:
The memory reduction looks good. Do you have a bash snippet that you used to produce those?

PR #2217, opened 2024-12-12
Script for timing large snapshot

simonbaird:
Not sure if we wanna merge this, but I'll push a draft PR to share my hacky script.

Somewhat on-topic for...

Ref: https://issues.redhat.com/browse/EC-976


simonbaird:
... rebased because I want to trigger some Konflux pipelines to confirm we didn't break production with https://github.com/redhat-appstudio/infra-deployments/pull/6225 ...

PR #2218, opened 2024-12-12
Add workers flag to `validate input` command

robnester-rh:
This commit reworks the concurrency model of `validate input` to be like that found in `validate image`. This commit adds a `--workers` flag, with a default of 5 that can be used to configure how many simultaneous fetches occur. This avoids situations where validating multiple input files can trigger HTTP 429 errors.


zregvart (approved):
Thanks for adding tests. :+1: LGTM

PR #2220, opened 2024-12-13
Enable the fetching of only the OCI descriptor

arewm:
Some OCI artifacts need to have a specific media type (i.e. they should only be an OCI image index or an OCI image manifest). In order to determine that with a single remote call, we can just get the descriptor and then check the mediaType of that object.


zregvart (approved):

internal/rego/oci/oci.go
```diff
@@ -73,6 +74,43 @@ func registerOCIBlob() {
 	})
 }
 
+func registerOCIDescriptor() {
+	manifest := types.NewObject(
+		[]*types.StaticProperty{
+			// Specifying the properties like this ensure the compiler catches typos when
+			// evaluating rego functions.
+			{Key: "mediaType", Value: types.S},
```
zregvart, line 82: Either we add all properties from the [Descriptor](https://pkg.go.dev/github.com/google/go-containerregistry@v0.20.2/pkg/v1#Descriptor) here or we call this function `ec.oci.descriptor_media_type`

zregvart (commented):
How do you feel about adding other properties to the returned object?

internal/rego/oci/oci.go
arewm: I missed that here. I was setting them within the `ociDescriptor` function.

zregvart (approved):

zregvart:
Do a `make generate` and squash commits so they're nicer a bit?

arewm:
Thanks. Squashed and regenerated the docs.

zregvart (approved):

PR #2229, opened 2024-12-19
Support running with no timeout, ignore/deprecate TIMEOUT task param

simonbaird:
See commits for details.

Ref: https://issues.redhat.com/browse/EC-1030

### DIFF
--- a/cmd/root/root_cmd.go
+++ b/cmd/root/root_cmd.go
@@ -79,7 +79,7 @@ func NewRootCmd() *cobra.Command {
 			var cancel context.CancelFunc
 			if globalTimeout > 0 {
 				ctx, cancel = context.WithTimeout(ctx, globalTimeout)
-				log.Debugf("globalTimeout is %d", globalTimeout)
+				log.Debugf("globalTimeout is %s", time.Duration(globalTimeout))
 			} else {
 				log.Debugf("globalTimeout is %d, no timeout used", globalTimeout)
 			}
--- a/cmd/root/root_cmd.go
+++ b/cmd/root/root_cmd.go
@@ -73,11 +73,18 @@ func NewRootCmd() *cobra.Command {
 			// set a custom message for context.DeadlineExceeded error
 			context.DeadlineExceeded = customDeadlineExceededError{}
 
-			// Create a new context now that flags have been parsed so a custom timeout can be used.
-			ctx, cancel := context.WithTimeout(cmd.Context(), globalTimeout)
+			// Create a new context now that flags have been parsed so a
+			// custom timeout can be used and traces can be added
+			ctx := cmd.Context()
+			var cancel context.CancelFunc
+			if globalTimeout > 0 {
+				ctx, cancel = context.WithTimeout(ctx, globalTimeout)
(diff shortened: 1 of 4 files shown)


cmd/root/root_cmd.go
```diff
@@ -73,11 +73,18 @@ func NewRootCmd() *cobra.Command {
...
 
-			// Create a new context now that flags have been parsed so a custom timeout can be used.
-			ctx, cancel := context.WithTimeout(cmd.Context(), globalTimeout)
+			// Create a new context now that flags have been parsed so a
+			// custom timeout can be used and traces can be added
+			ctx := cmd.Context()
+			var cancel context.CancelFunc
+			if globalTimeout > 0 {
+				ctx, cancel = context.WithTimeout(ctx, globalTimeout)
+				log.Debugf("globalTimeout is %d seconds", globalTimeout/time.Second)
```
zregvart, line 82: A bit nicer way would be this:
```suggestion
				log.Debugf("globalTimeout is %s seconds", time.Duration(globalTimeout))
```

zregvart (approved):

cmd/root/root_cmd.go
simonbaird: Nice, thanks.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -224,7 +226,8 @@ spec:
         - "$(params.WORKERS)"
         # NOTE: The syntax below is required to negate boolean parameters
         - "--info=$(params.INFO)"
-        - "--timeout=$(params.TIMEOUT)"
+        # No timeout for EC, though Tekton may apply a timeout
+        - "--timeout=0"
```
simonbaird, line 230: This works quite badly on any version of EC that doesn't have this PR merged. It's probably fine  (???), but I was considering using `--timeout=100h` for a while just to be extra cautious/paranoid.

zregvart: We can't release this version of the task without the new version of EC

simonbaird: Agreed. I was wondering about the roll-out process. See [this slack thread](https://redhat-internal.slack.com/archives/C031J4KBFME/p1734637683259059) where @lcarva shared a doc that describes the details.

simonbaird: Decided to be cautious. See new top commit.

lcarva (approved):

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
lcarva: The official Task bundle is always built to use the latest EC CLI at the time (via OCI tag to a specific git commit ID). Also, updates to Konflux are always driven from Task updates. In other words, new CLI images are used indirectly via a new Task bundle. However, I agree with the general instance of being cautious with the change.

PR #2252, opened 2025-01-08
Allow choosing OPA evaluator

joejstuart:
This just creates the option to use OPA by setting the EC_USE_OPA env variable.

https://issues.redhat.com/browse/EC-1044

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -45,6 +45,7 @@ import (
 type imageValidationFunc func(context.Context, app.SnapshotComponent, *app.SnapshotSpec, policy.Policy, []evaluator.Evaluator, bool) (*output.Output, error)
 
 var newConftestEvaluator = evaluator.NewConftestEvaluator
+var newOPAEvaluator = evaluator.NewOPAEvaluator
 
 func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 	data := struct {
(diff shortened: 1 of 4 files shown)


cmd/validate/image.go
```diff
@@ -45,6 +46,7 @@ import (
 type imageValidationFunc func(context.Context, app.SnapshotComponent, *app.SnapshotSpec, policy.Policy, []evaluator.Evaluator, bool) (*output.Output, error)
 
 var newConftestEvaluator = evaluator.NewConftestEvaluator
+var newOPAEvaluator = evaluator.NewOPAEvaluator
```
simonbaird, line 49: Would it be more efficient to not instantiate both of them?

simonbaird: Oh, never mind. I see that we're not calling `evaluator.NewConftestEvaluator`, we're just referencing it here. Disregard the above.

cmd/validate/image.go
```diff
@@ -318,7 +320,14 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 					log.Debugf("policySource: %#v", policySource)
 				}
 
-				c, err := newConftestEvaluator(cmd.Context(), policySources, data.policy, sourceGroup)
+				var c evaluator.Evaluator
+				var err error
+				if os.Getenv("EC_USE_OPA") != "" {
```
simonbaird, line 325: Maybe this:
```suggestion
				if os.Getenv("EC_USE_OPA") != "1" {
```

Since it I might do this `export EC_USE_OPA=0` or `EC_USE_OPA=no` and expect it to dwim.

simonbaird (approved):

cmd/validate/image.go
joejstuart: I used "true" 
```golang
if os.Getenv("EC_USE_OPA") != "true" {
```

cmd/validate/image.go
joejstuart: yeah, I think using the vars made it easier to test.

cmd/validate/image.go
joejstuart: Good idea. Hadn't noticed that.

PR #2263, opened 2025-01-14
Rollback update to softprops/action-gh-release

lcarva:
We're seeing a similar issue to
https://github.com/enterprise-contract/ec-cli/commit/2378fc68fcc0880c267c0d280bb33ac9a6941b2a during a release.

Let's roll back the version to unblock the release.


lcarva:
Merging this to unblock the release.

PR #2265, opened 2025-01-15
Update to newer ubi-minimal base image (v0.5)

simonbaird:


simonbaird:
This is actually newer than the base image we're using in main branch currently, which seems unusual, but I couldn't think of a reason not to use the latest.

PR #2270, opened 2025-01-17
Update CLI release policy

lcarva:
This commit updates the EC policy config for the CLI image in two ways.

First, the policy config now enforces hermetic builds since the builds are indeed hermetic 🎉.

Second the policy config no longer enforces the schedule restriction. Those policy rules are usually meant to avoid releases by a vendor on restricted dates, e.g. weekends. Since this is being used for a community release, this is not a concern.

Ref: EC-911

### DIFF
--- a/policies/cli/policy.yaml
+++ b/policies/cli/policy.yaml
@@ -36,8 +36,8 @@ sources:
       include:
         - '@redhat'
       exclude:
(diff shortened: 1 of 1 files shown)


robnester-rh (approved):
LGTM

simonbaird (approved):

PR #2274, opened 2025-01-17
Upgrade opa deps to 1.0 and get tests passing

simonbaird:
We probably want to also update conftest, so this is a draft for now. I'm pushing it to see if the acceptance tests will pass.

One idea: The "Use v1 compatible `if` syntax.." commit can be split out and merged early I think.


simonbaird:
I split some of this into #2281 which could be merged early.

simonbaird:
Re conftest and opa 1.0 support, I found https://github.com/open-policy-agent/conftest/pull/1032 .

simonbaird:
This could be rejuvenated by rebasing on #2281, but at this point it's probably easier to start again when the time comes.

PR #2281, opened 2025-01-21
Use consistent rego syntax in prep for opa 1.0

simonbaird:
The idea here is that once we update to opa 1.0 we'll need to always use the new `deny contains foo if` syntax. This change is supposed to make it easier by using that syntax consistently now.

It requires importing `rego.v1` everywhere, but we do that consistently for all the rego in ec-policies, so let's also do it here also.

See also #2274 which I'd like to rebase on this once it's merged.

Ref: https://issues.redhat.com/browse/EC-1130

### DIFF
--- a/acceptance/examples/allow_all.rego
+++ b/acceptance/examples/allow_all.rego
@@ -1,4 +1,6 @@
 # Simplest never-failing policy
 package main
 
+import rego.v1
+
 allow := []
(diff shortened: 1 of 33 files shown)


simonbaird:
For some reason I can't run acceptance tests locally, so it's slow going to get all the small snapshot changes done.

simonbaird:
Note the acceptance tests are not actually passing, despite what it looks like.

simonbaird:
Fixed my local acceptance tests by adding `127.0.0.1 apiserver.localhost` and `127.0.0.1 rekor.localhost` to `/etc/hosts`. Now I think this should be going green and just about ready to merge.

robnester-rh (approved):
LGTM
(that was a lot, but makes sense)

PR #2282, opened 2025-01-23
Trigger docs updates on new website

simonbaird:
Originally created the PR with both triggers, but now the old repo is gone, so it's time to remove the old trigger.


simonbaird:
Similar to https://github.com/enterprise-contract/user-guide/pull/161/files

PR #2294, opened 2025-01-30
testing glog update

joejstuart:


simonbaird:
As per slack discussion, the clair is finding git-go vulnerability, upgrading that one should help get EC check green.

PR #2300, opened 2025-02-05
EC-960 Remove last remnants of go-getter

robnester-rh:
This PR removes any direct dependencies on go-getter and replaces them with go-gather as appropriate. Note that go-getter does remain as an indirect dependency.

### DIFF
--- a/internal/policy/source/git_config_test.go
+++ b/internal/policy/source/git_config_test.go
@@ -30,7 +30,6 @@ func TestSourceIsFile(t *testing.T) {
 		want bool
 	}{
 		{src: "", want: false},
-		{src: "foo", want: true},
 		{src: "https://foo.bar/asdf", want: false},
 		{src: "git::https://foo.bar/asdf", want: false},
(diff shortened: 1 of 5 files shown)


internal/policy/source/git_config_test.go
```diff
@@ -30,7 +30,6 @@ func TestSourceIsFile(t *testing.T) {
 		want bool
 	}{
 		{src: "", want: false},
-		{src: "foo", want: true},
```
simonbaird, line 33: Why did this need removing? Is this a behavior change?

simonbaird, line 33: If it is intentional, would it be better to make it `want: false` instead of removing it?

simonbaird (approved):
Lgtm, but see my question about the (maybe unexpected?) test change.

internal/policy/source/git_config_test.go
robnester-rh: Yes, I think you're correct. I'll add this back and set it false.

PR #2305, opened 2025-02-07
Replace adoc refs with url

joejstuart:
The adoc ref comes through in the logs which isn't helpful, so this replaces it with a url.

https://issues.redhat.com/browse/EC-1098


joejstuart:
setting to draft to work on the regex

simonbaird (approved):
I didn't test it, but lgtm. The test coverage gives me confidence.

PR #2311, opened 2025-02-11
Use archful repo ids in cachi2 rpm lock file

simonbaird:
This is needed because of the policy change implemented in https://github.com/release-engineering/rhtap-ec-policy/pull/99 .
    
See also https://issues.redhat.com/browse/KONFLUX-6218
    
Ref: https://issues.redhat.com/browse/EC-1134

### DIFF
--- a/hack/update-rpm-lock.sh
+++ b/hack/update-rpm-lock.sh
@@ -80,6 +80,9 @@ for r in $(dnf repolist --setopt=reposdir="${base_image_repos}" --disabled --qui
     dnf config-manager --quiet --setopt=reposdir="${base_image_repos}" "${r}" --set-enabled
 done
 
+# convert ubi repo ids to archful format for better sbom data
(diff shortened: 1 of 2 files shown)


simonbaird:
The new revision just updates the rpm lock file format. I was going to do something similar when building without cachi2, but decided against it since we'd be failing EC for other reasons in that scenario, so it doesn't seem worth the effort.

For the record, it was something like this before the `RUN microdnf` line:

```
+# Use archful rpm repo ids for ubi rpms
+# (Only relevant if the rpms are not prefetched, since cachi2 creates its own repo file based on rpms.lock.yaml)
+RUN [ -f "/etc/yum.repos.d/ubi.repo" ] && sed -i 's/^\[ubi-9-/\[ubi-9-for-$basearch-/' /etc/yum.repos.d/ubi.repo
+
```

simonbaird:
Second commit is easier to review. The rest if the PR diff (i.e. the first commit) is because we didn't freshen that lock file for a while.

lcarva (approved):

PR #2312, opened 2025-02-12
build: update Dockerfile to reduce size

robnester-rh:
This commit updates the Dockerfile and Dockerfile.dist to reduce their final image size. This was acheived by:
* Installing all binaries and ca-certs in a build stage based on ubi-minimal
* Copying needed files in a build stage based on ubi-micro
* Combining COPY commands where possible to reduce number of layers generated.

Ref: EC-962

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,7 @@ LABEL \
   io.openshift.tags="conforma ec opa cosign sigstore"
 
 # Install tools we want to use in the Tekton task
-RUN microdnf upgrade --assumeyes --nodocs --setopt=keepcache=0 --refresh && microdnf -y --nodocs --setopt=keepcache=0 install git-core jq
+RUN microdnf upgrade --assumeyes --nodocs --setopt=keepcache=0 --refresh && microdnf -y --nodocs --setopt=keepcache=0 install gzip jq
 
 # Copy all the binaries so they're available to extract and download
 # (Beware if you're testing this locally it will copy everything from
@@ -86,6 +86,6 @@ COPY --from=build /build/LICENSE /licenses/LICENSE
 USER 1001
 
 # Show some version numbers for troubleshooting purposes
-RUN git version && jq --version && ec version && ls -l /usr/local/bin
+RUN jq --version && ec version && ls -l /usr/local/bin
 
 ENTRYPOINT ["/usr/local/bin/ec"]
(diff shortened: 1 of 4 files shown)


simonbaird:
Also I think there's some work to do with the rpms.lock.yaml.

Firstly, there are a bunch of dependencies there now that we maybe could still need. I'm thinking either those were needed by git-core, so that's why you don't need to copy them over from the minimal to the micro image, or, maybe we do still need to copy them.

(If the "remove git-core" was in a separate commit or PR, it would make it easy to know which it is.)

Running hack/update-rpm-lock.sh refresh the rpm lock file so give it a try.

Dockerfile
```diff
@@ -58,8 +70,13 @@ LABEL \
...
 
-# Install tools we want to use in the Tekton task
-RUN microdnf upgrade --assumeyes --nodocs --setopt=keepcache=0 --refresh && microdnf -y --nodocs --setopt=keepcache=0 install git-core jq
+# Copy in the packages from the 'packages' stage.
+# These are the binaries needed at runtime.
+COPY --from=packages /usr/bin/jq /usr/bin/gzip /usr/bin/
+# Copy the jq shared library that jq depends on.
+COPY --from=packages /usr/lib64/libjq.so.1 /usr/lib64/libonig.so.5 /usr/lib64/
+# Copy the ca certs from the 'packages' stage.
+COPY --from=packages /etc/ssl/certs/ca-bundle.crt /etc/ssl/certs/
```
simonbaird, line 79: Luiz: Copying like this breaks the SBOM content, but.. use install root?

Dockerfile
```diff
@@ -58,8 +70,13 @@ LABEL \
   io.k8s.display-name="${CLI_NAME}" \
   io.openshift.tags="conforma ec opa cosign sigstore"
 
-# Install tools we want to use in the Tekton task
-RUN microdnf upgrade --assumeyes --nodocs --setopt=keepcache=0 --refresh && microdnf -y --nodocs --setopt=keepcache=0 install git-core jq
+# Copy in the packages from the 'packages' stage.
+# These are the binaries needed at runtime.
+COPY --from=packages /usr/bin/jq /usr/bin/gzip /usr/bin/
```
lcarva, line 75: Consider [this](https://github.com/enterprise-contract/ec-cli/pull/955#issuecomment-1688787997) approach to retain the rpm db record to make security folks happy.

robnester-rh:
Closing in lieu of #2345

PR #2314, opened 2025-02-12
dropping the data output target

cuipinghuo:
dropping the `data` output target
refer: https://issues.redhat.com/browse/EC-1061

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -170,9 +170,6 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 
 			  ec validate image --image registry/name:tag --output yaml --output appstudio=<path>
 
-			Write the data used in the policy evaluation to a file in YAML format
-
-			  ec validate image --image registry/name:tag --output data=<path>
 
 			Validate a single image with keyless workflow.
 
@@ -302,7 +299,6 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			type result struct {
 				err         error
 				component   applicationsnapshot.Component
-				data        []evaluator.Data
 				policyInput []byte
 			}
 
@@ -382,7 +378,6 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 							res.component.Attestations = append(res.component.Attestations, attResult)
 						}
 						res.component.ContainerImage = out.ImageURL
-						res.data = out.Data
 						res.policyInput = out.PolicyInput
 					}
 					res.component.Success = err == nil && len(res.component.Violations) == 0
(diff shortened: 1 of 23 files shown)


cmd/validate/image.go
```diff
@@ -446,7 +443,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 				data.output = append(data.output, fmt.Sprintf("%s=%s", applicationsnapshot.JSON, data.outputFile))
 			}
 
-			report, err := applicationsnapshot.NewReport(data.snapshot, components, data.policy, evaluatorData, manyPolicyInput, showSuccesses)
+			report, err := applicationsnapshot.NewReport(data.snapshot, components, data.policy, manyPolicyInput, showSuccesses)
```
lcarva, line 446: Since `evaluatorData` is no longer being used, remove it from above as well. This should also cascade in other things being removed. For example, you'll remove this line:

```
evaluatorData = append(evaluatorData, r.data)
```

Look at `r.data`, and remove that too, and whatever populates that, e.g. `res.data = out.Data`. Then keeping doing that until there's nothing else to remove.

lcarva: Similar changes are needed in `cmd/validate/input.go` as well.

simonbaird:
Not 100% sure, but I think this needs to come out too:

```diff
diff --git a/internal/output/output.go b/internal/output/output.go
index 6d00805e..59ced805 100644
--- a/internal/output/output.go
+++ b/internal/output/output.go
@@ -85,7 +85,6 @@ type Output struct {
        Attestations              []attestation.Attestation   `json:"attestations,omitempty"`
        ImageURL                  string                      `json:"-"`
        Detailed                  bool                        `json:"-"`
-   Data                      []evaluator.Data            `json:"-"`
        Policy                    policy.Policy               `json:"-"`
        PolicyInput               []byte                      `json:"-"`
 }

```

cuipinghuo:
> Not 100% sure, but I think this needs to come out too:
> 
> ```diff
> diff --git a/internal/output/output.go b/internal/output/output.go
> index 6d00805e..59ced805 100644
> --- a/internal/output/output.go
> +++ b/internal/output/output.go
> @@ -85,7 +85,6 @@ type Output struct {
>         Attestations              []attestation.Attestation   `json:"attestations,omitempty"`
>         ImageURL                  string                      `json:"-"`
>         Detailed                  bool                        `json:"-"`
> -   Data                      []evaluator.Data            `json:"-"`
>         Policy                    policy.Policy               `json:"-"`
>         PolicyInput               []byte                      `json:"-"`
>  }
> ```

resolved.

simonbaird:
Since it's green, I think we could merge it, and create another PR with that one extra change.

cuipinghuo:
> Since it's green, I think we could merge it, and create another PR with that one extra change.

@simonbaird the change is already included in this pr.

simonbaird:
> > Since it's green, I think we could merge it, and create another PR with that one extra change.
> 
> @simonbaird the change is already included in this pr.

Oh, sorry I didn't see that for some reason. Thanks!

PR #2327, opened 2025-02-19
feat(RELEASE-1425): Conforma Konflux based task

scoheb:
- uses Trusted Artifacts to obtain the Snapshot


joejstuart:
ok-to-test

simonbaird:
I think we could merge it. It does need some testing though, especially if you're going to start using it right away.

scoheb:
It will only get used in production once we have a bundle being produced that has been tested with CI.

We can merge now and use a git resolver for pre-production testing

PR #2329, opened 2025-02-19
chore(dockerfile): Update ubi reference.

robnester-rh:
This commit updates the UBI version to address CVE-2022-9043.

Ref: EC-1139


robnester-rh:
Closed in favor of 2331

PR #2331, opened 2025-02-19
EC-1139 - Ensure all EC builds have libxml2-2.9.13-6.el9_5.1 or better

robnester-rh:


simonbaird:
No rpm lock file in v0.4 branch. It's green so let's merge.

PR #2347, opened 2025-02-26
Support image ref and digest

joejstuart:
This adds support for exceptions being created using an image url and/or just a digest.

https://issues.redhat.com/browse/EC-1100


features/validate_image.feature
```diff
@@ -465,6 +465,46 @@ Feature: evaluate enterprise contract
...
+      | filtering.rego | examples/filtering.rego |
+    Given policy configuration named "ec-policy" with specification
+    """
+    {
+      "sources": [
+        {
+          "volatileConfig": {
+            "exclude": [
+              {
+                "value": "filtering.always_fail",
```
lcarva, line 485: This is not included given the `config` below. Intentional?

lcarva: Would it be simpler to simply include `filtering`?

internal/evaluator/criteria.go
```diff
@@ -48,6 +50,8 @@ func (c *Criteria) addItem(key, value string) {
 		if c.digestItems == nil {
 			c.digestItems = make(map[string][]string)
 		}
+		// the image url can contain https://
+		key = cleanKey(key)
```
lcarva, line 54: How?

internal/evaluator/criteria.go
```diff
@@ -86,33 +122,8 @@ func computeIncludeExclude(src ecc.Source, p ConfigProvider) (*Criteria, *Criter
 
 	vc := src.VolatileConfig
 	if vc != nil {
-		at := p.EffectiveTime()
```
lcarva, line 89: This feels like a good change but unrelated? Maybe put it in a different commit?

lcarva (approved):

internal/evaluator/criteria.go
joejstuart: I allowed users to add the component/image url with a `https://` prefix. But now that I think about it, I'm not sure that adds any value. I may remove that option.

internal/evaluator/criteria.go
joejstuart: Yes, unrelated. I split it off. https://github.com/enterprise-contract/ec-cli/pull/2347/commits/488dba5fdb30f8ac272976411de24be0f3c80c55

features/validate_image.feature
joejstuart: I copied/pasted from another test. Doesn't make sense it's not included though, so fixed both.

PR #2361, opened 2025-03-05
Use Tekton's conditional step instead of bash

lcarva:
This commit changes the `initialize-tuf` step of the `verify-enterprise-contract` Tekton Task to use a `when` expression instead of using bash as the conditional.

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -174,20 +174,18 @@ spec:
 
     - name: initialize-tuf
       image: quay.io/enterprise-contract/ec-cli:snapshot
-      script: |-
-        set -euo pipefail
-
-        if [[ -z "${TUF_MIRROR:-}" ]]; then
-            echo 'TUF_MIRROR parameter not provided. Skipping TUF root initialization.'
-            exit
-        fi
-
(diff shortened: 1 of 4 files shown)


lcarva:
Ah need to add a placeholder for the timestamps in the snapshot:
```
            -   "ts": 1741205504.415519,
            +   "ts": 1741206615.4840739,
```

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -174,20 +174,19 @@ spec:
...
+        - initialize
+        - --mirror
+        - $(params.TUF_MIRROR)
+        - --root
+        - $(params.TUF_MIRROR)/root.json
+      when:
+        - input: $(params.TUF_MIRROR)
+          operator: "notin"
+          values: [""]
+        # - cel: "$(params.TUF_MIRROR).size() > 0"
```
simonbaird, line 189: Intentional?

simonbaird (approved):

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
lcarva: Good catch! We could use a cel expression, but that would require enabling another feature-flag. I don't think it's worth it. The alternative of using `input/operator/values` seems good enough.

PR #2366, opened 2025-03-06
Create Konflux pipelines for v0.6 branch

simonbaird:
Notes:
- I did this manually based on the current pipeline definitions in main branch. Generally Konflux should create a PR for us, but for some reason it didn't work this time. See https://issues.redhat.com/browse/KFLUXSPRT-2145

- I removed the build-tekton-bundle task, which creates a Tekton bundle in the main branch Konflux build because we don't plan to create a Tekton bundle in the release branch builds.

- One reason we might *want* to use the generated PR is that we get a fresh pipeline definition. But I think actually we should do that in main branch first, using the method described at https://konflux.pages.redhat.com/docs/users/how-tos/configuring/reconfiguring-build-pipeline.html I think overall it's better if we stick to the same pipeline from main branch when cutting a release branch.

Ref: https://issues.redhat.com/browse/EC-1135

### DIFF
--- a/.tekton/cli-build.yaml
+++ b/.tekton/cli-build.yaml
@@ -28,15 +28,6 @@ spec:
     - description: Fully Qualified Output Image
       name: output-image
       type: string
-    - description: >-
-        OCI repository of the CLI image to use as a reference in the Tekton bundle. When setting
-        this value, take into account where the CLI image will be available for usage. For certain
-        workflows, e.g. pull request, this should be the repo in which the CLI image is built into
-        because those CLI images are not promoted to another location. For merge workflows that go
-        through a release, for example, this should be the repository for which the CLI image will
-        be released to.
-      name: bundle-cli-ref-repo
-      type: string
     - default: .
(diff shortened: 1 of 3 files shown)


simonbaird:
Hmm:

```
step-create-trusted-artifact
Prepared artifact from /var/workdir/source (sha256:...)
Token not found for quay.io/redhat-user-workloads/rhtap-contract-tenant/ec-v06/cli-v06
Error response from registry: recognizable error message not found: HEAD "https://quay.io/v2/redhat-user-workloads/rhtap-contract-tenant/ec-v06/cli-v06/manifests/sha256:...": response status code 401: Unauthorized
Command exited with non-zero status 1
```

simonbaird:
Build failures might be fixed with [this](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/4814) merged.

simonbaird:
I think #2368 is preferred now, but they're pretty similar.

simonbaird:
Decided to go with #2368 .

PR #2379, opened 2025-03-13
feat(KONFLUX-7335): Setup limits and resource for verify-enterprise-contract

rh-rahulshetty:
### Description

This PR updates resource limits and requests for the task **verify-enterprise-contract** as part of this effort - [KONFLUX-7127](https://issues.redhat.com/browse/KONFLUX-7127). 

The new request/limit values have been defined taking into account usages observed historically (tracked in this [spreadsheet](https://docs.google.com/spreadsheets/d/1d4Tn7PYkY2EVgpNyT29xL7y62akZNPIjy18VTzw54iQ/edit?gid=1735880927#gid=1735880927)) in the **stone-prd-rh01** cluster.

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -269,7 +269,7 @@ spec:
           value: "false"
       computeResources:
         requests:
-          cpu: 1200m
+          cpu: 1800m
           memory: 2Gi
         limits:
           memory: 2Gi
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -173,6 +173,12 @@ spec:
   steps:
 
     - name: initialize-tuf
+      computeResources:
+        requests:
+          cpu: 100m
+          memory: 256Mi
+        limits:
+          memory: 256Mi
       image: quay.io/enterprise-contract/ec-cli:snapshot
       command: [ec]
       args:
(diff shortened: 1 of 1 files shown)


tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -268,13 +280,25 @@ spec:
           readOnly: true
```
lcarva, line 280: I can't comment further up, but the data seems to indicate that `cpu: 250m` is not enough. Should this be `1` (`1000m`) instead or maybe even higher?

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -268,13 +280,25 @@ spec:
...
 
     - name: report-json
+      computeResources:
+        requests:
+          cpu: 100m
+          memory: 2Gi
```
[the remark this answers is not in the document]
rh-rahulshetty: Thank for your pointing it out, I think I made a mistake while copying the values. Anyway I've updated it to `256Mi`.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
rh-rahulshetty: 95% CPU Usage over last 2 weeks

After checking the latest runs, it seems the CPU usage for this step gets close to 2 cpu cores. I've updated this request value to `1800m` for now.

lcarva (approved):

simonbaird:
I'm having trouble seeing the violations, but it might be the konflux references need a bump. Will take a closer look soon.

simonbaird:
I rebased on latest main, let's see if the builds goes green.

PR #2384, opened 2025-03-14
Test updating github actions with renovate

joejstuart:
This is part of an automated PR cleanup effort. 

https://issues.redhat.com/browse/EC-1143

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,5 +3,12 @@
   "extends": ["github>konflux-ci/mintmaker//config/renovate/renovate.json"],
   "timezone": "America/New_York",
   "schedule": ["before 10pm"],
-  "enabledManagers": ["asdf", "tekton"]
+  "enabledManagers": ["asdf", "tekton", "github-actions"],
+  "github-actions": {
+    "enabled": true,
+    "schedule": ["before 10pm"],
+    "separateMinorPatch": true,
+    "groupName": "GitHub Actions Updates"
+  }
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -3,5 +3,12 @@
   "extends": ["github>konflux-ci/mintmaker//config/renovate/renovate.json"],
   "timezone": "America/New_York",
   "schedule": ["before 10pm"],
-  "enabledManagers": ["asdf", "tekton"]
+  "enabledManagers": ["asdf", "tekton", "github-actions"],
+  "github-actions": {
+    "enabled": true,
+    "schedule": ["before 10pm"],
```
lcarva, line 9: Won't this increase the frequency of the github-actions updates from weekly to daily?

joejstuart: yes, but it should group them and that will help me while I try and sort through this. I can make a note in the Jira to move it back to weekly if we prefer that.

renovate.json
```diff
@@ -3,5 +3,12 @@
...
+    "separateMinorPatch": true,
```
simonbaird, line 10: What does this one mean?

simonbaird (approved):

renovate.json
joejstuart: it should break the PRs up. https://github.com/renovatebot/renovate/blob/main/docs/usage/faq.md

PR #2394, opened 2025-03-18
Include reduce-snapshot in Konflux built image

simonbaird:
This was added for the upstream build only, and we didn't notice until now, since RHTAP pipelines are starting to use an updated task definition that include the reduce-snapshot step.

Ref: https://issues.redhat.com/browse/RHTAP-4567
(cherry picked from commit 5efea4e5bd8459fb83b2c8c2a31baf6713e21c27 in release-v0.6 branch)


simonbaird:
See also #2393 .

lcarva (approved):

PR #2421, opened 2025-03-21
Rename Enterprise Contract to Conforma

dheerajodha:
* Also, updated the License Header, and
* Updated the links to the Conforma site

resolves: EC-1129


simonbaird:
I think you need to do a `make generate` and check in the changes it produces.

One note: Not a blocker for this PR, but I would have suggested doing the license header change it its own separate commit.

dheerajodha:
> One note: Not a blocker for this PR, but I would have suggested doing the license header change it its own separate commit.

Thanks! I added new commits containing specific changes.

PR #2460, opened 2025-04-01
add step show-config

cuipinghuo:
Add a step to the verify-enterprise-contract task to output the policy.config used

refer:


features/task_validate_image.feature
```diff
@@ -257,6 +257,7 @@ Feature: Verify Enterprise Contract Tekton Tasks
     And the task logs for step "initialize-tuf" should match the snapshot
      And the task logs for step "report" should match the snapshot
      And the task logs for step "summary" should match the snapshot
+     And the task logs for step "show-config" should match the snapshot
```
simonbaird, line 260: If you add something like this, there needs to be a corresponding snapshot file change somewhere in `features/__snapshots__/task_validate_image.snap` so it has something to match.

There's a way to generate it, but you could try creating it by hand. Look for the ` [Outputs are there:summary - 1]` and `[Outputs are there:assert - 1]` sections. Right between those you can add a `[Outputs are there:show-config - 1]` and inside it you can put the expected output, then the `---` delimiter.

You can start with any garbage in there, and then run the test. It will show you what it got and what was expected, which you can then use to update the snapshot file.

tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
```diff
@@ -308,6 +308,13 @@ spec:
       command: [ec]
       args:
         - version
+    
+    - name: show-config
+      image: quay.io/enterprise-contract/ec-cli:snapshot
+      command: [jq]
+      args:
+        - ".policy.sources[].config"
+        - "json=$(params.HOMEDIR)/report-json.json"
```
simonbaird, line 317: Not sure about the `json=` here. Maybe this?

```suggestion
        - "$(params.HOMEDIR)/report-json.json"
```

In any case I would expect it to be the same as in the other file.

cuipinghuo: the error
```
Status  Error
----- Logs -----
jq: error (at /tekton/home/report-json.json:1): Cannot iterate over null (null)
```
shows /tekton/home/report-json.json is null.

cuipinghuo: it is not the jq, but the test case input iteself doesn't have a field of policy.sources[].config. Move test case to a different scenario and have proper snapshot defined solve this.

features/task_validate_image.feature
cuipinghuo: @simonbaird Could you PTAL?

features/__snapshots__/task_validate_image.snap
```diff
@@ -272,6 +272,14 @@ policy:
...
 success: true
 
+[Golden container image:show-config - 1]
+{
+  "include": [
+    "slsa_provenance_available"
+  ]
+}
+---
+
```
simonbaird, line 282: What I expected to see here is the full content of the config, i.e. the whole thing from line 13 to line 28 in [features/task_validate_image.feature](https://github.com/enterprise-contract/ec-cli/pull/2460/files#diff-658ed1262bea09265fe40c9bf025b6b9d126a20e5800c4db958767b7c791cbbb)

This looks like just a part of it.

Not sure why, but see if you can figure it out maybe..?

features/task_validate_image.feature
simonbaird: Looks alright, but see my other comment about the content of the snapshot change.

tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
```diff
@@ -308,6 +308,13 @@ spec:
       command: [ec]
       args:
         - version
+    
+    - name: show-config
+      image: quay.io/enterprise-contract/ec-cli:snapshot
+      command: [jq]
+      args:
+        - ".policy.sources[0].config"
```
simonbaird, line 316: Oh I think it should be just this:

```suggestion
        - ".policy"
```

(I guess the name is confusing since we call different things "policy" and "config" in different contexts.)

cuipinghuo: jq .policy would show all the content of policy, not just config.

features/__snapshots__/task_validate_image.snap
cuipinghuo: Do we expect all the scenarios in [features/task_validate_image.feature](https://github.com/enterprise-contract/ec-cli/pull/2460/files#diff-658ed1262bea09265fe40c9bf025b6b9d126a20e5800c4db958767b7c791cbbb) expect a check for the show-config? I only add [Golden container image:show-config - 1] right now. Thanks.

features/task_validate_image.feature
```diff
@@ -15,8 +16,8 @@ Feature: Verify Enterprise Contract Tekton Tasks
         "sources": [
           {
             "policy": [
-              "github.com/enterprise-contract/ec-policies//policy/release",
-              "github.com/enterprise-contract/ec-policies//policy/lib"
+              "github.com/enterprise-contract/ec-policies//policy/release?ref=0de5461c14413484575e63e96ddb514d8ab954b5",
```
cuipinghuo, line 19: I need to explicitly add a commit id here because report-json.json added the latest commit id if not specified.

cuipinghuo: @simonbaird could you PTAL if this make sense to you?

features/task_validate_image.feature
```diff
@@ -1,3 +1,4 @@
+@focus
```
simonbaird, line 1: This should not be checked in.

features/task_validate_image.feature
simonbaird: Yeah, it makes sense.

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -367,4 +380,4 @@ spec:
         items:
           - key: $(params.CA_TRUST_CONFIG_MAP_KEY)
             path: ca-bundle.crt
-        optional: true
+        optional: true
```
simonbaird, line 383: Here also.

tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
```diff
@@ -336,4 +343,4 @@ spec:
             path: ca-bundle.crt
         optional: true
     - name: workdir
-      emptyDir: {}
+      emptyDir: {}
```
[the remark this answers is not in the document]
simonbaird: One way to do it:
```
git checkout -p HEAD^ # and choose just this hunk to remove
```

simonbaird (approved):

PR #2461, opened 2025-04-01
Use global manager settings

joejstuart:
This update removes disabling go updates and
inherits from the global setting of updating
the minor and patch versions.


robnester-rh (approved):
LGTM

simonbaird (approved):

PR #2469, opened 2025-04-02
Upgrade golang version

joejstuart:
Upgrading to 1.23.6 as certain modules are
requiring it for security upgrades.

https://issues.redhat.com/browse/EC-1205

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.22.5
+golang 1.23.6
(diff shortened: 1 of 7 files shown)


simonbaird:
Does it need a `go mod tidy` or something? I'm comparing it with #2471 .

joejstuart:
> Does it need a `go mod tidy` or something? I'm comparing it with #2471 .

Just tried it and it didn't update anything.

PR #2474, opened 2025-04-07
Avoid maximum argument length

joejstuart:
Use printf in place of echo to avoid shell
argument expansion.

https://issues.redhat.com/browse/KONFLUX-7603


hack/reduce-snapshot.sh
```diff
@@ -65,4 +65,4 @@ if [ "${SINGLE_COMPONENT}" == "true" ]; then
 fi
 
 # we need to create snapshot file to be passed to later stages.
-echo "${SNAPSHOT}" | jq '.' | tee "${SNAPSHOT_PATH}"
+printf "%s" "${SNAPSHOT}" | jq '.' | tee "${SNAPSHOT_PATH}"
```
lcarva, line 68: Interesting. Have you been able to reproduce the issue and verify this indeed fixed it?

joejstuart: I can't reproduce it, but from reading up on it, it sounds like `printf` might be helpful here. It's a low risk change, so I thought it would be worth the try. The resource I used. https://www.in-ulm.de/~mascheck/various/argmax/

joejstuart: I do wonder if there's something else going on too, because 174 images in a file just doesn't seem like enough to cause a problem. But either way, `printf` _should_ be more efficient.

lcarva: Nice resource! I do think the `printf` option there works only because it's using a null character and pipeline through xargs with a null character limit. It probably won't apply here.

Maybe instead of holding on to the snapshot in a variable it can just be a file. That way we don't have to pipe it to jq. We can just have jq read the file directly.

Anyways, it's worth trying out different things. It would be great to verify that any of these options (including printf) do fix the problem before having to go through the whole release process.

joejstuart: After looking at the script better, this all might be unnecessary. We're reading it in from a file and then dumping it back to a file without any changes.

joejstuart: Updated to read from a file. PTAL when you can. Thanks!

hack/reduce-snapshot.sh
```diff
@@ -29,9 +29,18 @@ set -o errexit
...
+# Make sure to move snapshot contents to the WORKING_SNAPSHOT location. Then allow jq to
+# work with it there. This avoids having to read SNAPSHOT to memory.
+
+if [[ -f "$SNAPSHOT" ]]; then
+  jq empty "$SNAPSHOT" || { echo "JSON is invalid"; exit 1; }
+  WORKING_SNAPSHOT="$SNAPSHOT"
+else
+  WORKING_SNAPSHOT="$(mktemp snapshot.XXXXXX)"
+  printf "%s" "$SNAPSHOT" > "$WORKING_SNAPSHOT"
+  jq empty "$WORKING_SNAPSHOT" || { echo "JSON is invalid"; exit 1; }
```
simonbaird, line 42: Not a blocker, but: you could move this outside the if block and remove line 37.

hack/reduce-snapshot.sh
```diff
@@ -65,4 +76,4 @@ if [ "${SINGLE_COMPONENT}" == "true" ]; then
 fi
 
 # we need to create snapshot file to be passed to later stages.
-echo "${SNAPSHOT}" | jq '.' | tee "${SNAPSHOT_PATH}"
+cp "$WORKING_SNAPSHOT" "$SNAPSHOT_PATH"
```
simonbaird, line 79: Why remove the tee? I think it can be kept safely like this:

```suggestion
jq '.' "${WORKING_SNAPSHOT}" | tee "${SNAPSHOT_PATH}"
```

joejstuart: I just didn't know what value any of it added. If there's value in it, I can definitely add it back.

simonbaird: I presume the idea was to make it visible in the taskRun log.

simonbaird: We could reconsider if that's useful, but I think it's better not to change it in this PR.

simonbaird: I guess it would fit better with the debug output, since the component list is clearly shown in the next step. (I'm looking in https://console.redhat.com/application-pipeline/workspaces/rhtap-contract/applications/ec-main-ci/pipelineruns/ec-main-ci-enterprise-contract-9jfvx ).

(But still, better to change that it a later PR.)

simonbaird (approved):
Very nice

PR #2475, opened 2025-04-08
Allow using rekor public key in ec.sigstore func

lcarva:
This commit extends the `ec.sigstore.verify_image` and `ec.sigstore.verify_attestation` to accept a new option, `rekor_public_key`. When provided, an offline Rekor verification is performed on the signature.

Ref: EC-1200

### DIFF
--- a/docs/modules/ROOT/pages/ec_sigstore_verify_attestation.adoc
+++ b/docs/modules/ROOT/pages/ec_sigstore_verify_attestation.adoc
@@ -4,12 +4,12 @@ Use sigstore to verify the attestation of an image.
 
 == Usage
 
-  result = ec.sigstore.verify_attestation(ref: string, opts: object<certificate_identity: string, certificate_identity_regexp: string, certificate_oidc_issuer: string, certificate_oidc_issuer_regexp: string, ignore_rekor: boolean, public_key: string, rekor_url: string>)
+  result = ec.sigstore.verify_attestation(ref: string, opts: object<ignore_rekor: boolean>[string: string])
 
 == Parameters
 
 * `ref` (`string`): OCI image reference
-* `opts` (`object<certificate_identity: string, certificate_identity_regexp: string, certificate_oidc_issuer: string, certificate_oidc_issuer_regexp: string, ignore_rekor: boolean, public_key: string, rekor_url: string>`): Sigstore verification options
+* `opts` (`object<ignore_rekor: boolean>[string: string]`): Sigstore verification options. Dynamic string properties: `certificate_identity`, `certificate_identity_regexp`, `certificate_oidc_issuer`, `certificate_oidc_issuer_regexp`, `public_key`, `rekor_url`, `rekor_public_key`.
 
 == Return
 
--- a/internal/rego/sigstore/sigstore.go
+++ b/internal/rego/sigstore/sigstore.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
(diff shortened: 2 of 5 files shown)


lcarva:
I wanted to add more acceptance tests but it was not possible due to EC-1210.

robnester-rh (approved):
LGTM

docs/modules/ROOT/pages/ec_sigstore_verify_attestation.adoc
```diff
@@ -4,12 +4,12 @@ Use sigstore to verify the attestation of an image.
 
 == Usage
 
-  result = ec.sigstore.verify_attestation(ref: string, opts: object<certificate_identity: string, certificate_identity_regexp: string, certificate_oidc_issuer: string, certificate_oidc_issuer_regexp: string, ignore_rekor: boolean, public_key: string, rekor_url: string>)
+  result = ec.sigstore.verify_attestation(ref: string, opts: object<ignore_rekor: boolean>[string: string])
```
simonbaird, line 7: Not sure if it's intentional to keep just one, but should it be this, since the keys for `opts` are documented below?

```suggestion
  result = ec.sigstore.verify_attestation(ref: string, opts: object[string: string])
```

simonbaird: Oh, maybe it's generated, in which case disregard..?

lcarva: Yeah. It's generated :)

internal/rego/sigstore/sigstore.go
```diff
@@ -52,23 +54,36 @@ const (
...
+	)).Description(
+	fmt.Sprintf("Sigstore verification options. Dynamic string properties: `%s`.",
+		strings.Join([]string{
+			certificateIdentityAttribute,
+			certificateIdentityRegExpAttribute,
+			certificateOIDCIssuerAttribute,
+			certificateOIDCIssuerRegExpAttribute,
+			publicKeyAttribute,
+			rekorURLAttribute,
+			rekorPublicKeyAttribute,
```
simonbaird, line 84: It's not obvious why we're changing the static to dynamic for these others. I assume the reasoning is good, but consider mentioning it in the commit message.

simonbaird (approved):

internal/rego/sigstore/sigstore.go
lcarva: I tried, unsuccessfully, to explain this in the large commit block above. :sob: 

Can you help me with the specifics of what is not clear?

lcarva: Maybe this... I made them dynamic properties so they are not all always required. Without this, something like this will cause a compile error:

```rego
ec.signstore.verify_image(ref, {"public_key": my_key, "ignore_rekor": false})
```

You need to provide all attributes, always:

```rego
ec.signstore.verify_image(ref, {
  "public_key": my_key,
 "ignore_rekor": false,
  "certificate_identity": "",
  "certificate_identity_regexp": "",
  "certificate_oidc_issuer" "",
  ...
})
```

With this change, the first form works.

lcarva: Ok, added an explanation to the commit message. I'll merge once CI passes.

simonbaird: Sorry, I didn't notice the detailed explanation in the comments, just went looking for a commit message.. :sweat_smile:

simonbaird: Commit message is perfect!

PR #2478, opened 2025-04-11
Set BUILD_SUFFIX build arg for main branch builds

simonbaird:
Use the build-args-file pipeline param to do it. I considered using the build-args pipeline param instead, but since we're already using build-args-file figured let's continue using it.

Another reasonable way to do this would have been to change the default value and then use a build arg for the release branch builds, but I decided this way was a little nicer since we have multiple release branches and only one main branch.

Ref: https://issues.redhat.com/browse/EC-1223

### DIFF
--- a/.tekton/cli-main-ci-pull-request.yaml
+++ b/.tekton/cli-main-ci-pull-request.yaml
@@ -39,7 +39,7 @@ spec:
     - name: build-source-image
       value: 'true'
     - name: build-args-file
-      value: quick-build-args.conf
+      value: main-pre-merge-build-args.conf
     - name: hermetic
       value: 'true'
   pipelineRef:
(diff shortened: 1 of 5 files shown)


simonbaird:
Something's wrong, I'm seeing this still:
```
EC_FULL_VERSION=v0.7.43-ci-09917c17+redhat
```

The actual pipeline failure seems unrelated - it's in the sast-unicode-check:
```
the step "upload" in TaskRun "cli-main-ci-on-pull-request-k44l7-sast-unicode-check" failed to pull the image "". The pod errored with the message: "Back-off pulling image "quay.io/konflux-ci/oras:latest@sha256:d9fea2ee280880feef6909bef3e18318444231c83736bcc41d54b4e5064f23c9"."
```

PR #2479, opened 2025-04-16
Clarify how to run acceptance tests locally

st3penta:
This commit makes it clear that the acceptance folder is a go module on its own, and that to run the acceptance tests is important to `cd` into the directory.
It also removes all the references to the `acceptance` tag, since it is not found in the tests code


simonbaird (approved):
The explanation in the PR is really nice and helps me understand the change. My preference would be to put that also in the commit message. Reasoning: Suppose I'm looking at this in (say) six months time, I might do a git blame and a git show. If the commit message has the explanation, that's more convenient than having to then go into github and locate the PR to find it.

Also: It's not particularly likely I guess, but suppose we move to another git hoster such as gitlab, or codeberg, bitbucket, or something other internal thing. Now the PR descriptions are quite likely lost forever, but commit messages are safe and permanent.

That said, not a blocker for this PR, just sharing my thoughts on commit messages.

dheerajodha (approved):
Thanks for the PR!

PR #2483, opened 2025-04-18
Include 'term' attribute in report by default

st3penta:
Add the 'term' attribute in the YAML and JSON reports produced by 'ec validate image', eliminating the need for the '--info' flag. The term may be useful in include/exclude clauses

Ref: https://issues.redhat.com/browse/EC-673

### DIFF
--- a/features/__snapshots__/validate_image.snap
+++ b/features/__snapshots__/validate_image.snap
@@ -76,7 +76,7 @@
 ---
 
 [JUnit and AppStudio output format:stdout - 1]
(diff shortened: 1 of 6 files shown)


robnester-rh (approved):
LGTM

features/__snapshots__/validate_image.snap
```diff
@@ -1944,6 +1955,7 @@ Results:
 [31m✕[0m [31m[Violation] main.reject_with_term[0m
   ImageRef: ${REGISTRY}/acceptance/image@sha256:${REGISTRY_acceptance/image:latest_DIGEST}
   Reason: Fails always (term2)
+  Term: [term2 term3]
```
simonbaird, line 1958: Idea: This output is for human consumption, so we could make it a little nicer by formatting a list something like this:

```suggestion
  Terms: term2, term3
```

(We could worry about ambiguity for terms with spaces or commas in them, but that feels like it will not be a problem in reality, i.e. an edge case we could ignore.)

simonbaird: Thinking about how to do that: It might be a bit fiddly with the templating. I was thinking a helper in `internal/utils/templates.go`, either to output nice term list or to detect if term is a list of strings or a single string could be useful.

simonbaird (commented):
Nice patch! I have a little nitpick about the text output, but otherwise looks great.

simonbaird:
It's hard to tell, but I think the [Red Hat Konflux / ec-main-ci-enterprise-contract / cli-main-ci](https://github.com/enterprise-contract/ec-cli/pull/2483/checks?check_run_id=40930369981) failure is due to trusted task violations. A rebase on latest upstream/main should fix it, particularly https://github.com/enterprise-contract/ec-cli/commit/b9ef669e241e850861e8e2c80dd1ced065bf4076

simonbaird:
Ideally the Konflux page you see by following PipelineRun link [here](https://github.com/enterprise-contract/ec-cli/pull/2483/checks?check_run_id=40930369981) would show the violations, but it seems to be broken currently. :cry:

simonbaird (approved):

PR #2480, opened 2025-04-16
Use new Konflux built Tekton Bundle in Red Hat Konflux

simonbaird:
Update two scripts used by the automation which updates build-definitions and infra-deployments.

See also https://github.com/enterprise-contract/infra-deployments-ci/blob/main/.github/workflows/create-pr.yaml .

Ref: [EC-913](https://issues.redhat.com/browse/EC-913)

PR #2484, opened 2025-04-18
Upgrade tkn bundle task to 0.2

simonbaird:
(Split from #2480 so I can test the automation with a freshly built Tekton bundle.)

Updates the tkn bundle task to 0.2 to get the new annotations. (Also includes a references bump for the other tasks, since we seems to not be getting those created automatically lately.)

Following rabbit holes related to:
Ref: [EC-913](https://issues.redhat.com/browse/EC-913)

### DIFF
--- a/.tekton/cli-build.yaml
+++ b/.tekton/cli-build.yaml
@@ -113,7 +113,7 @@ spec:
           - name: name
             value: init
           - name: bundle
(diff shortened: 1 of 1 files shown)


simonbaird:
I don't think it needs review, so I will likely merge this once it turns green.

PR #2485, opened 2025-04-21
KFLUXDP-171: Onboard ec-cli checks to Sealights

ascerra:
This commit onboards ec-cli to Sealights

Adding a new github workflow that will:
- install and use sealights to scan repo
- run unit tests and acceptance tests with sealights agent reporting to [Sealights dashboard ](https://redhat.sealights.co/dashboard/v2)

For acceptance tests
- We needed an xml output file to upload to sealights to record the acceptance test run.
- Did this by adding the tool gotestsum then adding the option to the acceptance test run command
- Created separate acceptance test in makefile for now.

We also need to add a working Sealights token to a secret named `SEALIGHTS_AGENT_TOKEN` for the github workflow

# Testing 
To test I needed to create [this PR](https://github.com/ascerra/ec-cli/pull/19) against my forks main branch which is up to date with ec-cli main and I added the workflow, new make command and tools to my forks main with this [commit](https://github.com/ascerra/ec-cli/commit/5fff7c9475352212effa614d4a15edaf15ac07ac). Doing this so that I can show working code as the a workflow being added here cannot be tested as it needs to already exist in the main branch to be able to use the github actions secret required.
- Passing PR job: https://github.com/ascerra/ec-cli/actions/runs/14844729829
- Passing push job: https://github.com/ascerra/ec-cli/actions/runs/14845155455

resolves: KFLUXDP-171


cuipinghuo (approved):

Makefile
```diff
@@ -123,12 +123,24 @@ acceptance: ## Run all acceptance tests
...
-	go run acceptance/coverage/coverage.go && \
 	$(MAKE) build && \
 	export COVERAGE_FILEPATH="$$ACCEPTANCE_WORKDIR"; \
 	export COVERAGE_FILENAME="-acceptance"; \
-	cd acceptance && go test -coverprofile "$$ACCEPTANCE_WORKDIR/coverage-acceptance.out" -timeout $(ACCEPTANCE_TIMEOUT) ./... && \
-	go run -modfile "$$ACCEPTANCE_WORKDIR/tools/go.mod" github.com/wadey/gocovmerge "$$ACCEPTANCE_WORKDIR/coverage-acceptance.out" > "$(ROOT_DIR)/coverage-acceptance.out"
+	cd acceptance && SEALIGHTS_LOG_LEVEL=none go run -modfile "$$ACCEPTANCE_WORKDIR/tools/go.mod" gotest.tools/gotestsum --junitfile "$(ROOT_DIR)/junit-acceptance.xml" -- -parallel 1 -timeout $(ACCEPTANCE_TIMEOUT) ./...
+
+acceptance-sealights: ## Run all acceptance tests with sealights integration
+	@ACCEPTANCE_WORKDIR="$$(mktemp -d)"; \
```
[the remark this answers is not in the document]
ascerra: sorry I need to fix the first acceptance block, I'll update now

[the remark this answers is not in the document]
ascerra: Alright I removed any changes to the first acceptance block, there shouldn't have been changes regarding sealights there, it was left over from previous testing.

We are adding a new a command acceptance-sealights due to codecov and sealights not working well together on the same build. It could be possible to make them work in the future but for now I wanted to see if we could isolate them.

.gitignore
```diff
@@ -180,3 +180,6 @@ hack/**/charts
 
 # Pregenerated benchmark data
 benchmark/*/data.tar.gz
+
+# JUnit file output from acceptance tests
+ junit-acceptance.xml
```
simonbaird, line 185:
```suggestion
junit-acceptance.xml
```

.github/workflows/checks-sealights.yaml
```diff
@@ -0,0 +1,184 @@
...
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+---
+name: Sealights
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
simonbaird, line 22: I can it's different to the similar triggering config in .github/workflows/checks.yaml . Should we make them the same?

simonbaird: Why include labeled & unlabeled here?

ascerra: Good catch we're not using labels in this way so I should remove labeled , unlabeled.

About doing this the same for every workflow. It could be good for consistency but its not needed IMO.
- I'm just learning about most of this today but the reason we are doing it here is that this workflow relies on a secret and so once this workflow is merged we'll be able to use this in future PRs since we have the pull_request_target here instead of pull_request. It means we'll use the workflow and secret from the upstream repo not the incoming PR. This prevents contributors from editing the workflow to cat the secret or something.

So for the other workflows if they do not require a secret to be used then using pull_request as they already do should be fine.

simonbaird:
IIUC the brand new workflow won't be run until this is merged. I know it's been tested in @ascerra 's fork, but still there might be some extra changes needed to make sure it works in our repo. Is that right? Shall we merge it anyway and see how it goes?

PR #2486, opened 2025-04-22
Fix duplicate package import for gcr/v1

simonbaird:
I rebased the commit from #2482, which should get the Konflux checks green, and pushed a new PR.

### DIFF
--- a/internal/image/validate_test.go
+++ b/internal/image/validate_test.go
@@ -28,7 +28,6 @@ import (
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
-	gcr "github.com/google/go-containerregistry/pkg/v1"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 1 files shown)


simonbaird:
It's a trivial change and the CI is green, so I'll merge without further review.

simonbaird:
Actually I was just grepping the code, and everywhere else we use v1.Descriptor. So actually I want to make a revision to use that here also.

internal/image/validate_test.go
```diff
@@ -311,7 +310,7 @@ func TestEvaluatorLifecycle(t *testing.T) {
 	client.On("Head", mock.Anything).Return(&v1.Descriptor{MediaType: types.OCIManifestSchema1}, nil)
```
simonbaird, line 310: The one use of `v1.Descriptor`.

robnester-rh (approved):
LGTM

PR #2487, opened 2025-04-23
Remove the debug output from the tekton task

cuipinghuo:
Remove the debug output from the tekton task since it's too long and not that useful

Ref: https://issues.redhat.com/browse/EC-1206

### DIFF
--- a/features/__snapshots__/ta_task_validate_image.snap
+++ b/features/__snapshots__/ta_task_validate_image.snap
@@ -4,6 +4,29 @@
 }
 ---
 
+[Golden container image with trusted artifacts:show-config - 1]
+{
+  "policy": {
+    "sources": [
+      {
+        "policy": [
+          "git::github.com/enterprise-contract/ec-policies//policy/release?ref=d34eab36b23d43748e451004177ca144296bf323",
+          "git::github.com/enterprise-contract/ec-policies//policy/lib?ref=d34eab36b23d43748e451004177ca144296bf323"
+        ],
+        "config": {
+          "include": [
+            "slsa_provenance_available"
+          ]
+        }
+      }
+    ],
+    "publicKey": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAERhr8Zj4dZW67zucg8fDr11M4lmRp\nzN6SIcIjkvH39siYg1DkCoa2h2xMUZ10ecbM3/ECqvBV55YwQ2rcIEa7XQ==\n-----END PUBLIC KEY-----"
+  },
+  "key": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAERhr8Zj4dZW67zucg8fDr11M4lmRp\nzN6SIcIjkvH39siYg1DkCoa2h2xMUZ10ecbM3/ECqvBV55YwQ2rcIEa7XQ==\n-----END PUBLIC KEY-----\n",
+  "effective-time": "${TIMESTAMP}"
+}
+---
+
 [Golden container image with trusted artifacts:report-json - 1]
 {
   "success": true,
--- a/features/task_validate_image.feature
+++ b/features/task_validate_image.feature
@@ -297,27 +297,7 @@ Feature: Verify Enterprise Contract Tekton Tasks
       | IGNORE_REKOR         | true                                                                          |
       | EFFECTIVE_TIME       | 2020-01-01T00:00:00Z                                                          |
     Then the task should succeed
-      And the task logs for step "debug-log" should contain "Using provided effective time 2020-01-01T00:00:00Z"
-
-  # Previously we did allow a custom timeout to be set via the TIMEOUT param, but now it's ignored.
-  # (This test could be removed in the future, but let's keep it for now I guess.)
-  Scenario: Deprecated timeout param is ignored
-    Given a working namespace
-      And a key pair named "known"
-      And an image named "acceptance/timeout"
-      And a valid image signature of "acceptance/timeout" image signed by the "known" key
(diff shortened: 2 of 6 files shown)


tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
```diff
@@ -321,12 +320,6 @@ spec:
         - ".policy"
         - "$(params.HOMEDIR)/report-json.json"
```
simonbaird, line 322: I think you need to remove `"--verbose"` from above.

features/task_validate_image.feature
```diff
@@ -281,44 +281,6 @@ Feature: Verify Enterprise Contract Tekton Tasks
...
-      {"publicKey": ${known_PUBLIC_KEY}}
-      ```
-    When version 0.1 of the task named "verify-enterprise-contract" is run with parameters:
-      | IMAGES               | {"components": [{"containerImage": "${REGISTRY}/acceptance/timeout"}]} |
-      | POLICY_CONFIGURATION | ${NAMESPACE}/${POLICY_NAME}                                            |
-      | IGNORE_REKOR         | true                                                                   |
-      | TIMEOUT              | 666s                                                                   |
-    Then the task should succeed
-      And the task logs for step "debug-log" should contain "globalTimeout is 100h0m0s"
-
```
simonbaird, line 321: It's too bad we're removing this test coverage. I wonder if there's another way to check the globalTimeout param is working as expected.

cuipinghuo: Since globalTimeout is deprecated, we will remove the check of globalTimeout and check effectiveTime from report-json

cuipinghuo: check the effective-time from report-json.

features/task_validate_image.feature
```diff
@@ -297,27 +297,7 @@ Feature: Verify Enterprise Contract Tekton Tasks
...
-      {"publicKey": ${known_PUBLIC_KEY}}
-      ```
-    When version 0.1 of the task named "verify-enterprise-contract" is run with parameters:
-      | IMAGES               | {"components": [{"containerImage": "${REGISTRY}/acceptance/timeout"}]} |
-      | POLICY_CONFIGURATION | ${NAMESPACE}/${POLICY_NAME}                                            |
-      | IGNORE_REKOR         | true                                                                   |
-      | TIMEOUT              | 666s                                                                   |
-    Then the task should succeed
-      And the task logs for step "debug-log" should contain "globalTimeout is 100h0m0s"
+      And the task logs for step "report-json" should contain "effective-time":"2020-01-01T00:00:00Z"
```
simonbaird, line 300: I think it will work the same right:

```suggestion
      And the task logs for step "show-config" should contain "effective-time":"2020-01-01T00:00:00Z"
```

simonbaird:
Some commit message nitpicks:
* Should remove "WIP" from the first commit
* The second commit's subject line is too long. See [here](https://cbea.ms/git-commit/#seven-rules) for some good tips on commit messages.

Acepresso:
I've reviewed this PR and learned from it. Thank you

features/task_validate_image.feature
lcarva: Just to be clear, we're not deprecating the timeout of the CLI. We're only deprecating the timeout of the Tekton Task.

If someone is running the EC CLI directly (locally, for example), the timeout is still useful.

features/__snapshots__/ta_task_validate_image.snap
```diff
@@ -4,6 +4,29 @@
...
+          "include": [
+            "slsa_provenance_available"
+          ]
+        }
+      }
+    ],
+    "publicKey": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAERhr8Zj4dZW67zucg8fDr11M4lmRp\nzN6SIcIjkvH39siYg1DkCoa2h2xMUZ10ecbM3/ECqvBV55YwQ2rcIEa7XQ==\n-----END PUBLIC KEY-----"
+  },
+  "key": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAERhr8Zj4dZW67zucg8fDr11M4lmRp\nzN6SIcIjkvH39siYg1DkCoa2h2xMUZ10ecbM3/ECqvBV55YwQ2rcIEa7XQ==\n-----END PUBLIC KEY-----\n",
+  "effective-time": "${TIMESTAMP}"
```
simonbaird, line 26: It's showing the key twice, which might suggest we don't need the "key" here, but I think in the real cluster it might show `k8s://openshift-pipelines/public-key`.

simonbaird: I could be wrong.

features/task_validate_image.feature
```diff
@@ -297,27 +297,7 @@ Feature: Verify Enterprise Contract Tekton Tasks
...
-      ```
-      {"publicKey": ${known_PUBLIC_KEY}}
-      ```
-    When version 0.1 of the task named "verify-enterprise-contract" is run with parameters:
-      | IMAGES               | {"components": [{"containerImage": "${REGISTRY}/acceptance/timeout"}]} |
-      | POLICY_CONFIGURATION | ${NAMESPACE}/${POLICY_NAME}                                            |
-      | IGNORE_REKOR         | true                                                                   |
-      | TIMEOUT              | 666s                                                                   |
-    Then the task should succeed
-      And the task logs for step "debug-log" should contain "globalTimeout is 100h0m0s"
```
simonbaird, line 320: I was looking in the unit tests, and it seems there is no other coverage for this. One idea: We could create a Jira "Add test coverage for globalTimeout flag" and fix it later.

simonbaird (approved):

PR #2488, opened 2025-04-24
Update README with more troubleshooting cases for acceptance tests

Acepresso:
Add solutions to the troubleshooting section of the README to address the following acceptance test failures:

- Variety of issues - podman running as a system service
- Create keyring: Disk quota exceeded

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,29 @@ $ go env -w GOPROXY='https://proxy.golang.org,direct'
 
 #### **2. Docker Container Start Failures in Acceptance Tests**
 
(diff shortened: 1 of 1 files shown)


Acepresso:
I thought that it was worth mentioning the "running podman as a system service" even though it's written in the Docker documentation, since the resulting errors can be confusing. 

For example:

`unable to find network with name or ID bridge: network not found`

`docker: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock`

PR #2506, opened 2025-05-05
Properly handle error output when validating image

st3penta:
The 'ec validate image' command now handles error cases more organically, by collecting them and returning them as a list at the end of the execution, returning with exit status 1.
The previous implementation caused the command to exit with status 2 and to output the whole exception in the stderr channel.

Ref: https://issues.redhat.com/browse/EC-1020


st3penta:
The low coverage CodeCov is complaining about refers to some safeguards that are pretty tricky to test and should never happen under normal conditions.
I prefer to leave the coverage as it is and keep only the relevant unit tests instead.

simonbaird (approved):
Lgtm. I agree we can ignore the Codecov complaints.

joejstuart (approved):

PR #2507, opened 2025-05-05
Split report-json into more manageable line lengths

cuipinghuo:
split report-json.json at 8000 characters

improves readability by splitting the file when it exceeds 8000 characters.

Ref: https://issues.redhat.com/browse/EC-1207

### DIFF
--- a/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
+++ b/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
@@ -286,9 +286,15 @@ spec:
           memory: 256Mi
       image: quay.io/enterprise-contract/ec-cli:snapshot
       onError: continue  # progress even if the step fails so we can see the debug logs
-      command: [cat]
+      command: [sh, -c]
       args:
-        - "$(params.HOMEDIR)/report-json.json"
+        # Format the JSON output to wrap lines at 8000 characters per line.
+        # The report can get very large, so add some line breaks
+        # rather than print it as a single line. This makes it easier to render
+        # in the UI, easier to copy/paste, and less likely to cause problems
+        # with logging systems or other consumers of the data (assuming they
+        # correctly parse the full output).
+        - "jq . $(params.HOMEDIR)/report-json.json | awk '{gsub(/^ +/, \"\"); acc += length; if (acc >= 8000) { printf \"\\n\"; acc=length } printf $0 }'"
 
     - name: summary
       computeResources:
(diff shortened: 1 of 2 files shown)


tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -286,9 +286,10 @@ spec:
           memory: 256Mi
       image: quay.io/enterprise-contract/ec-cli:snapshot
       onError: continue  # progress even if the step fails so we can see the debug logs
-      command: [cat]
+      command: [sh, -c]
       args:
-        - "$(params.HOMEDIR)/report-json.json"
+        # Format the JSON output to wrap lines at 8000 characters for readability
+        - "jq . $(params.HOMEDIR)/report-json.json | awk '{gsub(/^ +/, \"\"); acc += length; if (acc >= 8000) { printf \"\n\"; acc=length } printf $0 }'"
```
simonbaird, line 292: Looking at the error in the acceptance test made be thing perhaps you have to escape the \ char in the \n. Not sure though...

```suggestion
        - "jq . $(params.HOMEDIR)/report-json.json | awk '{gsub(/^ +/, \"\"); acc += length; if (acc >= 8000) { printf \"\\n\"; acc=length } printf $0 }'"
```

tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml
```diff
@@ -286,9 +286,10 @@ spec:
           memory: 256Mi
       image: quay.io/enterprise-contract/ec-cli:snapshot
       onError: continue  # progress even if the step fails so we can see the debug logs
-      command: [cat]
+      command: [sh, -c]
       args:
-        - "$(params.HOMEDIR)/report-json.json"
+        # Format the JSON output to wrap lines at 8000 characters for readability
```
simonbaird, line 291: I'm trying to write a better explanation. It's not really readability, since you still can read it.

```suggestion
        # The report can get very large, so add some line breaks rather than
        # render as a single line. This makes it easier to render in the UI,
        # easier to copy/paste, and less likely to cause problems with logging
        # systems or other consumers of the data, (assuming they correctly parse
        # the full output).
```

simonbaird (approved):
Lgtm.

simonbaird:
I guess there's not much test coverage for a report that is longer than 8000 chars. Can you think of some way to test this before it hits production?

cuipinghuo:
> I guess there's not much test coverage for a report that is longer than 8000 chars. Can you think of some way to test this before it hits production?

we have https://github.com/enterprise-contract/ec-cli/blob/deac91a14964b9c8904363623b7850f5286dd697/features/task_validate_image.feature#L241-L263

simonbaird:
About the testing as discussed it in slack: Summary: Never mind trying to write an acceptance test with a long wrapped report.json, but let's just check manually it works well in a real Konflux UI to improve confidence about this change. (This was done, see details [here](https://redhat-internal.slack.com/archives/C065E0PFN77/p1746553257035799).)

PR #2510, opened 2025-05-07
Remove base branches

joejstuart:
Having the multiple base branches is suspected
of causing OOM errors with Mintmaker due to the
size of the updates.

https://issues.redhat.com/browse/EC-1244

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,6 +2,5 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "github>conforma/.github//config/renovate/renovate.json"
-  ],
-  "baseBranches": ["main", "release-v0.5", "release-v0.6"]
+  ]
 }


simonbaird (approved):
Would be nice to add comment. Too bad we don't have json5 here.

renovate.json
```diff
@@ -2,6 +2,5 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "github>conforma/.github//config/renovate/renovate.json"
-  ],
-  "baseBranches": ["main", "release-v0.5", "release-v0.6"]
+  ]
```
simonbaird, line 5: I expect it would work the same, but consider this:

```suggestion
    "baseBranches": ["main"]
  ]
```

joejstuart: From support it sounds like we shouldn't have any base branches.

PR #2512, opened 2025-05-08
Updating conditional value to sealights workflow

ascerra:
We saw the first true run of the new Sealights workflow skip the configuration step due to the conditional looking for `pull_request` instead of `pull_request_target` which was changed during the review process.

This PR updates the conditional to be `pull_request_target`


simonbaird:
[Related docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target).

Not sure I understand the nuances, but lgtm.

simonbaird (approved):

PR #2535, opened 2025-05-15
Revert "Workaround for release pipeline tekton task"

simonbaird:
See PR #2513. I don't think we'll need this, but just in case it does break something, and we decide to quickly revert, here's a PR to do it.

Related slack thread:
https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1746425126665539

This reverts commit a5ba155a781386f6de60e332fef12cbac4b884cf.


simonbaird:
Nothing bad happened, so I guess we're keeping it. Abandoning this "just-in-case" revert PR.

PR #2541, opened 2025-05-19
chore(deps): Update Konflux bundle references & some golang dependencies (v0.5)

simonbaird:
Similar to #2540.


simonbaird:
I think the acceptance test works for me locally, so I'll try re-running.

PR #2544, opened 2025-05-20
Update ubi9 and go toolset (v06)

simonbaird:
Similar to #2543 but without the changes to the rpm lock file (since we don't install git in this branch).

Includes a switch to :latest tag for the base image, and a switch of go-toolset image. See commit messages for more details.

Ref: [EC-1282](https://issues.redhat.com/browse/EC-1282)


simonbaird:
Same thing as #2543 so no need for re-review.

PR #2549, opened 2025-05-27
Issue warning for unmatched include directives

st3penta:
When the 'include' directive in a policy configuration references a rule or collection that does not exist in the specified policy sources, the `ec validate image` command now issues a warning instead of silently ignoring the entry.

This addresses a bug where users could mistakenly believe a misconfigured policy was being fully applied, leading to misleading validation results.

Ref: https://issues.redhat.com/browse/EC-729

### DIFF
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -463,6 +463,20 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, target EvaluationTarget
 	// at all was processed.
 	totalRules := 0
 
+	// Populate a list with all the include directives specified in the
+	// policy config.
+	// Each include matching a result will be pruned from the list, so
+	// that in the end the list will contain all the unmatched includes.
+	missingIncludes := map[string]bool{}
+	for _, defaultItem := range c.include.defaultItems {
+		missingIncludes[defaultItem] = true
+	}
+	for _, digestItems := range c.include.digestItems {
+		for _, digestItem := range digestItems {
+			missingIncludes[digestItem] = true
+		}
+	}
+
 	// loop over each policy (namespace) evaluation
 	// effectively replacing the results returned from conftest
 	for i, result := range runResults {
@@ -476,7 +490,7 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, target EvaluationTarget
 			warning := result.Warnings[i]
 			addRuleMetadata(ctx, &warning, rules)
 
-			if !c.isResultIncluded(warning, target.Target) {
+			if !c.isResultIncluded(warning, target.Target, missingIncludes) {
 				log.Debugf("Skipping result warning: %#v", warning)
 				continue
 			}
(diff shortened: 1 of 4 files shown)


internal/evaluator/conftest_evaluator_test.go
```diff
@@ -1282,6 +1282,33 @@ func TestConftestEvaluatorIncludeExclude(t *testing.T) {
...
+			name: "warning for missing includes",
+			results: []Outcome{
+				{
+					Failures: []Result{
+						{Metadata: map[string]any{"code": "lunch.spam"}},
+						{Metadata: map[string]any{"code": "breakfast.spam", "term": []any{"bacon", "nocab"}}},
+					},
+				},
+			},
+			config: &ecc.EnterpriseContractPolicyConfiguration{Include: []string{"breakfast:eggs", "lunch"}},
```
simonbaird, line 1295: A nitpick:

Here "eggs" is a "term". Most of the time, we don't use a term in an include. (I just learned there is test coverage for doing that, but I don't think it's ever used in real life situation). The term is most often used for an exclude to do some fine-grained control of when the rule can be excluded.

It won't impact the test behavior, but I think it's more understandable if we make the missing include a rule, not a rule package and a term. E.g. something like this:

```suggestion
			config: &ecc.EnterpriseContractPolicyConfiguration{Include: []string{"breakfast.pancakes", "lunch"}},
```

internal/evaluator/conftest_evaluator_test.go
```diff
@@ -1282,6 +1282,33 @@ func TestConftestEvaluatorIncludeExclude(t *testing.T) {
...
+					Warnings:   []Result{},
+					Exceptions: []Result{},
+					Failures: []Result{
+						{Metadata: map[string]any{"code": "lunch.spam"}},
+					},
+				},
+				{
+					Warnings: []Result{
+						{Message: "Include criterion 'breakfast:eggs' doesn't match any policy rule"},
+					},
```
simonbaird, line 1308: Looks like we're producing a list of length 2 in `[]Outcome`, but all the other tests here have just single outcome in `[]Outcome`. Is that okay? I was wondering if the `Result` here could or should be merged into the `Warnings` field on line 1299.

(I'm not sure though. I might try running this with a real example to see what output gets produced.)

simonbaird: Based on the acceptance test snapshot, seems like it all gets collated together, so perhaps this is fine.

internal/evaluator/conftest_evaluator.go
```diff
@@ -844,20 +869,21 @@ func isResultEffective(failure Result, now time.Time) bool {
...
 
 // scoreMatches returns the combined score for every match between needles and haystack.
-func scoreMatches(needles, haystack []string) int {
+func scoreMatches(needles, haystack []string, toBePruned map[string]bool) int {
 	var s int
 	for _, needle := range needles {
 		for _, hay := range haystack {
 			if hay == needle {
 				s += score(hay)
+				delete(toBePruned, hay)
```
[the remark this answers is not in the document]
st3penta: Yes you're right, i added some comments to clarify the logic

simonbaird (approved):
Lgtm.

PR #2557, opened 2025-06-05
Add unit test for variable globalTimeout

cuipinghuo:
ref: https://issues.redhat.com/browse/EC-1237


cuipinghuo:
We have [a discussion](https://redhat-internal.slack.com/archives/C065E0PFN77/p1747754562155039) whether should as a unit test or acceptance test for the coverage of globalTimeout. I mislead @simonbaird about with [this](https://github.com/enterprise-contract/ec-cli/blob/fb12315e58b1ab9a5f39eb5be0ff92a1c352d05a/cmd/root/root_cmd.go#L81-L86). It is about how globalTimeout is set and logged not an unit test. I'm back to original ideas to add unit test.

simonbaird (approved):
Lgtm.

I think it will help our total coverage numbers go up in Sealights and Codecov, which is nice also.

PR #2571, opened 2025-06-10
Generate VSA for each validated image

joejstuart:
- Generates a predicate for each image
- Cli args are created to toggle vsa creation
- VSA upload location argument added

https://issues.redhat.com/browse/EC-1307

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -32,15 +32,15 @@ import (
 	"github.com/spf13/cobra"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 
-	"github.com/enterprise-contract/ec-cli/internal/applicationsnapshot"
-	"github.com/enterprise-contract/ec-cli/internal/evaluator"
-	"github.com/enterprise-contract/ec-cli/internal/format"
-	"github.com/enterprise-contract/ec-cli/internal/output"
-	"github.com/enterprise-contract/ec-cli/internal/policy"
-	"github.com/enterprise-contract/ec-cli/internal/policy/source"
-	"github.com/enterprise-contract/ec-cli/internal/utils"
-	validate_utils "github.com/enterprise-contract/ec-cli/internal/validate"
-	"github.com/enterprise-contract/ec-cli/internal/validate/vsa"
+	"github.com/conforma/cli/internal/applicationsnapshot"
+	"github.com/conforma/cli/internal/evaluator"
+	"github.com/conforma/cli/internal/format"
+	"github.com/conforma/cli/internal/output"
+	"github.com/conforma/cli/internal/policy"
+	"github.com/conforma/cli/internal/policy/source"
+	"github.com/conforma/cli/internal/utils"
+	validate_utils "github.com/conforma/cli/internal/validate"
+	"github.com/conforma/cli/internal/validate/vsa"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/validate/image.go
```diff
@@ -453,6 +457,61 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
...
+							continue
+						}
+						uploadResult, err := uploader(cmd.Context(), att, comp.ContainerImage)
+						if err != nil {
+							log.Errorf("[VSA] Failed to upload VSA attestation for image %s: %v", comp.ContainerImage, err)
+							continue
+						}
+						log.Infof("[VSA] VSA attestation uploaded for %s: %s", comp.ContainerImage, uploadResult)
+					}
+					// Rekor upload is skipped for now
```
st3penta, line 512: is it correct that the upload is inside the `if` block of the signing?
If so, the rekor upload (currently skipped) outside of the block seems redundant, no?

cmd/validate/image.go
```diff
@@ -453,6 +457,61 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
...
+					vsaPath := fmt.Sprintf("%s.vsa.json", comp.ContainerImage) // TODO: sanitize filename
+					log.Debugf("[VSA] Writing VSA to %s", vsaPath)
+					err = vsa.WriteVSA(pred, vsaPath)
+					if err != nil {
+						log.Errorf("[VSA] Failed to write VSA for image %s: %v", comp.ContainerImage, err)
+						continue
+					}
+					log.Debugf("[VSA] VSA written to %s", vsaPath)
+
+					if data.vsaSigningKey != "" {
```
st3penta, line 485: consider using an early return here
```suggestion
					if data.vsaSigningKey == "" {
						log.Debugf("[VSA] No VSA signing key provided, skipping signing and upload")
						continue
					}
```

cmd/validate/image.go
```diff
@@ -453,6 +457,61 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 				return err
 			}
 
+			if data.vsaEnabled {
```
st3penta, line 460: this is already a long function, i'd move this in its own function for readability 
```suggestion
			if data.vsaEnabled {
				generateVSA(...)
			}
```

internal/validate/vsa/vsa.go
```diff
@@ -0,0 +1,166 @@
...
+	ruleResults = append(ruleResults, component.Successes...)
+
+	validationResult := "failed"
+	if component.Success {
+		validationResult = "passed"
+	}
+
+	policySource := ""
+	if report.Policy.PublicKey != "" {
+		policySource = report.Policy.Name
```
st3penta, line 84: why is the Name extracted only if there is a PublicKey? Maybe a comment would be useful here

st3penta (commented):
The logic seems ok to me, I just left some comments about the style of the code.

Also, there are a few TODOs here and there, i guess they will be addressed in later PRs?

cmd/validate/image.go
joejstuart: It is. Just a leftover comment. I removed it.

internal/validate/vsa/vsa.go
joejstuart: Good catch. `policySource` initially stored the `PublicKey`, but I later changed it to the `Name`.

st3penta (approved):

PR #2588, opened 2025-06-16
Remove the WORKERS parameter

arewm:
This task is being used by the release service which overrides the WORKERS parameter. While this change can be made there, it will take time for the parameter modification to make it to production.

In order to unblock some users with large Snapshots, I am proposing to remove the configuration altogether. We already increased the default resource requirements for the task, so this worker count should suffice. If this number of workers is too many in the future, users can just override the resources to further increase them.

Once the release pipeline users can customize the worker count in the RPA (https://issues.redhat.com/browse/KONFLUX-7850), we can re-enable this parameter.


davidmogar:
I would probably just hardcode the flag value but leave the parameter. This parameter will be used in the future when https://issues.redhat.com/browse/KONFLUX-7850 gets implemented.

Independently of this, I support ignoring the value we (release) send. It would be way faster than if we remove the parameter from our tasks.

simonbaird (approved):
If the plan is to bring the param back later, then maybe it is sensible to leave the param in, but ignore it.

But I don't mind too much. It's not a blocker.

tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
```diff
@@ -248,7 +243,7 @@ spec:
         - "$(params.REKOR_HOST)"
         - "--ignore-rekor=$(params.IGNORE_REKOR)"
         - "--workers"
-        - "$(params.WORKERS)"
+        - "35"
```
simonbaird, line 246: I know the commit message explains it well, but even so:

```suggestion
        # Temporarily hard code this instead of using the WORKERS param
        - "35"
```

simonbaird (approved):

arewm:
I just added the parameter back in with two notes.

PR #2592, opened 2025-06-17
Make tenant release push to old and new quay orgs

simonbaird:
To make this work some Konflux secrets needed to be updated. See details in Jira.

Ref: https://issues.redhat.com/browse/EC-1331

### DIFF
--- a/release/cli.yaml
+++ b/release/cli.yaml
@@ -22,11 +22,15 @@ metadata:
 spec:
   description: Tekton pipeline to release Snapshots to a registry.
   params:
-  - default: quay.io/enterprise-contract/cli
+  - default: quay.io/conforma/cli
+    # ...but see below for the additional hardcoded push to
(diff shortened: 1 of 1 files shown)


simonbaird:
I think I figured out which Konflux secrets I'd need to tweak to make it work.

release/cli.yaml
```diff
@@ -185,6 +189,26 @@ spec:
         - $(params.bundle-target-repo)
         image: quay.io/konflux-ci/appstudio-utils:latest
         name: copy-bundle
+      # To ease the transtion from the old quay org to the new quay org,
+      # also push the two images to the old quay org at. Rather than
```
simonbaird, line 193: It's a typo, but I don't want to redo the tests.

```suggestion
      # also push the two images to the old quay org. Rather than
```

PR #2606, opened 2025-06-26
document rule filtering process

joejstuart:


simonbaird:
A few questions:
* What is the motivation?
* Why the mismatched branch name?
* What is an `.mdc` file? Should be be `.md`?
* Are these docs for humans or robots or both?

joejstuart:
> A few questions:
> 
> * What is the motivation?
> * Why the mismatched branch name?
> * What is an `.mdc` file? Should be be `.md`?
> * Are these docs for humans or robots or both?

I'm working on returning all of the results and realized I have no idea how the filtering works, so I asked Cursor. The `.mdc` is a Cursor markdown file. They are for humans and robots. You can refer to this in Cursor if you're doing any work on rule filtering and it will give it a better understanding of the code. Sorry for the lack of description. I just did this on a whim. If it's accurate and we want to start using it ( I think we should ) I can amend the commit.

st3penta:
Cool! I am in favor of adding this to the project :+1: 

The contents seem correct to me.

PR #2607, opened 2025-06-26
Use Konflux built image refs for tekton task & cli image

simonbaird:
Note that the old repos are still being updated, and still work fine, it's just that they're built and pushed using GitHub workflows instead of using Konflux pipelines. The Konflux built images are preferred, so let's use them consistently.

Ref: https://issues.redhat.com/browse/EC-1255


st3penta (approved):
lgtm

PR #2625, opened 2025-07-09
Tag fix for infra deployment update

simonbaird:
Since the `{{sha}}` tag isn't being created any more for the tekton task bundle, the script was unable to find it and crashing. Use the new `kf-{{sha}}` tag instead.

Ref: https://issues.redhat.com/browse/EC-1372


robnester-rh (approved):
Lgtm

PR #2626, opened 2025-07-09
Remove gh- tag prefix temporarily

simonbaird:
There's a regex bug causing major chaos to the $TARGETOS value in the Makefile because of this prefix.

As far as I know the kf- prefix isn't causing problems, so let's keep that one.

Ref: https://issues.redhat.com/browse/EC-1373


simonbaird:
If it goes green I might merge it.

PR #2630, opened 2025-07-14
Remove "effective_on" from trusted_tasks

dheerajodha:
resolves: EC-423


internal/tracker/tracker.go
```diff
@@ -314,20 +302,23 @@ func filterRecords(records []taskRecord, prune bool) []taskRecord {
...
-// effective_on value of the n-1 record. The first record on the list does not contain an expires_on
-// value since there is no newer record that invalidates it in the future.
-// TODO: Probably need to compute the expires_on value without requiring the "effective_on" value so
-// we don't have to always require both values. But this may be required during some transition
-// period.
-func (t *Tracker) setExpiration() {
+// setExpiration sets the expires_on attribute on records using a duration-based approach.
+// Each record expires after a configurable duration based on inEffectDays from when it's added.
+// The most recent record (index 0) gets no expiration date, making it the current active record.
+func (t *Tracker) setExpiration(inEffectDays int) {
```
dheerajodha, line 308: As explained above, the new way of adding the expires_on field is to set the same value, taken from the `--in-effect-days` flag, to all the older records except the latest one.

dheerajodha:
>  The acceptance timeout should be fixed if you give it a rebase.

Thanks! Now the current failure is due to my changes. Pushing a new commit, apologies.

cmd/track/track_bundle.go
```diff
@@ -62,14 +62,14 @@ func trackBundleCmd(track trackBundleFn, pullImage pullImageFn, pushImage pushIm
...
-			the most recent effective_on date *not* in the future are considered
-			acceptable.
+			most recent Tekton Bundle is used. Each entry contains an "expires_on"
+			date which indicates when that specific bundle version should no longer
+			be used. The most recent bundle has no expiration date until it is
+			replaced by a newer version.
+
+			If --prune is set, on by default, expired entries are removed.
+			Any entry with an expires_on date in the future (or no expires_on date)
+			is considered acceptable and will be kept.
```
simonbaird, line 72:
```suggestion
			is considered current and will be kept.
```

cmd/track/track_bundle.go
```diff
@@ -181,7 +181,7 @@ func trackBundleCmd(track trackBundleFn, pullImage pullImageFn, pushImage pushIm
 
 	cmd.Flags().BoolVar(&params.freshen, "freshen", params.freshen, "resolve image tags to catch updates and use the latest image for the tag")
 
-	cmd.Flags().IntVar(&params.inEffectDays, "in-effect-days", params.inEffectDays, "number of days representing when the added reference becomes effective")
+	cmd.Flags().IntVar(&params.inEffectDays, "in-effect-days", params.inEffectDays, "number of days after which older bundle entries expire (most recent entry stays valid until replaced)")
```
simonbaird, line 184:
```suggestion
	cmd.Flags().IntVar(&params.inEffectDays, "in-effect-days", params.inEffectDays, "number of days after which older bundle entries expire when a new bundle entry is added. (most recent entry stays valid until replaced)")
```

simonbaird: maybe..

cmd/track/track_bundle_test.go
```diff
@@ -204,6 +205,19 @@ func Test_TrackBundleCommand(t *testing.T) {
...
+				"--bundle",
+				"registry/image:tag",
+				"--in-effect-days",
+				"100",
+			},
+			expectUrls:         []string{"registry/image:tag"},
+			expectStdout:       true,
+			expectPrune:        true,
+			expectInEffectDays: 100,
+		},
```
simonbaird, line 220: I don't understand what this test adds that isn't tested in the previous (line 196) test case. Remove it maybe?

internal/tracker/tracker.go
```diff
@@ -111,20 +110,20 @@ func Track(ctx context.Context, urls []string, input []byte, prune bool, freshen
 
 	imageUrls, gitUrls := groupUrls(urls)
 
-	days := oneDay * time.Duration(inEffectDays)
-	effectiveOn := time.Now().Add(days).UTC().Round(oneDay)
+	// Note: inEffectDays parameter is kept for backward compatibility but no longer used
```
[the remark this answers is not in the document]
simonbaird: I would think the parameter is still used to decide what the expires_on values should be. Is that correct?

internal/tracker/tracker.go
```diff
@@ -111,20 +110,20 @@ func Track(ctx context.Context, urls []string, input []byte, prune bool, freshen
...
+	// Records now use creation timestamps instead of future effective dates
```
simonbaird, line 114: I don't follow this comment. What do you mean by creation timestamps?

internal/tracker/tracker.go
```diff
@@ -263,9 +260,7 @@ func (t *Tracker) filterBundles(prune bool) {
 
 // filterRecords reduces the list of records by removing superfluous entries.
 // It removes records that have the same reference in a certain group. If prune is
-// true, it skips any record that is no longer acceptable. Any record with an
-// EffectiveOn date in the future, and the record with the most recent
-// EffectiveOn date *not* in the future are considered acceptable.
+// true, it removes records that have already expired based on their expires_on date.
```
simonbaird, line 263: :+1:

internal/tracker/tracker.go
```diff
@@ -314,20 +302,23 @@ func filterRecords(records []taskRecord, prune bool) []taskRecord {
...
-// TODO: Probably need to compute the expires_on value without requiring the "effective_on" value so
-// we don't have to always require both values. But this may be required during some transition
-// period.
-func (t *Tracker) setExpiration() {
+// setExpiration sets the expires_on attribute on records using a duration-based approach.
+// Each record expires after a configurable duration based on inEffectDays from when it's added.
+// The most recent record (index 0) gets no expiration date, making it the current active record.
+func (t *Tracker) setExpiration(inEffectDays int) {
+	expirationDuration := time.Duration(inEffectDays) * oneDay // Use --in-effect-days flag value
+	now := time.Now().UTC().Round(oneDay)
```
simonbaird, line 310: Maybe this is nicer:

```suggestion
	expiresOn := time.Now().UTC().Add(expirationDuration)
```

then use it on line 319.

Ps, what is the idea behine the `.Round(oneDay)`? What does that do actually?

internal/tracker/tracker.go
```diff
@@ -290,17 +285,10 @@ func filterRecords(records []taskRecord, prune bool) []taskRecord {
...
-			if skip {
-				continue
-			}
-			relevant = append(relevant, r)
-			if !skip {
-				if now.After(r.EffectiveOn) {
-					skip = true
-				}
+			// Keep records that haven't expired yet, or records with no expiration date
+			if r.ExpiresOn == nil || now.Before(*r.ExpiresOn) {
```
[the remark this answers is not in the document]
dheerajodha: Yes, exactly, without the `*` we get a compile-time error:
```
cannot use r.ExpiresOn (variable of type *time.Time) as time.Time value in argument to now.Before
```
as the `ExpiresOn` field was initially defined as a pointer so that it gets omitted if there isn't a value ([ref](https://github.com/conforma/cli/blob/main/internal/tracker/tracker.go#L39-L40))

internal/tracker/tracker.go
dheerajodha: This makes sense now, thank you!

cmd/track/track_bundle_test.go
```diff
@@ -269,6 +270,33 @@ func Test_TrackBundleCommand(t *testing.T) {
...
+
+	// Verify the in-effect-days flag is still documented (backward compatibility)
+	foundFlag := false
+	trackBundleCmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		if flag.Name == "in-effect-days" {
+			foundFlag = true
+			assert.Equal(t, "30", flag.DefValue, "Default value should be 30")
+		}
+	})
+	assert.True(t, foundFlag, "in-effect-days flag should still exist for backward compatibility")
```
simonbaird, line 297: I feel like we can't talk about "backward compatibility" unless we we created a new flag and then deprecate the old one. Unless I'm missing something, we didn't do that, right?

cmd/track/track_bundle_test.go
```diff
@@ -269,6 +270,33 @@ func Test_TrackBundleCommand(t *testing.T) {
...
+
+	// Verify it explains the new behavior
+	assert.Contains(t, trackBundleCmd.Long, "expiration date is added",
+		"Command help should explain that expiration dates are added")
+
+	// Verify pruning explanation is updated
+	assert.Contains(t, trackBundleCmd.Long, "expired entries are removed",
+		"Command help should explain pruning removes expired entries")
+
+	// Verify the in-effect-days flag is still documented (backward compatibility)
```
simonbaird, line 289:
```suggestion
	// Verify the in-effect-days flag is documented```

simonbaird:
Nice patch!

PR #2632, opened 2025-07-14
Update golang/glog (v06)

simonbaird:
Includes some other dependency bumps, but the main motivation is to upgrade `github.com/golang/golog`.

Note that it was only tools/go.mod that had the older version with the vulnerability, but I updated to the latest in the main mod file anyhow.

Ref: https://issues.redhat.com/browse/EC-1385 .


simonbaird:
Seems like a real failure in the validate_input feature, rather than a timeout.

Let's retry and see if it happens again.

PR #2636, opened 2025-07-16
Upgrade tektoncd/pipeline to fix acceptance tests

simonbaird:
The acceptance tests are failing due to a timeout, which seems to be related to installing Tekton in the local Kind cluster.

The reason why is that that older Tekton "release.yaml" files use Google's gcr.io registry for the images, and that registry is no longer accepting anonymous pulls.

Updating to a newer Tekton version should mean we install from a newer "release.yaml" file which uses GitHub's ghcr.io registry for the images. That registry does still allow anonymous pulls, so should be able to install Tekton successfully, hence the acceptance tests will be able to get a working Kind cluster.

I chose to use the version of Tekton that is already used in acceptance/go.mod rather than try to use the latest tekton dependency (which is now up to 1.2.0). Not sure how or why the go.mod and the acceptance/go.mod versions diverged, but making them the same seems sensible.

(Another way to fix this might have been to change what directory the `go list -f '{{.Version}}' -m github.com/tektoncd/pipeline` was run in to ./acceptance instead of ./., but I figure it's better to keep the go module versions in sync anyhow.)

I did confirm that 0.66.0 does have ghcr.io in the its release.yaml and not gcr.io.

This change had some success getting PRs to pass in release-v0.5 and release-v0.6 branch already, so hopefully main branch will be green again very soon.

See also:
* https://tekton.dev/blog/2025/04/03/migration-to-github-container-registry/
* https://github.com/tektoncd/pipeline/releases
* hack/tekton/tekton.sh and hack/tekton/tekton/kustomization.yaml
* https://github.com/conforma/cli/pull/2632 and https://github.com/conforma/cli/pull/2631

Ref: https://issues.redhat.com/browse/EC-1387


tools/go.sum
```diff
@@ -823,8 +823,8 @@ github.com/golang-jwt/jwt/v4 v4.5.2/go.mod h1:m21LjoU+eqJr34lmDMbreY2eSTRJ1cv77w
 github.com/golang-jwt/jwt/v5 v5.2.2 h1:Rl4B7itRWVtYIHFrSNd7vhTiz9UpLdi6gZhZ3wEeDy8=
 github.com/golang-jwt/jwt/v5 v5.2.2/go.mod h1:pqrtFR0X4osieyHYxtmOUWsAWrfe1Q5UVIyoH402zdk=
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b/go.mod h1:SBH7ygxi8pfUlaOkMMuAQtPIUF8ecWP5IEl/CR7VP2Q=
-github.com/golang/glog v1.2.1 h1:OptwRhECazUx5ix5TTWC3EZhsZEHWcYWY4FQHTIubm4=
-github.com/golang/glog v1.2.1/go.mod h1:6AhwSGph0fcJtXVM/PEHPqZlFeoLxhs7/t5UDAwmO+w=
+github.com/golang/glog v1.2.2 h1:1+mZ9upx1Dh6FmUTFR1naJ77miKiXgALjWOZ3NVFPmY=
```
cuipinghuo, line 826: https://issues.redhat.com/browse/EC-1385 mentioned to have golang/glog 1.2.4 or higher. I saw v05 and v6 use 1.2.5. Do we also upgrade for main branch?

simonbaird:
Trying with the acceptance tests moved to their old location in the non-sealights workflow.

simonbaird:
Replaced by #2637 .

PR #2637, opened 2025-07-16
Fix acceptance test timeout

simonbaird:
See commits for explanation.

Ref: https://issues.redhat.com/browse/EC-1387


simonbaird:
To get this green I temporarily removed "Acceptance" from the required test in the ruleset at https://github.com/conforma/cli/settings/rules/139585 .

Will put it right back afterwards.

PR #2639, opened 2025-07-16
 implement pluggable rule filtering system with pipeline intentions

joejstuart:
feat: implement pluggable rule filtering system with pipeline intentions

- Add extensible filtering framework with RuleFilter interface
- Implement PipelineIntentionFilter for filtering rules by pipeline intentions
- Implement IncludeListFilter for filtering by collections, packages, and rules
- Add pipeline_intention metadata field to rule annotations
- Fix filtering leak by ensuring AllNamespaces=false prevents bypassing filters
- Add comprehensive test coverage for all filtering components
- Maintain backward compatibility with existing filtering logic

The new system allows for flexible rule filtering based on pipeline intentions
and include lists, while preventing any evaluation of excluded namespaces
regardless of filtering results.

Assisted by: Cursor (powered by Claude 4 Sonnet)

### DIFF
--- a/internal/evaluator/conftest_evaluator.go
+++ b/internal/evaluator/conftest_evaluator.go
@@ -167,17 +167,18 @@ type testRunner interface {
 }
 
 const (
-	effectiveOnFormat   = "2006-01-02T15:04:05Z"
-	effectiveOnTimeout  = -90 * 24 * time.Hour // keep effective_on metadata up to 90 days
-	metadataCode        = "code"
-	metadataCollections = "collections"
-	metadataDependsOn   = "depends_on"
-	metadataDescription = "description"
-	metadataSeverity    = "severity"
-	metadataEffectiveOn = "effective_on"
-	metadataSolution    = "solution"
-	metadataTerm        = "term"
-	metadataTitle       = "title"
+	effectiveOnFormat         = "2006-01-02T15:04:05Z"
+	effectiveOnTimeout        = -90 * 24 * time.Hour // keep effective_on metadata up to 90 days
+	metadataCode              = "code"
+	metadataCollections       = "collections"
+	metadataDependsOn         = "depends_on"
+	metadataDescription       = "description"
+	metadataPipelineIntention = "pipeline_intention"
+	metadataSeverity          = "severity"
+	metadataEffectiveOn       = "effective_on"
+	metadataSolution          = "solution"
+	metadataTerm              = "term"
+	metadataTitle             = "title"
 )
 
 const (
@@ -205,6 +206,7 @@ type conftestEvaluator struct {
 	exclude       *Criteria
 	fs            afero.Fs
 	namespace     []string
+	source        ecc.Source
 }
 
 type conftestRunner struct {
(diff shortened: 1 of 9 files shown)


joejstuart:
I have a fix coming for the acceptance test errors.

internal/evaluator/conftest_evaluator.go
```diff
@@ -414,32 +423,86 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, target EvaluationTarget
...
 	var r testRunner
 	var ok bool
 	if r, ok = ctx.Value(runnerKey).(testRunner); r == nil || !ok {
 
-		// should there be a namespace defined or not
-		allNamespaces := true
-		if len(c.namespace) > 0 {
-			allNamespaces = false
+		// Determine which namespaces to use
+		namespaceToUse := c.namespace
```
simonbaird, line 489:
```suggestion
		namespacesToUse := c.namespace
```

internal/evaluator/filters.go
```diff
@@ -0,0 +1,322 @@
...
+//   - Supports @collection, package names, and package.rule patterns
+//
+// 3. Combined Logic:
+//   - All filters are applied with AND logic - a package must pass ALL filters
+//   - This allows fine-grained control over which rules are evaluated
+func (f *DefaultFilterFactory) CreateFilters(source ecc.Source) []RuleFilter {
+	var filters []RuleFilter
+
+	// ── 1. Pipeline‑intention ───────────────────────────────────────────────
+	intentions := extractStringArrayFromRuleData(source, "pipeline_intention")
```
simonbaird, line 83: Uhoh, this might be a problem.

The way we typically set the pipeline_intention is with a command line flag like this: ``ec validate image .... --extra-rule-data pipeline_intention=production`. It's not generally present in the ECP.

This code makes me think it will try to find it in the ECP directly.

Maybe we have to think of another way to detect the pipeline_indention. One quick idea (not sure if practiccal): `opa eval 'data.lib.rule_data("pipeline_intention")'` will display it.

simonbaird: Another idea: We could do some creating interpretation of the `--extra-rule-data pipeline_intention=production` flag and set the value to var somewhere. That actually has some benefits in that we're moving away from the need for pipeline_intention being in the data at all.

joejstuart:
> High level suggestions/comments/thoughts:
> 
> * I think I like the filtering mechanism, and it feels like it's useful no matter what. However I wonder if we could add it with just the include filtering in a nice tidy PR. Then introduce the pipeline_intention filtering support separately.
> * The more I think about it, the more I wonder why we need this new special "pipeline_intention" filter. What if we put all the pipeline intention rules in a collection called "redhat-release", then just included them by collection? It might require some extra flags to do `--include-only "@redhat-release"` on the command line, and have that override or merge with the include from the ECP.

We can look at including it as a collection, but it takes on multiple values (release, production, staging). I noticed it's being used in ITS's with `pipeline_intention=staging`. We'd probably want to investigate more if we change how this works so we don't break anyone.

simonbaird:
> We can look at including it as a collection, but it takes on multiple values
Yeah good point. It think let's go ahead with this design and maybe consider other options long term.

simonbaird:
I had one or two minor suggestions, but overall lgtm.

PR #2646, opened 2025-07-21
Add comprehensive test suite for policy cache implementation

cuipinghuo:
This commit introduces thorough test coverage for the policy cache package, including unit tests for all major operations (Get, Set, cache creation), context integration tests, concurrent access validation, and type safety verification. The test suite ensures reliable cache behavior across various edge cases and concurrent scenarios, providing confidence in the cache's thread-safety and correctness.

Function TestGet_ConcurrentAccess tests performance and thread-safety under concurrent load, while TestGet_TypeSafety covers type safety and error handling when invalid types are stored.

Assisted-by: Cursor (powered by granite-3-2-8b-instruct)

Ref: https://issues.redhat.com/browse/EC-1364


internal/policy/cache/cache_test.go
```diff
@@ -18,50 +18,508 @@ package cache
...
+
+			// Verify the value was set correctly
+			value, found := testCache.Get(tt.key)
+			assert.Equal(t, tt.expectedValue, value)
+			assert.Equal(t, tt.expectedFound, found)
+		})
+	}
 }
+
+func TestCreatePolicyCache(t *testing.T) {
```
st3penta, line 191: I think the `CreatePolicyCache` function being tested here could be merged in `NewPolicyCache`, resulting in a simpler constructor ([here](https://github.com/conforma/cli/blob/main/internal/policy/cache/cache.go#L75-L94))

internal/policy/cache/cache_test.go
```diff
@@ -18,50 +18,508 @@ package cache
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := tt.setupContext()
+
+			// Execute test
+			cache, err := NewPolicyCache(ctx)
+
+			// Assert results
+			if tt.expectError {
+				assert.Error(t, err)
```
st3penta, line 269:
```suggestion
			if tt.expectError {
				assert.EqualError(t, err, tt.expectedError)
			}
```

internal/policy/cache/cache_test.go
```diff
@@ -18,50 +18,508 @@ package cache
...
+			}
+		})
+	}
+}
+
+func TestNewPolicyCache(t *testing.T) {
+	tests := []struct {
+		name           string
+		setupContext   func() context.Context
+		expectError    bool
```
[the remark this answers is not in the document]
cuipinghuo: since I simplified NewPolicyCache(ctx context.Context) *PolicyCache and the function cannot return errors, so I remove the expectError field  from function TestCreatePolicyCache and TestNewPolicyCache

cuipinghuo:
> Side note on the tested class `PolicyCache`: i don't understand what is the purpose of setting an error in the cache entry ([here](https://github.com/conforma/cli/blob/main/internal/policy/cache/cache.go#L69)), if it gets ignored when the entry is retrieved ([here](https://github.com/conforma/cli/blob/main/internal/policy/cache/cache.go#L61)). Perhaps the Set method could be simplified?

 Made changes in TestSet function accordingly.

internal/policy/cache/cache.go
```diff
@@ -72,25 +70,19 @@ func (c *PolicyCache) Set(key string, value string, err error) {
...
-
-	c, err := CreatePolicyCache()
-	if err != nil {
-		log.Debug("Failed to create PolicyCache")
-		return nil, err
+		return cache
 	}
 
-	return c, nil
+	return CreatePolicyCache()
```
st3penta, line 79:
```suggestion
	return &PolicyCache{
		Data: sync.Map{},
	}
```
With this change you could get rid of the CreatePolicyCache entirely, i think it is used only here

PR #2648, opened 2025-07-22
Fix and re-add acceptance tests for sealights workflow

ascerra:
# Description
Adding
  DOCKER_BUILDKIT: 1
  COMPOSE_DOCKER_CLI_BUILD: 1
to the acceptance test env field fixes the timeout issue we were seeing.

I believe this regressions could come from an update in either ubunut-latest image or podman as we are pulling latest changes for both each time we run.

# Testing
This has been tested in my mirror of the cli repo

See the many passing actions with this change in place
https://github.com/ascerra/cli-mirror/actions/workflows/checks-sealights.yaml

Assisted by: Cursor AI

### DIFF
--- a/.github/workflows/checks-codecov.yaml
+++ b/.github/workflows/checks-codecov.yaml
@@ -102,45 +102,6 @@ jobs:
           files: ./coverage-integration.out
           flags: integration
 
-  Acceptance:
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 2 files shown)


simonbaird:
Related to https://issues.redhat.com/browse/EC-1400 IIUC.

simonbaird:
Hmm, it won't actually run the new acceptance tests in this PR, since Sealights runs with "pull_request_target", so it uses the main branch workflows.

PR #2653, opened 2025-07-24
chore: call ImageReferenceInStubRegistry with format string

dheerajodha:
* without the format string as the 2nd argument, Sealights
   was throwing error as: "non-constant format string in call
   to ImageReferenceInStubRegistry"
* So, I added a format string.

### DIFF
--- a/.github/workflows/checks-sealights.yaml
+++ b/.github/workflows/checks-sealights.yaml
@@ -54,6 +54,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
           # Needed in hack/derive-version.sh
           fetch-depth: 0
 


.github/workflows/checks-sealights.yaml
```diff
@@ -54,6 +54,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
simonbaird, line 57: Not sure, but we might need some changes here to make sure it behaves as expected for post-merge CI.

dheerajodha: I feel that would be taken care of by this environment variable
https://github.com/conforma/cli/blob/7bc924d7ef2bae6754b1f0d29f420620fa874aac/.github/workflows/checks-sealights.yaml#L115

in case of push events, it would use the sha of the merge commit, which is expected.

simonbaird:
PR needs a new title, but I think actually it's been replaced by #2665 now.

dheerajodha:
> PR needs a new title, but I think actually it's been replaced by https://github.com/conforma/cli/pull/2665 now.

Hey Simon, yes youre right. I will close this PR.

PR #2666, opened 2025-07-30
post evaluation filter

joejstuart:


internal/evaluator/conftest_evaluator.go
```diff
@@ -445,11 +468,41 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, target EvaluationTarget
...
+					code := ""
+
+					// Look for result assignment with code in the rule body
+					if ruleRef.Body != nil {
+						for _, expr := range ruleRef.Body {
+							if expr.IsAssignment() {
+								// Check if this is a result assignment
+								if len(expr.Operands()) >= 2 {
+									// Look for result := { "code": "...", ... }
+									if term, ok := expr.Operands()[1].Value.(ast.Object); ok {
```
st3penta, line 481: You could use early returns here to flatten the syntax. Something like:
```suggestion
							if !expr.IsAssignment() {
								// Check if this is a result assignment
								continue
							}
							if len(expr.Operands()) < 2 {
								continue
							}
							
							// Look for result := { "code": "...", ... }
							if term, ok := expr.Operands()[1].Value.(ast.Object); !ok {
								continue
							}
```
and so on

internal/evaluator/filters.go
```diff
@@ -353,3 +390,1014 @@ func extractStringArrayFromRuleData(src ecc.Source, key string) []string {
...
+// - Validate that all include criteria were matched
+// - Generate comprehensive policy reports
+func (r *ECPolicyResolver) ResolvePolicy(rules policyRules, target string) PolicyResolutionResult {
+	return r.baseResolvePolicy(rules, target, r.processPackage)
+}
+
+// ResolvePolicy determines which rules and packages are included/excluded
+// based on the policy configuration and available rules, ignoring pipeline intention filtering.
+func (r *IncludeExcludePolicyResolver) ResolvePolicy(rules policyRules, target string) PolicyResolutionResult {
+	return r.baseResolvePolicy(rules, target, r.processPackage)
```
st3penta, line 531: is it really needed to pass the `processPackage` function as parameter here, if it can't be controlled by the user?

internal/evaluator/filters.go
```diff
@@ -353,3 +390,1014 @@ func extractStringArrayFromRuleData(src ecc.Source, key string) []string {
...
+// baseDeterminePackageInclusion contains the shared logic for determining package inclusion
+func (r *basePolicyResolver) baseDeterminePackageInclusion(pkg string, pkgRules []rule.Info, target string, result *PolicyResolutionResult) {
+	// Check if any rule in the package is included
+	hasIncludedRules := false
+	hasExcludedRules := false
+
+	for _, ruleInfo := range pkgRules {
+		ruleID := ruleInfo.Code
+		if result.IncludedRules[ruleID] {
+			hasIncludedRules = true
```
st3penta, line 646: given the `if-else` block below, you can add a break here
```suggestion
			hasIncludedRules = true
			break
```

internal/evaluator/filters.go
st3penta: ooh i see why, you're calling a base function so you want to pass the processPackage funcs from  `ECPolicyResolver` and `IncludeExcludePolicyResolver` respectively

internal/evaluator/filters.go
```diff
@@ -353,3 +390,1014 @@ func extractStringArrayFromRuleData(src ecc.Source, key string) []string {
...
+
+	matchers = append(matchers, "*")
+
+	matchers = append(matchers, extractCollections(result)...)
+
+	return matchers
+}
+
+// ComputeSuccesses computes success results for rules that didn't appear in warnings, failures, exceptions, or skipped.
+func ComputeSuccesses(
```
st3penta, line 1012: unused function?

internal/evaluator/filters.go
```diff
@@ -353,3 +390,1014 @@ func extractStringArrayFromRuleData(src ecc.Source, key string) []string {
...
+//  2. Add 1 if the name covers everything, i.e. "*"
+//  3. Add 10 if the name specifies a package name, e.g. "pkg", "pkg.", "pkg.*", or "pkg.rule",
+//     and an additional 10 based on the namespace depth of the pkg, e.g. "a.pkg.rule" adds 10
+//     more, "a.b.pkg.rule" adds 20, etc
+//  4. Add 100 if a term is used, e.g. "*:term", "pkg:term" or "pkg.rule:term"
+//  5. Add 100 if a rule is used, e.g. "pkg.rule", "pkg.rule:term"
+//
+// The score is cumulative. If a name is covered by multiple items in the guidelines, they
+// are added together. For example, "pkg.rule:term" scores at 210.
+func Score(name string) int {
```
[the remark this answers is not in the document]
joejstuart: I actually think I can remove the legacy filter. I'll take a look. I just had it there for backwards compatibility, but the `IncludeExclude` filter is good enough for that.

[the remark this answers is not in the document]
joejstuart: Looks like they are needed for one edge case. I renamed them to `Legacy*`

internal/evaluator/filters.go
joejstuart: idk. If it has both hasInclude and hasExclude, it's considered included.

st3penta (approved):
overall looks good, up to you if you want to address the outstanding nitpicks

simonbaird (commented):
I'd like to see something about the motivation in the commit message and/or PR description, i.e. something about the "why", i.e. why is this useful, why do we want it, etc.

cmd/validate/image_integration_test.go
```diff
@@ -47,45 +44,23 @@ import (
...
+	// Clear the download cache to ensure a clean state for this test
+	source.ClearDownloadCache()
+
 	ctx := utils.WithFS(context.Background(), afero.NewMemMapFs())
 	client := fake.FakeClient{}
 	commonMockClient(&client)
 	ctx = oci.WithClient(ctx, &client)
 	mdl := MockDownloader{}
-	downloaderCall := mdl.On("Download", mock.Anything, mock.Anything, false).Return(&ociMetadata.OCIMetadata{Digest: "sha256:da54bca5477bf4e3449bc37de1822888fa0fbb8d89c640218cb31b987374d357"}, nil).Times(noEvaluators)
+	// The downloader should be called twice per policy source: once during PreProcessPolicy and once during evaluator evaluation
```
simonbaird, line 55: Are we talking about a go-gather downloader here? Why does it get downloaded twice now?

internal/evaluator/filters.go
```diff
@@ -656,7 +644,7 @@ func (r *basePolicyResolver) baseDeterminePackageInclusion(pkg string, pkgRules
 		result.Explanations[pkg] = "package contains included rules"
 	} else if hasExcludedRules {
 		result.ExcludedPackages[pkg] = true
-		result.Explanations[pkg] = "package contains only excluded rules"
+		result.Explanations[pkg] = "package contains an excluded rule"
```
simonbaird, line 647: I don't understand the reason for this change. Seems like the original explanation is better and more accurage, since we're here only if there are no included rules..?

simonbaird (commented):
See comments inline.

internal/evaluator/filters.go
st3penta: That's because i left a comment on this (that i can't find anymore): 

> Here we only know that the package has NO included rules, and SOME excluded rules (without guarantee that they're all excluded)

internal/evaluator/filters.go
```diff
@@ -1008,83 +970,6 @@ func MakeMatchers(result Result) []string {
 	return matchers
 }
 
-// ComputeSuccesses computes success results for rules that didn't appear in warnings, failures, exceptions, or skipped.
-func ComputeSuccesses(
```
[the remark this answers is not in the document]
joejstuart: Yes, it's part of the evaluator. https://github.com/conforma/cli/blob/2eb90f5c4a87151505157f6d9b1fdc2075ebc4b9/internal/evaluator/conftest_evaluator.go#L678

cmd/validate/image_integration_test.go
joejstuart: I had a lot of problems with this test and part of the solution at one point was to clear the policy cache which made it so the policy was downloaded twice. It's downloaded once in `PreProcessPolicy` and again in the `Evaluate` function.
I see now that the cache doesn't need to be cleared, so I'll fix that.

simonbaird:
Looks like a rebase is needed here.

PR #2680, opened 2025-08-06
Migrate CRDs from `enterprise-contract/enterprise-contract-controller`

robnester-rh:
This PR is a first pass at migrating the CRDs from the `enterprise-contract/enterprise-contract-controller` repository.

### DIFF
--- a/.github/workflows/checks-codecov.yaml
+++ b/.github/workflows/checks-codecov.yaml
@@ -78,10 +78,11 @@ jobs:
 
       - name: Check for uncommitted changes
         run: |
-          if ! git diff --exit-code -s; then
-            for f in $(git diff --exit-code --name-only); do
-              echo "::error file=$f,line=1,col=1,endColumn=1::File was modified in build"
-            done
+          git status --porcelain
+          if ! git diff --exit-code; then
+            echo "::error::Files were modified during build"
+            echo -e "\033[1;33mHint:\033[0m Maybe you need to run \033[1;32mmake all\033[0m"
+            git diff --name-only
             exit 1
           fi
 
--- a/.github/workflows/crd-schema.yaml
+++ b/.github/workflows/crd-schema.yaml
@@ -64,6 +64,7 @@ jobs:
 
       - name: Download CRD dependencies
         run: |
+          go mod download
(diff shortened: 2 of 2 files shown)


.github/workflows/checks-codecov.yaml
```diff
@@ -69,22 +69,29 @@ jobs:
...
       - name: Show diff after `make generate`
         run: git diff --stat --patch
 
       - name: Check for uncommitted changes
         run: |
-          if ! git diff --exit-code -s; then
-            for f in $(git diff --exit-code --name-only); do
-              echo "::error file=$f,line=1,col=1,endColumn=1::File was modified in build"
-            done
+          git status --porcelain
```
joejstuart, line 81: I see this replaces the existing check. Is that ok, and should we now run `make all` when this check is triggered?

config/crd/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright The Conforma Contributors
```
joejstuart, line 1: I think this file is missing some content. Here is the original.

```yaml
$ cat kustomization.yaml 
# This kustomization.yaml is not intended to be run by itself,
# since it depends on service name and namespace that are out of this kustomize package.
# It should be run by config/default
resources:
- bases/appstudio.redhat.com_enterprisecontractpolicies.yaml
- enterprisecontractpolicy_editor_role.yaml
- enterprisecontractpolicy_viewer_role.yaml
- openshift_console_example.yaml
#+kubebuilder:scaffold:crdkustomizeresource

patchesStrategicMerge:
# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
# patches here are for enabling the conversion webhook for each CRD
#- patches/webhook_in_enterprisecontractpolicies.yaml
#+kubebuilder:scaffold:crdkustomizewebhookpatch

# the following config is for teaching kustomize how to do kustomization for CRDs.
configurations:
- kustomizeconfig.yaml
```

robnester-rh:
This is being closed as the underlying story has changed. The CRDs will be migrated to their own repository per team discussions.

PR #2681, opened 2025-08-07
Add test coverage for top level data key merging

simonbaird:
The motivation is to make sure PR #2656 doesn't change this behavior and potentially break things.

The two things being verified:
- If the value of the top level key is a map, then the map keys from each data source get merged
- If the value of the top level key is not a map, then we fail hard with an error.

Ref: https://issues.redhat.com/browse/EC-1418


simonbaird:
Rebased to remove tekton lint fails.

PR #2685, opened 2025-08-11
Konflux and ubi digest bumps (v05)

simonbaird:


simonbaird:
I think the CI didn't get triggered because I created this with the wrong target branch initially. Let's redo it..

PR #2696, opened 2025-08-20
Remove CLI_NAME build arg and version.CliName()

simonbaird:
Since we're calling Conforma by the same name in upstream and Red Hat builds, we don't need the CLI_NAME arg any more.

Also includes some container label adjustments based on new names for RHTAP and Enterprise Contract.


simonbaird:
This is low risk IMO. I want it in the v0.7 release branch, which I'm about to create, so let's merge.

PR #2703, opened 2025-08-25
Update the ubi base image digest (v0.7)

simonbaird:


simonbaird:
Should produce first green release of v0.7, see [EC-1384](https://issues.redhat.com/browse/EC-1384)

PR #2704, opened 2025-08-26
feat: build arm variant of image

arewm:
While we cross compile all binaries in the image, that is only useful if we are extracting binaries out of it. If a Task step uses the image, however, the container needs need to be able to be built on arm.


arewm:
This PR will not test arm builds as it is. If desired, I will temporarily enable arm builds for PRs to ensure that it works OOTB.

PR #2898, opened 2025-08-29
feat(vsa/rekor): store both intoto + dsse entries per attestation

joejstuart:
Rekor’s intoto:0.0.2 entries preserve the Statement but drop the DSSE signatures. Rekor’s dsse:0.0.1 entries keep signatures but aren’t suited for retrieving the Statement.

To support both use cases we now upload both entry types for each signed attestation and link them via the payload hash (sha256 of decoded payload).

Changes:
- UploadWithSigner: dual-writes intoto + dsse, logs UUIDs/indices + payload hash.
- prepareDSSEForRekor: canonicalizes base64, injects publicKey, computes payload hash.
- Retrieval updated to find entries by payload hash and classify by kind.
- This ensures attestations are retrievable and signatures verifiable directly from Rekor.

https://issues.redhat.com/browse/EC-1454


internal/validate/vsa/storage_rekor.go
```diff
@@ -94,51 +110,255 @@ func (r *RekorBackend) UploadWithSigner(ctx context.Context, envelopeContent []b
...
+	}
+
+	// Decode payload bytes (try URL-safe no-pad first, then std no-pad, then std)
+	decodeB64 := func(s string) ([]byte, error) {
+		if b, err := base64.RawURLEncoding.DecodeString(s); err == nil {
+			return b, nil
+		}
+		if b, err := base64.RawStdEncoding.DecodeString(s); err == nil {
+			return b, nil
+		}
```
st3penta, line 329: duplicate?

PR #2931, opened 2025-09-09
Add --show-warnings flag to control warning output

dheerajodha:
* Add --show-warnings flag to validate command with default `true`
* Modify text template to conditionally show warnings
* Update both image and input subcommands to use the flag

resolves: EC-1304

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -350,6 +350,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			}
 
 			showSuccesses, _ := cmd.Flags().GetBool("show-successes")
+			showWarnings, _ := cmd.Flags().GetBool("show-warnings")
 
 			// worker is responsible for processing one component at a time from the jobs channel,
 			// and for emitting a corresponding result for the component on the results channel.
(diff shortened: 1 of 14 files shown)


cmd/validate/image.go
```diff
@@ -350,6 +350,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			}
 
 			showSuccesses, _ := cmd.Flags().GetBool("show-successes")
+			showWarnings, _ := cmd.Flags().GetBool("show-warnings")
```
simonbaird, line 353: How does it know that the default should be "true"? I can't figure it out.

cmd/validate/validate.go
```diff
@@ -43,5 +43,6 @@ func NewValidateCmd() *cobra.Command {
 		Short: "Validate conformance with the provided policies",
 	}
 	validateCmd.PersistentFlags().Bool("show-successes", false, "")
+	validateCmd.PersistentFlags().Bool("show-warnings", true, "")
```
simonbaird, line 46: Oooh, here it is. Disregard previous question.

simonbaird (approved):
Lgtm.

PR #2945, opened 2025-09-15
feat(equivalence): add library to compare EnterpriseContractPolicy specs

joejstuart:
Implements a reusable package to determine if two ECP specs are functionally equivalent for a given image at a point in time. Includes normalization of sources, RuleData, include/exclude matchers, and volatile config handling.


st3penta (commented):
i left a few comments, but overall it looks very nice and tidy!

st3penta (approved):

PR #2960, opened 2025-09-24
chore(deps): update enterprise contract CRD source to conforma/crds

robnester-rh:
Switch from enterprise-contract-controller to conforma/crds repository for CRD definitions.

Ref: EC-1422


joejstuart:
A rebase with main will fix the Konflux build.

PR #2963, opened 2025-09-25
Support retrieving the vsa from file system path

joejstuart:
This adds support for retrieving a vsa from a file system path. This is helpful for debugging. it also make retrieveVSA more generic by supporting an indentifier instead of an image digest.


st3penta (approved):
lgtm

PR #2964, opened 2025-09-25
Fix concurrent symlink creation race condition

Acepresso:
When multiple workers validate images concurrently using the same policy sources, a race condition occurs where workers would simultaneously attempt to create symlinks from cached policy downloads to their work directories, resulting in "file exists" errors.

This fix introduces per-destination mutex synchronization to ensure only one worker can create a symlink for each specific destination path.

Also included a unit test that reproduces the bug.

Ref: https://issues.redhat.com/browse/EC-1406
Assisted-by: Claude 4 Sonnet (using Cursor)

### DIFF
--- a/internal/policy/source/source.go
+++ b/internal/policy/source/source.go
@@ -78,9 +78,13 @@ type PolicyUrl struct {
 // downloadCache is a concurrent map used to cache downloaded files.
 var downloadCache sync.Map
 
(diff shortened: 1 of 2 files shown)


simonbaird (approved):
Nice.

dheerajodha:
Merging this since @Acepresso is on PTO.

PR #2998, opened 2025-10-07
validate vsa command

joejstuart:
Adding `validate vsa subcommand`. The command supports validating a VSA from the filesystem, a single image or a snapshot. If an image or snapshot is provided, the VSA will be looked up from Rekor.
This does not do signature verification. That will be in a follow-up PR.

### DIFF
--- a/acceptance/go.mod
+++ b/acceptance/go.mod
@@ -4,10 +4,10 @@ go 1.24.4
 
 require (
 	cuelang.org/go v0.11.1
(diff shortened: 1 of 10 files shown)


joejstuart:
No idea about the SL tests. Getting this on 93 tests.

```
2025-10-08 01:49:44.412 INFO [Go Agent] Agent running mode set to integration tests [internal/agent.go:818] 
```

PR #2999, opened 2025-10-07
Upgrade oauth2 golang module and Konflux task refs (v0.5)

simonbaird:
I'm also including an unrelated commit to add a CPE label, similar to #2950 .

Ref: https://issues.redhat.com/browse/EC-1518


simonbaird:
Just a label change and some module upgrades. Should be low risk, so merging without a review.

PR #3000, opened 2025-10-07
Upgrade go deps & konflux refs, add CPE label (v0.6)

simonbaird:
Similar to #2999 but for v0.6 branch.

Ref: https://issues.redhat.com/browse/EC-1518


simonbaird:
I couldn't reproduce the acceptance test failure in `validate_input`, so let's try re-running it.

PR #3003, opened 2025-10-09
Use longer repoids in ubi.repo

simonbaird:
IIUC this file is used by Mintmaker. It was added in EC-1244 to prevent Mintmaker from failing. More specifically it's used by `rpm-lockfile-prototype` which IIUC Mintmaker is using to freshen rpm lockfiles.

For this change I'm hoping to prevent the "good" repoids (such as "ubi-9-for-aarch64-baseos-rpms" from being converted to "bad" repoids (such as "ubi-9-baseos-rpms") in our `rpms.lock.yaml file`, in PRs such as #2934.

The "bad" repoids can produce Conforma violations, since the shorter repoids are not in the allowed repo list, visible [here](https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/known_rpm_repositories.yml).

See also the `hack/update-rpm-lock.sh` script in this repo, which is our own custom script we've used in the past for updating the rpm lockfile.

### DIFF
--- a/ubi.repo
+++ b/ubi.repo
@@ -1,60 +1,60 @@
-[ubi-9-baseos-rpms]
+[ubi-9-for-$basearch-baseos-rpms]
 name = Red Hat Universal Base Image 9 (RPMs) - BaseOS
 baseurl = https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/$basearch/baseos/os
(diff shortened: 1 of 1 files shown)


simonbaird:
We're ignoring the spurious Sealights failures currently. Let's merge this and see how it goes.

PR #3012, opened 2025-10-23
Remove all Sealights integrations

ascerra:
The Sealights license is not being renewed, calls to Sealights agents must be removed
- Delete check-sealights github workflow
- Remove acceptance-sealights block from makefile
- Remove note about Sealights being the reason for 10s timeout


ascerra:
I see the sealights check still run here but that is because these checks use the configuration from the main branch, so once this PR merges the next PR should not have any Sealights checks running.

PR #3022, opened 2025-11-03
update multi-arch build matrix

robnester-rh:
This commit updates the multi-arch build matrix to include support for `ppc64le` and `s390x` subsequent to work from #2704 which added support for multi-arch builds.

Ref: EC-1519


simonbaird (approved):
Lgtm.

PR #3024, opened 2025-11-05
optimize build and cross-compiling.

robnester-rh:
This commit updates the `build.sh` script to optimize the multi-arch building and cross-compilation from 28 to 10 builds.

Ref: EC-1520

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -24,6 +24,22 @@ BUILDS="${1}"
 # Generally blank, but will be set to "redhat" for Konflux builds
 BUILD_SUFFIX="${2:-""}"
 
+# When building multi-arch images, only the linux/amd64 build needs to cross-compile all platform/arch combinations for download.
+# The other platforms only need their native binary. This reduces builds from 28 (7 binaries × 4 platforms) to 10 (7 + 1 + 1 + 1).
+
+# Check if TARGETOS and TARGETARCH are set (Dockerfile sets these via buildah --platform)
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -24,6 +24,23 @@ BUILDS="${1}"
...
+# When building multi-arch images, only the linux/amd64 build needs to cross-compile all platform/arch combinations for download.
+# The other platforms only need their native binary. This reduces builds from 28 (7 binaries × 4 platforms) to 10 (7 + 1 + 1 + 1).
+
+# Check if TARGETOS and TARGETARCH are set (Dockerfile sets these via buildah --platform)
+if [[ -n "${TARGETOS:-}" && -n "${TARGETARCH:-}" ]]; then
+    # We're in a Dockerfile build with TARGETOS/TARGETARCH set
+    if [[ "${TARGETOS}" == "linux" && "${TARGETARCH}" == "amd64" ]]; then
+        # amd64 builds ALL binaries for download availability
+        echo "Platform linux/amd64: Building all cross-compiled binaries for download"
+        BUILDS="darwin_amd64 darwin_arm64 linux_amd64 linux_arm64 linux_ppc64le linux_s390x windows_amd64"
```
simonbaird: The list is also defined in Dockerfile.dist, and passed in as `$1` via line 22 above. So I think we don't need to set BUILDS again there.

simonbaird: ..and see `Dockerfile`, which is the GitHub version IIUC. It already specifies that only the target platform is built, so the echo line on line 35 is misleading in that case.

PR #3025, opened 2025-11-05
Refine validate vsa output

joejstuart:
Refactor fallback image validation

- Walk each data source directory returned by GetPolicy. These are the actual directories (possibly symlinks) where data was downloaded. Walking them directly ensures we find files even if they're symlinks. This was the cause of inconsistent violation reporting.
 - Move report writing to shared code under internal/validate. This is so validate image and the fallback validation will show the same result format.


st3penta (approved):
As far as i can tell the PR seems fine, but tbh I don't feel like I understood it well, due to its size and to the fact that is a refactoring (harder to understand with respect to 'new stuff' PRs)

PR #3118, opened 2026-02-17
fix: VSA generation issues in verify-conforma-konflux-ta task

arewm:
The ENABLE_VSA code path had several bugs that together prevented VSA files
from being produced and archived correctly. The ec CLI rejects output paths
outside /tmp or cwd, --vsa-upload must only be passed for dsse format since
the local@ backend requires a signing key, and echo without -n writes a
trailing newline that breaks Tekton When Expression matching.

The trusted artifact archival was also broken: workDir was hardcoded to
/var/workdir rather than \$(params.TRUSTED_ARTIFACTS_EXTRACT_DIR), so
downstream tasks would find the snapshot nested under a conforma/ prefix
instead of at the archive root. The VSA_UPLOAD default was similarly
misaligned, pointing outside the extract dir so VSA files were never
captured in the archive. The report-json.json is now also copied into the
VSA directory so downstream tasks can generate SLSA VSAs from the
Conforma evaluation report.

Adds an acceptance test scenario for ENABLE_VSA=true with predicate format,
a new taskResultShouldEqual step definition for targeted result assertions,
and updates the existing snapshot for the echo -n fix.

## Test plan

- [x] Validated end-to-end in a Kind cluster with Konflux: verify-conforma produces VSA files in the trusted artifact, downstream attach-vsa task extracts and attaches them to container images

### DIFF
--- a/acceptance/kubernetes/kind/kubernetes.go
+++ b/acceptance/kubernetes/kind/kubernetes.go
@@ -381,7 +381,7 @@ func (k *kindCluster) TaskInfo(ctx context.Context) (*types.TaskInfo, error) {
 
 	results := map[string]any{}
 	for _, r := range tr.Status.Results {
-		results[r.Name] = r.Value
+		results[r.Name] = paramValue(r.Value)
 	}
 
 	info := types.TaskInfo{
--- a/acceptance/kubernetes/kubernetes.go
+++ b/acceptance/kubernetes/kubernetes.go
@@ -395,6 +395,30 @@ func taskResultsShouldMatchTheSnapshot(ctx context.Context) error {
 	return snaps.MatchSnapshot(ctx, "results", string(j), nil)
 }
 
+func taskResultShouldEqual(ctx context.Context, resultName, expectedValue string) error {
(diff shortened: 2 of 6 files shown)


simonbaird:
I found we do have some test coverage for this task in `features/ta_task_validate_image.feature`. That said, it looks like the ENABLE_VSA code path might be untested. Consider adding some extra scenarios to that feature so we have some coverage.

arewm:
I am still working on some additional changes when testing locally. I moved this PR to draft until I finish working on it.

arewm:
@simonbaird, I have moved this out of draft. The content of this PR is ready for review again. I tested this task out in my [own release pipeline](https://github.com/arewm/slsa-konflux-example/blob/main/managed-context/pipelines/slsa-e2e-release/slsa-e2e-release.yaml#L297 ) on KinD. The content of [that task](https://github.com/arewm/slsa-konflux-example/blob/main/managed-context/tasks/verify-conforma/0.1/verify-conforma-vsa.yaml) should match the one in this PR.

simonbaird:
Let's merge.

simonbaird (approved):

PR #3123, opened 2026-02-25
Support Sigstore bundle verification for cosign v3

SequeI:
cosign v3 stores signatures as OCI referrer bundles by default, but ec
only supported the legacy tag-based format. Detect bundles via
cosign.GetBundles() and route verification through the bundle-aware
code path, matching what the cosign CLI already does internally.

Also fix typos and incorrect examples in CLI help text.

You can check this behaviour out by using Cosign V3, signing an image, attaching the predicate.json, and trying to use `ec validate` from main branch, then from this branch.

Ref: https://issues.redhat.com/browse/EC-1689

### DIFF
--- a/internal/attestation/attestation_test.go
+++ b/internal/attestation/attestation_test.go
@@ -20,6 +20,7 @@ package attestation
 
 import (
 	"crypto/x509"
+	"encoding/base64"
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


simonbaird:
FYI I've tested this with a v2 and v3 signed image in a version of #3140 rebased on this PR, and it seems to be working nicely.

joejstuart (approved):

PR #3124, opened 2026-02-25
feat: ec pipeline supports large snapshot specs

kasemAlem:
Modify script to enable snapshots with large specs 

 
 
[STONEINTG-1470](https://issues.redhat.com/browse/STONEINTG-1470)


kasemAlem:
We intentionally do not pass -n so the fetch uses the current kubectl context. In Tekton the task runs in the pipeline namespace; locally the context is the workspace. The script comments have been updated to document this so the intended namespace (context namespace) is explicit.

Issue #1528, opened 2024-04-15
ec validate input is errs when policy set publicKey


lcarva:
With a policy.yaml that specifies a publicKey, e.g.:

```yaml
publicKey: k8s://openshift-pipelines/public-key
sources:
- data:
  - ...
  policy:
  - ...
```

The `ec validate input` command always fails:

```text
Error: 1 error occurred:
	* error validating file input.json: no check options or sig verifier configured

```

With the `--debug flag`:

```text
DEBU[0000] helpers.go:80 readPolicyConfigurationFile Loaded policy.yaml as policyConfiguration    
DEBU[0000] policy.go:394 parseEffectiveTime Chosen to use effective time of `now`, using current time 2024-04-15T18:44:38Z 
DEBU[0000] policy.go:309 loadPolicy Read EnterpriseContractPolicy as YAML        
DEBU[0000] policy.go:314 loadPolicy Unable to parse EnterpriseContractPolicy from "configuration:\n  # exclude:\n  # - tasks.required_tasks_found:prefetch-dependencies\n  # - hermetic_build_task.build_task_hermetic\n  # - test.no_test_warnings\n  # - redhat_manifests.redhat_manifests_missing\n  # - labels.required_labels\n  # - rpm_ostree_task.builder_image_param\n  # - sbom_cyclonedx.found\n  # - test.no_failed_tests\n  # include:\n  # - '@redhat'\n  include:\n  - attestation_task_bundle\ndescription: Rules for shipping content to registry.redhat.io\npublicKey: k8s://openshift-pipelines/public-key\nsources:\n- data:\n  - github.com/release-engineering/rhtap-ec-policy//data\n  - oci::quay.io/redhat-appstudio-tekton-catalog/data-acceptable-bundles:latest\n  name: Release Policies\n  policy:\n  # - oci::quay.io/enterprise-contract/ec-release-policy:latest\n  - /home/lucarval/src/enterprise-contract/ec-policies/policy/lib\n  - /home/lucarval/src/enterprise-contract/ec-policies/policy/release\n  # ruleData:\n  #   allowed_registry_prefixes:\n  #   - registry.access.redhat.com/\n  #   - registry.redhat.io/\n  #   - quay.io/rh-osbs/ubi9\n  #   - localhost/rhtap-final-image\n  #   allowed_rpm_ostree_builder_image_prefixes:\n  #   - quay.io/centos-bootc/bootc-image-builder\n" 
[...]

PR #3134, opened 2026-02-27
fix ec validate input failing when policy has publicKey

robnester-rh:
When a policy spec includes a publicKey field, ec validate input would fail with 'no check options or sig verifier configured' because the signature verifier isn't initialized for input validation scenarios.

Return the publicKey from the policy spec directly when SigVerifier is not initialized.

Ref: #1528
Ref: EC-1666

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto"
 	_ "embed"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


simonbaird:
We were talking about whether the key (and other sigstoreopts) matters for `ec validate input`. Actually I think maybe it does because of this:

```
  1006   >   // Add the policy config we just prepared
  1007   >   config["config"] = map[string]interface{}{
  1008   >   >   "policy":                pc,
  1009   >   >   "default_sigstore_opts": opts,
  1010   >   }
```

So maybe the isPEM check is reasonable after all.

PR #3221, opened 2026-04-07
Dep bump v0.7

Elijah-Destigni:
https://redhat.atlassian.net/browse/SECURESIGN-4030
https://redhat.atlassian.net/browse/SECURESIGN-4078
https://redhat.atlassian.net/browse/SECURESIGN-3984
https://redhat.atlassian.net/browse/SECURESIGN-3982

PR #3231, opened 2026-04-13
Upgrade golang and docker/cli (v07)

simonbaird:
Since we've having trouble getting the tests to pass in PR #3221, I thought I'd try splitting it into smaller chunks. This PR contains just the golang upgrade, which IIUC does fix a couple of the vulnerabilities we're aiming resolve.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.25.3
+golang 1.25.8
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@
 
 ## Build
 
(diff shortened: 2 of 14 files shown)


simonbaird:
Acceptance tests are green, which is great. I guess the snapshot change in the wiremock fix was significant.

The only thing failing now is the Conforma check, and it's due to a `github.com/docker/cli` CVE being detected by clair (even though it's Windows-only IIUC).

So I'm adding a docker/cli update here in this PR too.

st3penta (approved):

PR #3232, opened 2026-04-14
fix(RELEASE-2158): refactor trusted-ca mounts in verify-conforma task

querti:
Refactor trusted-ca volume mounts in verify-conforma-konflux-ta to use
directory mounts (/etc/ssl/certs and /mnt/trusted-ca) instead of
subPath-based mounts. This is consistent with the mount style used
across release-service-catalog tasks and ensures that the task doesn't
exit with a failure when the configmap is not present. Remove the
redundant trusted-ca volumeMount from the report step, which does
not need CA certificate access.

Assisted-by: Cursor

### DIFF
--- a/tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
+++ b/tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
@@ -264,11 +264,10 @@ spec:
       - mountPath: /var/workdir
         name: workdir
       - name: trusted-ca
-        mountPath: /mnt/trusted-ca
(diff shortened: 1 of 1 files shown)


simonbaird:
@querti what do you think about code rabbit's comment?

simonbaird:
Lgtm, but see what you think about the bot's comment.

I retriggered the tests because it seems likely the acceptance failure was a temporary flake.

dheerajodha:
@querti do you plan to open this PR for review again?

querti:
I don't completely remember the context of this PR, but it's not needed anymore. I'm gonna close it.

PR #3245, opened 2026-04-16
Update some deps in v0.6 branch

simonbaird:
Ref: https://redhat.atlassian.net/browse/EC-1766


simonbaird:
I think it's nearly impossible to update the tekton and buildkit dependencies, so I've back those out and simplified this PR. If it goes green we can merge and call it the final v0.6 builds, since v0.6 branch is about to be retired.

simonbaird:
With https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/17207 merged I think the ec check might go green. Let's see.

simonbaird:
Pushed even fresher ubi and konflux refs...

robnester-rh (approved):

PR #3259, opened 2026-04-21
fix(RELEASE-2397): custom ca support in verify-conforma task

mmalina:
Refactor trusted-ca volume mounts in verify-conforma-konflux-ta to use directory mount (/mnt/trusted-ca) instead of subPath-based mounts. This is consistent with the mount style used across release-service-catalog tasks and ensures that the task doesn't exit with a failure when the configmap is not present. Remove the redundant trusted-ca volumeMount from the report step, which does not need CA certificate access.

Also, update the build-trusted-artifacts image to include a related fix for custom ca handling.

Related build-trusted-artifacts PR: https://github.com/konflux-ci/build-trusted-artifacts/pull/315

### DIFF
--- a/tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
+++ b/tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml
@@ -266,10 +266,6 @@ spec:
       - name: trusted-ca
         mountPath: /mnt/trusted-ca
         readOnly: true
-      - name: trusted-ca
-        mountPath: /etc/ssl/certs/ca-custom-bundle.crt
-        subPath: ca-bundle.crt
-        readOnly: true
     env:
       - name: "ORAS_OPTIONS"
         value: "$(params.ORAS_OPTIONS)"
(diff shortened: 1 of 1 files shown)


mmalina:
Cc: @querti @simonbaird

querti (approved):

simonbaird (approved):

mmalina:
Thanks for the approval @simonbaird . It seems you will also need to do /ok-to-test

Will you be able to merge this as well? Or shall I add this to Jeff's spreadsheet to get his approval?

simonbaird:
I'm happy to merge it if you are. Will it hit the prod release pipeline immediately, or is there a git sha pin involved?

mmalina:
> I'm happy to merge it if you are. Will it hit the prod release pipeline immediately, or is there a git sha pin involved?

Since @querti approved this as well, I think we are ready. The git revision sha of the task is controlled by the verify_ec_task_git_revision pipeline parameter: https://github.com/konflux-ci/release-service-catalog/blob/development/pipelines/managed/rh-rpm-advisories/rh-rpm-advisories.yaml#L44

This is passed by the release-service operator when it creates the pipeline. Let me check where it comes from. I thought it was in the ReleaseServiceCatalog, but looks like it's not the case: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/release/staging/release_service_config.yaml

mmalina:
OK, I found out how it works. It relies on the value here: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/enterprise-contract/kustomization.yaml#L16
So after this is merged, we will also have to update that. (And we can test with main before we do that.)

Also, there are PRs from Joe Stuart to remove this logic and just use the konflux branch of conforma/cli:

- https://github.com/konflux-ci/release-service/pull/1453
- https://github.com/konflux-ci/release-service-catalog/pull/2095

BTW, I noticed Conforma is failing on this PR. There is a CVE. I see you merged a ubi bump last night. Let me rebase.

simonbaird:
If it turns green I'll likely merge.

mmalina:
I see Checks / Acceptance failed. I don't understand the errors. Is that something I should fix?

simonbaird:
There was an acceptance test fail, but I believe it's a known flake, unrelated to the PR. Retriggered the checks.

mmalina:
Thanks for the merge!

PR #3267, opened 2026-04-28
Add --skip-att-sig-check flag

simonbaird:
Implement --skip-att-sig-check flag to skip attestation signature validation checks, mirroring the existing --skip-image-sig-check flag. When enabled, attestation signature verification is bypassed during image validation.

Why? Often I'm debugging/troubleshooting something and I get given an image ref to look at. We can use cosign download attestation to inspect the attestation, which is very useful, but if we want to try running Conforma against it, we must either guess, find, or ask to be provided with the public key. Sometimes that's not so difficult, but other times it may be very difficult or even impossible. (Consider for example if the image was built on an ephemeral cluster and the signing secret used is gone forever.)

Now we can use --skip-image-sig-check and --skip-att-sig-check and carry on with the debugging. Note that we added the --skip-image-sig-check recently for other reasons, see https://redhat.atlassian.net/browse/EC-1647. The
--skip-att-sig-check is perhaps a little more complicated because we need to add a function that can download the attestation without verifying it.

The argument against this is that it may encourage less secure practices, but I would say it's acceptable because we're not changing the default behavior, which is always to require signature verification.

Ref: https://redhat.atlassian.net/browse/EC-1815

### DIFF
--- a/features/validate_image.feature
+++ b/features/validate_image.feature
@@ -56,6 +56,65 @@ Feature: evaluate enterprise contract
     # builtin.attestation.image_check is not found in the success output
     Then the output should match the snapshot
 
+  Scenario: invalid att signature with valid image sig and skip-att-sig-check flag
+    Given a key pair named "known"
+    Given a key pair named "unknown"
+    Given an image named "acceptance/invalid-att-signature"
+
+    # We're going to use the "known" public key when validating, so the image
+    # sig check should pass but the att sig check should fail (if it wasn't skipped).
+    Given a valid image signature of "acceptance/invalid-att-signature" image signed by the "known" key
+    Given a valid attestation of "acceptance/invalid-att-signature" signed by the "unknown" key
+
+    Given a git repository named "invalid-att-signature" with
+      | main.rego | examples/happy_day.rego |
+    Given policy configuration named "invalid-att-signature-policy" with specification
+    """
+    {
+      "sources": [
+        {
+          "policy": [
+            "git::https://${GITHOST}/git/invalid-att-signature.git"
+          ]
+        }
+      ]
+    }
+    """
+    # Notice we're using the --skip-att-sig-check flag, which is the reason this is
+    # green even though attestation is signed with a different key to the one we're providing
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/invalid-att-signature --policy acceptance/invalid-att-signature-policy --public-key ${known_PUBLIC_KEY} --skip-att-sig-check --rekor-url ${REKOR} --show-successes --output json"
+    Then the exit status should be 0
+    Then the output should match the snapshot
+
+  Scenario: skip both image and att sig checks for debugging
+    Given a key pair named "known"
+    Given a key pair named "unknown"
+    Given an image named "acceptance/skip-both-sig-checks"
+
+    # Both signatures use the wrong key. Without the skip flags this would fail.
+    Given a valid image signature of "acceptance/skip-both-sig-checks" image signed by the "unknown" key
+    Given a valid attestation of "acceptance/skip-both-sig-checks" signed by the "unknown" key
+
+    Given a git repository named "skip-both-sig-checks" with
+      | main.rego | examples/happy_day.rego |
+    Given policy configuration named "skip-both-sig-checks-policy" with specification
+    """
+    {
+      "sources": [
+        {
+          "policy": [
+            "git::https://${GITHOST}/git/skip-both-sig-checks.git"
+          ]
+        }
+      ]
+    }
+    """
+    # This is the expected real-world debugging usage: skip all signature checks
+    # to focus on policy evaluation when the signing key is unavailable
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/skip-both-sig-checks --policy acceptance/skip-both-sig-checks-policy --public-key ${known_PUBLIC_KEY} --skip-image-sig-check --skip-att-sig-check --rekor-url ${REKOR} --show-successes --output json"
+    Then the exit status should be 0
+    Then the output should match the snapshot
+
   Scenario: happy day with git config and yaml
     Given a key pair named "known"
     Given an image named "acceptance/ec-happy-day"
--- a/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
+++ b/internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
@@ -201,9 +201,32 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
 		return a.parseAttestationsFromBundles(layers)
 	}
 
-	// Extract the signatures from the attestations here in order to also validate that
-	// the signatures do exist in the expected format.
-	for _, sig := range layers {
+	return a.parseAttestationsFromSignatures(layers)
+}
+
+// FetchAttestationsWithoutVerification fetches attestations from the registry
+// without performing signature verification. This is used when --skip-att-sig-check
+// is enabled but we still need the attestation data for policy evaluation.
+func (a *ApplicationSnapshotImage) FetchAttestationsWithoutVerification(ctx context.Context) error {
+	if trace.IsEnabled() {
+		region := trace.StartRegion(ctx, "ec:fetch-attestations-without-verification")
+		defer region.End()
+	}
+
+	sigs, err := oci.NewClient(ctx).FetchAttestationLayers(a.reference)
+	if err != nil {
+		return err
+	}
+
+	if a.hasBundles(ctx) {
+		return a.parseAttestationsFromBundles(sigs)
+	}
+
+	return a.parseAttestationsFromSignatures(sigs)
+}
+
+func (a *ApplicationSnapshotImage) parseAttestationsFromSignatures(sigs []cosignOCI.Signature) error {
+	for _, sig := range sigs {
 		att, err := attestation.ProvenanceFromSignature(sig)
 		if err != nil {
 			return fmt.Errorf("unable to parse untyped provenance: %w", err)
@@ -283,7 +306,8 @@ func (a *ApplicationSnapshotImage) parseAttestationsFromBundles(layers []cosignO
 // ValidateAttestationSyntax validates the attestations against known JSON
 // schemas, errors out if there are no attestations to check to prevent
 // successful syntax check of no inputs, must invoke
-// [ValidateAttestationSignature] to prefill the attestations.
+// [ValidateAttestationSignature] or [FetchAttestationsWithoutVerification]
+// to prefill the attestations.
 func (a ApplicationSnapshotImage) ValidateAttestationSyntax(ctx context.Context) error {
 	if trace.IsEnabled() {
 		region := trace.StartRegion(ctx, "ec:validate-attestation-syntax")
--- a/internal/image/validate.go
+++ b/internal/image/validate.go
@@ -19,6 +19,7 @@ package image
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"runtime/trace"
 	"sort"
 	"time"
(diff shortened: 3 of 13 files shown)


simonbaird:
Making it a draft because it has no Jira (yet anyway).

simonbaird:
This isn't in the sprint currently, so maybe review it only after you reviewed everything else.

internal/image/validate.go
```diff
@@ -81,9 +81,18 @@ func ValidateImage(ctx context.Context, comp app.SnapshotComponent, snap *app.Sn
...
-	out.SetAttestationSignatureCheckFromError(a.ValidateAttestationSignature(ctx))
-	if !out.AttestationSignatureCheck.Passed {
-		return out, nil
+	// Handle attestation signature validation
+	if p.SkipAttSigCheck() {
+		log.Debug("Attestation signature check skipped, fetching attestations without verification")
+		if err := a.FetchAttestationsWithoutVerification(ctx); err != nil {
+			log.Debugf("Failed to fetch attestations without verification: %v", err)
+			// Continue with validation even if attestation fetch fails
+		}
```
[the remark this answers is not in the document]
simonbaird: Claude suggests this is intentional:

>  The silent-continue is intentional and correct here. When --skip-att-sig-check is used, the user explicitly wants to proceed even when attestation verification can't happen — that's the whole point of the flag. If we propagate the fetch error, the command fails on images where attestations can't be fetched without verification, which defeats the purpose.

>  Looking at what happens downstream (lines 100-122): if the fetch fails, a.Attestations() returns empty, and line 112-121 produces a clear policy violation: "No attestations contain a subject that match the given image." That's a much better user experience than an opaque registry error — the user sees the policy result and understands what happened.

>  The existing --skip-image-sig-check path (line 78-82) follows the same pattern — it silently skips without error propagation. The behavior is deliberate for a debugging/troubleshooting flag.

simonbaird:
Doing some extra work to get the Codecov check green.

internal/image/validate.go
st3penta, line 95: If this fails (registry unreachable, auth error, etc.), the error disappears into a Debug log and the user gets a confusing "No attestations contain a subject" violation downstream instead of the actual cause.

```suggestion
		if err := a.FetchAttestationsWithoutVerification(ctx); err != nil {
			log.Warnf("Failed to fetch attestations without verification: %v", err)
			out.SetAttestationSignatureCheckFromError(fmt.Errorf("failed to fetch attestations (signature check skipped): %w", err))
			return out, nil
		}
```

internal/image/validate.go
```diff
@@ -81,9 +81,18 @@ func ValidateImage(ctx context.Context, comp app.SnapshotComponent, snap *app.Sn
...
+	} else {
+		out.SetAttestationSignatureCheckFromError(a.ValidateAttestationSignature(ctx))
+		if !out.AttestationSignatureCheck.Passed {
+			return out, nil
+		}
 	}
```
st3penta, line 101: When this path is taken, `AttestationSignatureCheck` in the output is `{passed: false, result: null}` -- indistinguishable from "never reached." An explicit "skipped" result would make the output unambiguous for downstream consumers and audit trails. (Same gap exists for `--skip-image-sig-check`, but out of scope here.)

Also: combining both skip flags silently disables all crypto verification with no indication in the output. A Warn log when both are active would help catch accidental misuse.

internal/image/validate.go
```diff
@@ -81,9 +81,18 @@ func ValidateImage(ctx context.Context, comp app.SnapshotComponent, snap *app.Sn
 		out.SetImageSignatureCheckFromError(a.ValidateImageSignature(ctx))
 	}
 
-	out.SetAttestationSignatureCheckFromError(a.ValidateAttestationSignature(ctx))
-	if !out.AttestationSignatureCheck.Passed {
-		return out, nil
+	// Handle attestation signature validation
+	if p.SkipAttSigCheck() {
+		log.Debug("Attestation signature check skipped, fetching attestations without verification")
```
st3penta: Skipping a security check is operationally significant -- if this flag is accidentally left in a CI config, operators should be able to spot it without enabling Debug verbosity. Consider `log.Warn` here.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -201,9 +202,49 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
...
+// without performing signature verification. This is used when --skip-att-sig-check
+// is enabled but we still need the attestation data for policy evaluation.
+func (a *ApplicationSnapshotImage) FetchAttestationsWithoutVerification(ctx context.Context) error {
+	if trace.IsEnabled() {
+		region := trace.StartRegion(ctx, "ec:fetch-attestations-without-verification")
+		defer region.End()
+	}
+
+	remoteOpts := oci.CreateRemoteOptions(ctx)
+	signedEntity, err := ociremote.SignedEntity(a.reference, ociremote.WithRemoteOptions(remoteOpts...))
```
st3penta: This calls `ociremote.SignedEntity` directly, bypassing the `oci.Client` abstraction every other OCI operation in this file uses. That import was deliberately removed in #3269 to centralize OCI access. Adding a method to `oci.Client` would keep the pattern consistent and make this testable with `fake.FakeClient`.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -201,9 +202,49 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
...
+		return a.parseAttestationsFromBundles(sigs)
+	}
+
+	sigs, err := layers.Get()
+	if err != nil {
+		return fmt.Errorf("failed to get attestation signatures: %w", err)
+	}
+
+	return a.parseAttestationsFromSignatures(sigs)
+}
```
st3penta, line 226: This method has four error paths and a bundle/non-bundle branch, none tested at the unit level. The integration test does not mock this path either, so the fetch silently fails there. Routing through `oci.Client` (per the comment above) would make it straightforward to test with the existing fake infrastructure.

internal/image/validate_test.go
```diff
@@ -708,3 +708,122 @@ func TestValidateImageSkipImageSigCheck(t *testing.T) {
...
+			expectAttSigResult:      true,
+			expectImageSigCheckCall: true,
+		},
+		{
+			name:                    "skip attestation signature check enabled",
+			skipAttSigCheck:         true,
+			expectAttSigResult:      false,
+			expectImageSigCheckCall: true,
+		},
+	}
```
st3penta, line 694: No mocks are set up for `ociremote.SignedEntity`, so `FetchAttestationsWithoutVerification` always fails silently in the skip-enabled case. This validates "error was swallowed" rather than "attestations were fetched without verification" -- a regression that ignores the flag entirely would still pass.

features/validate_image.feature
```diff
@@ -56,6 +56,31 @@ Feature: evaluate enterprise contract
...
+    """
+    When ec command is run with "validate image --image ${REGISTRY}/acceptance/ec-happy-day --policy acceptance/ec-policy --public-key ${known_PUBLIC_KEY} --skip-att-sig-check --rekor-url ${REKOR} --show-successes --output json"
+    Then the exit status should be 0
+    # The only difference to the happy day scenario is that
+    # builtin.attestation.signature_check is not found in the success output
+    Then the output should match the snapshot
+
   Scenario: happy day with git config and yaml
     Given a key pair named "known"
     Given an image named "acceptance/ec-happy-day"
```
st3penta: The attestation here is signed with the correct key, so the skip flag is redundant. The `--skip-image-sig-check` has a companion scenario at line 214 with an *invalid* signature that proves the flag is needed. An equivalent scenario with a wrong attestation key would catch a regression that silently ignores this flag.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -201,9 +202,49 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
...
+			return fmt.Errorf("failed to get attestation signatures: %w", err)
+		}
+		return a.parseAttestationsFromBundles(sigs)
+	}
+
+	sigs, err := layers.Get()
+	if err != nil {
+		return fmt.Errorf("failed to get attestation signatures: %w", err)
+	}
+
```
st3penta: `layers.Get()` is identical in both branches -- could hoist it above the conditional.

```go
sigs, err := layers.Get()
if err != nil {
    return fmt.Errorf("failed to get attestation signatures: %%w", err)
}

if useBundles {
    return a.parseAttestationsFromBundles(sigs)
}
return a.parseAttestationsFromSignatures(sigs)
```

internal/image/validate_test.go
```diff
@@ -708,3 +708,122 @@ func TestValidateImageSkipImageSigCheck(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateImageSkipAttSigCheck(t *testing.T) {
+	tests := []struct {
+		name                    string
+		skipAttSigCheck         bool
+		expectAttSigResult      bool
+		expectImageSigCheckCall bool
```
st3penta: `expectImageSigCheckCall` is populated in both cases but never asserted. Remove it or add the intended check.

st3penta (changes requested):
Main concerns: error visibility when the fetch fails in skip mode, the new method bypasses the `oci.Client` abstraction centralized in #3269, and test coverage for the new code path needs work.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -201,9 +202,49 @@ func (a *ApplicationSnapshotImage) ValidateAttestationSignature(ctx context.Cont
...
+	return a.parseAttestationsFromSignatures(sigs)
+}
+
+func (a *ApplicationSnapshotImage) parseAttestationsFromSignatures(sigs []cosignOCI.Signature) error {
```
st3penta, line 228: `ValidateAttestationSyntax` docstring references `ValidateAttestationSignature` as the required prefill, but this function is now also a valid prefill path. Worth updating the docstring to mention both.

internal/image/validate_test.go
st3penta, line 694: Consider adding a case with both `--skip-image-sig-check` and `--skip-att-sig-check` enabled -- that is the expected real-world debugging usage per the PR description.

features/validate_image.feature
simonbaird: Yeah, I was being lazy here main to satisfy Codecov stats. :sweat_smile:

simonbaird:
Got some changes in progress. Will move back to draft while I'm working on them.

simonbaird:
Claude says all the suggestions have been addressed.

It might be true.. :sweat_smile:

simonbaird:
Claude and I agree that the Covscan number is close enough.

internal/image/validate.go
```diff
@@ -76,14 +77,27 @@ func ValidateImage(ctx context.Context, comp app.SnapshotComponent, snap *app.Sn
 
 	// Handle image signature validation
 	if p.SkipImageSigCheck() {
```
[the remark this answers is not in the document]
simonbaird: Yeah, this is intentional I guess.

simonbaird:
The [nil-deref] is being fixed. Here are my local agents take on the other items:

* audit-gap — Valid observation but out of scope for this PR. Both skip flags behave the same way, and introducing a "skipped" status would be a broader change to the output model. Good follow-up item.
* data-integrity — Intentional by design, no action needed.
* test-integrity — Fair point but the mock assertions are actually more precise (they verify the right code path ran). The old output-scanning was indirect. I'd leave as-is.

st3penta (approved):
lgtm

PR #3273, opened 2026-05-05
Update ubi and go jose

simonbaird:
This is going to fail because of the expired TUF root cert. See PR #3271 .

Also added a the Konflux refs bump, since I figured I might as well.


st3penta (approved):

simonbaird:
This was blended with https://github.com/conforma/cli/pull/3271 which is now merged.

PR #3274, opened 2026-05-06
Restore tekton task memory limits

st3penta:
Some tenants are experiencing OOM issues due to the much lower limit, so let's restore it to the original value.


simonbaird (approved):

simonbaird:
As discussed on slack, to get this green we have to pull in changes from #3271 and #3273 .

simonbaird:
As discussed on slack, the two teams that need this (openstack and ocp-art) have already modified their pipeline defaults to increase the memory resources.

Not sure if any other teams will be impacted, so we might decide this change is actually not needed now.

simonbaird:
Set it to draft, since I think the current plan is to abandon.

PR #3276, opened 2026-05-06
feat: add direct OPA evaluator with shared base infrastructure

joejstuart:
Add a new opaEvaluator that evaluates policies directly via OPA's rego API instead of going through conftest's runner. This eliminates the conftest runner overhead while reusing conftest's Engine for policy compilation and data loading.

Extract shared infrastructure into basePolicyEvaluator to eliminate ~400 lines of duplication between the two evaluators. Both now share policy download/inspection, data directory preparation, capabilities management, post-processing, and success computation.

https://redhat.atlassian.net/browse/EC-1806

Key changes:
- New opaEvaluator with direct rego.Eval() queries matching conftest's engine.Check() semantics (same regexes, exception handling, success counting)
- basePolicyEvaluator embedded struct with 12 shared methods
- sync.Once lazy initialization on both evaluators to prevent data races from concurrent worker goroutines
- BuildInput() on ApplicationSnapshotImage for in-memory OPA input delivery without disk I/O
- ParsedInput field on EvaluationTarget for passing pre-parsed input
- EC_USE_OPA=1 environment variable gate for selecting the OPA evaluator


internal/evaluator/conftest_evaluator.go
```diff
@@ -858,7 +859,7 @@ func (c conftestEvaluator) computeSuccesses(
 		if unifiedFilter != nil {
 			// Use the unified filter to check if this success should be included
 			filteredResults, _ := unifiedFilter.FilterResults(
-				[]Result{success}, rules, imageRef, componentName, missingIncludes, time.Now())
+				[]Result{success}, rules, imageRef, componentName, missingIncludes, effectiveTime)
```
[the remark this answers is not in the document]
joejstuart: Out of scope for this change

simonbaird:
Jira?

internal/evaluator/opa_evaluator.go
```diff
@@ -17,33 +17,374 @@
...
+	var results []Result
+	for _, result := range resultSet {
+		for _, expression := range result.Expressions {
+			expressionValues, ok := expression.Value.([]any)
+			if !ok || len(expressionValues) == 0 {
+				results = append(results, Result{})
+				continue
+			}
+			for _, v := range expressionValues {
+				switch val := v.(type) {
```
st3penta, line 382: The type switch has no `default` case. Values that aren't `string` or `map[string]any` (e.g. `bool` from exception queries, `int`) are silently dropped — an empty `Result{}` is never appended for them. Since this is a policy enforcement tool, a silently dropped denial is a fail-open bug.

Add a `default` case that either returns an error or appends a result with a descriptive message so unexpected types surface instead of vanishing.

internal/evaluator/opa_evaluator.go
```diff
@@ -17,33 +17,374 @@
...
+		}
+		for _, r := range module.Rules {
+			name := r.Head.Name.String()
+			if !isOPAFailure(name) && !isOPAWarning(name) {
+				continue
+			}
+			ruleCount++
+			found := false
+			for _, existing := range ruleNames {
+				if strings.EqualFold(existing, name) {
```
st3penta: `strings.EqualFold` makes dedup case-insensitive, but OPA/Rego rule names are case-sensitive. Two distinct rules like `deny_Check` and `deny_check` would collapse into one, silently dropping whichever comes second.

Not a problem with current conforma policies (all bare `deny`/`warn`), but it's a one-liner fix:

```suggestion
				if existing == name {
```

features/validate_input_opa.feature
```diff
@@ -0,0 +1,46 @@
+Feature: validate input with OPA evaluator
+  The ec command line should produce correct results for input validation using the OPA evaluator
+
+  Background:
+    Given the environment variable is set "EC_USE_OPA=1"
```
st3penta: `validate input` never checks `IsOpaEnabled()` — it always creates a conftest evaluator (see `internal/evaluation_target/input/input.go`). These acceptance tests set `EC_USE_OPA=1` but silently run against conftest, giving false confidence in OPA coverage.

Either wire `IsOpaEnabled()` into the input validation path, or rename/tag these tests to make clear they're validating conftest behavior.

internal/evaluator/opa_evaluator.go
```diff
@@ -17,33 +17,374 @@
...
+			} else {
+				outcome.Warnings = append(outcome.Warnings, rr)
+			}
+		}
+		outcome.Exceptions = append(outcome.Exceptions, exceptions...)
+	}
+
+	resultCount := len(outcome.Failures) + len(outcome.Warnings) + len(outcome.Exceptions) + successes
+	if resultCount < ruleCount {
+		successes += ruleCount - resultCount
```
st3penta: Success padding uses `ruleCount` (total rule occurrences across modules) instead of `len(ruleNames)` (unique rules actually evaluated). If the same rule name appears in multiple files within a namespace — normal for multi-file Rego packages — `ruleCount` exceeds `len(ruleNames)`, and the padding inflates the success count.

```suggestion
	if resultCount < len(ruleNames) {
		successes += len(ruleNames) - resultCount
```

internal/evaluator/opa_evaluator.go
```diff
@@ -17,33 +17,374 @@
...
+
+		ruleQuery := fmt.Sprintf("data.%s.%s", namespace, ruleName)
+		ruleResults, err := o.evalOPAQuery(ctx, input, ruleQuery)
+		if err != nil {
+			return Outcome{}, fmt.Errorf("query rule: %w", err)
+		}
+
+		for _, rr := range ruleResults {
+			if len(exceptions) > 0 {
+				continue
```
st3penta: When exceptions match, `continue` drops the rule result entirely — no entry in Failures, Warnings, or Exceptions with the original message. Downstream consumers and audit logs can't distinguish "rule passed" from "rule failed but was excepted".

No current conforma policies use the conftest exception mechanism, so this is latent. But if exceptions are ever adopted, the result will be invisible suppression with no audit trail.

internal/evaluator/base_evaluator.go
```diff
@@ -0,0 +1,426 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+
+	"github.com/conforma/cli/internal/opa"
+	"github.com/conforma/cli/internal/opa/rule"
+	"github.com/conforma/cli/internal/policy/source"
+	"github.com/conforma/cli/internal/utils"
+)
+
+type basePolicyEvaluator struct {
```
st3penta, line 41: `basePolicyEvaluator` is only embedded by `opaEvaluator`. The PR description says it extracts shared infrastructure for both evaluators, but `conftestEvaluator` still has its own parallel copies of the same fields and methods.

Refactoring `conftestEvaluator` to embed `basePolicyEvaluator` would deliver on the "shared base" goal and reduce the duplicated code.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
```diff
@@ -419,6 +419,54 @@ type Input struct {
 	PolicySpec    ecc.EnterpriseContractPolicySpec `json:"policy_spec,omitempty"`
 }
 
+// BuildInput constructs the OPA input as a Go map and JSON bytes without disk I/O.
+// The JSON marshal/unmarshal round-trip ensures correct types for OPA (e.g. numbers
+// as float64, consistent key ordering).
+func (a *ApplicationSnapshotImage) BuildInput(_ context.Context) (map[string]any, []byte, error) {
```
st3penta, line 423: `BuildInput()` has zero dedicated tests. The JSON marshal/unmarshal round-trip (converting Go types to OPA-compatible `map[string]any`) is the key differentiator from `WriteInputFile` and could introduce subtle type mismatches — e.g. `int64` becoming `float64` in JSON. The conditional parent-image logic and both error branches are also unexercised.

Unit tests covering the round-trip shape, parent ref inclusion/exclusion, and both error paths would catch divergence from conftest's input before it reaches production.

internal/evaluator/base_evaluator.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	if _, err := f.WriteString(data); err != nil {
+		return err
+	}
+	log.Debugf("Capabilities file written to %s", f.Name())
+	return nil
+}
+
+func (b *basePolicyEvaluator) downloadAndInspectPolicies(ctx context.Context) error {
```
st3penta, line 132: `downloadAndInspectPolicies` has ~80 lines of complex logic — policy download, annotation inspection, GitHub/GitLab error URL rewriting (4 nested conditions), and rule collection — with no direct unit tests. This is the shared infrastructure both evaluators depend on, so bugs here have wide blast radius.

Unit tests for annotation parsing (annotated vs bare rules) and URL rewriting (GitHub, GitLab, raw, non-matching) would be high-value.

internal/image/validate.go
```diff
@@ -112,7 +112,13 @@ func ValidateImage(ctx context.Context, comp app.SnapshotComponent, snap *app.Sn
 		return out, nil
 	}
 
-	inputPath, inputJSON, err := a.WriteInputFile(ctx)
+	inputMap, inputJSON, err := a.BuildInput(ctx)
```
st3penta, line 115: `BuildInput()` and `WriteInputFile()` are both called unconditionally, each independently constructing and marshaling the same `Input` struct. On the OPA path the disk write is wasted; on conftest the in-memory build is wasted.

The two evaluators also consume different input sources from `EvaluationTarget` (`ParsedInput` vs `Inputs` file path). Conftest silently ignores `ParsedInput`, making the contract implicit. If the two representations ever diverge (marshal differences, TOCTOU on the file), results will differ silently.

Unifying the input path — both from memory or both from disk — would eliminate the duplicate work and the dual-source contract.

internal/evaluator/opa_evaluator.go
```diff
@@ -17,33 +17,374 @@
...
+	if name == "violation" || name == "deny" || name == "warn" {
+		return ""
+	}
+	name = strings.TrimPrefix(name, "violation_")
+	name = strings.TrimPrefix(name, "deny_")
+	name = strings.TrimPrefix(name, "warn_")
+	return name
+}
+
+func (o *opaEvaluator) queryNamespace(ctx context.Context, fileName string, input any, namespace string) (Outcome, error) {
```
st3penta, line 237: `queryNamespace` is 70+ lines mixing five concerns: rule iteration, dedup, exception querying, rule evaluation, result categorization, and success counting. Extracting per-rule evaluation (exception check + rule query + categorization) into a helper would make the logic easier to follow and test independently.

acceptance/cli/cli.go
```diff
@@ -859,6 +859,12 @@ func AddStepsTo(sc *godog.ScenarioContext) {
 	sc.Step(`^the "([^"]*)" file should match the snapshot$`, matchFileSnapshot)
 	sc.Step(`^a file named "([^"]*)" containing$`, createGenericFile)
 	sc.Step(`^a track bundle file named "([^"]*)" containing$`, createTrackBundleFile)
+	sc.Before(func(ctx context.Context, _ *godog.Scenario) (context.Context, error) {
+		if v := os.Getenv("EC_USE_OPA"); v != "" {
+			ctx, _ = theEnvironmentVarilableIsSet(ctx, "EC_USE_OPA="+v)
```
st3penta: The error from `theEnvironmentVarilableIsSet` is discarded. If this call fails, `EC_USE_OPA` is never set and the OPA acceptance tests silently run against conftest — passing without testing the OPA path at all.

```suggestion
			ctx, err := theEnvironmentVarilableIsSet(ctx, "EC_USE_OPA="+v)
			if err != nil {
				return ctx, fmt.Errorf("setting EC_USE_OPA: %w", err)
			}
```

internal/evaluator/opa_evaluator_test.go
```diff
@@ -135,20 +126,828 @@ func TestCapabilitiesPath(t *testing.T) {
...
+		results, err := o.evaluateWithEngine(context.Background(), target, nil)
+		require.NoError(t, err)
+		require.Len(t, results, 1)
+		assert.Len(t, results[0].Failures, 1)
+	})
+}
+
+func TestEnsureInitialized(t *testing.T) {
+	t.Run("returns error from init", func(t *testing.T) {
+		expectedErr := fmt.Errorf("init failed")
```
st3penta: The "returns error from init" subtest sets `expectedErr` and `o.initErr`, then immediately overrides both — `o.initOnce` gets a fresh `sync.Once` and `o.initErr` is reset to nil. `expectedErr` is never read again. The test ends up exercising a filesystem error (no policy sources), not the intended "pre-set error" scenario.

Either remove `expectedErr` and rename the test to match what it actually covers, or fix it to assert on the pre-set error.

st3penta (commented):
I left a few comments, AI-found but human-reviewed

internal/evaluator/opa_evaluator.go
joejstuart: Good catch — added a `default` case that surfaces unexpected types as a result message with the type and value, so they're visible rather than silently dropped. Also added a test case for it.

internal/evaluator/opa_evaluator.go
joejstuart: Fixed — changed to `==` for case-sensitive comparison.

internal/evaluator/opa_evaluator.go
joejstuart: Fixed — switched to `len(ruleNames)` and removed the now-unused `ruleCount` variable.

features/validate_input_opa.feature
joejstuart: Good find — wired `IsOpaEnabled()` into `internal/evaluation_target/input/input.go` so `NewInput()` now uses `NewOPAEvaluator` when `EC_USE_OPA=1` is set, matching the image validation path.

acceptance/cli/cli.go
joejstuart: Fixed — the error is now checked and returned, wrapped with context.

internal/evaluator/opa_evaluator_test.go
joejstuart: Fixed — removed the dead `expectedErr` code and renamed the test to "returns error when no policy sources" to match what it actually exercises.

internal/evaluator/opa_evaluator.go
joejstuart: Agree this is worth tracking. The current behavior matches conftest's semantics — exceptions are recorded in `outcome.Exceptions` (line 304) but the original rule result message isn't preserved there. No current conforma policies use the conftest exception mechanism, so this is latent. I'll file a follow-up to add the original message to the exception entry when we revisit exception handling.

internal/evaluator/base_evaluator.go
joejstuart: Fair point — the PR description overstates the current state. `basePolicyEvaluator` is used by `opaEvaluator` now, and migrating `conftestEvaluator` to embed it is the intended next step. That refactor touches a lot of the conftest evaluator's internals so I kept it out of this PR to limit blast radius. I'll update the PR description to be clearer about the scope.

internal/evaluation_target/application_snapshot_image/application_snapshot_image.go
joejstuart: Added `TestBuildInput` with 4 subtests covering the round-trip shape, parent ref inclusion/exclusion, and parity with `WriteInputFile` output.

internal/evaluator/base_evaluator.go
joejstuart: This logic was moved from the conftest evaluator where it had the same coverage gap. The existing integration and acceptance tests exercise it end-to-end, but I agree direct unit tests for annotation parsing and URL rewriting would be high-value. I'll add those in a follow-up.

internal/image/validate.go
joejstuart: Intentional for now — both evaluator types need their respective input format, and the cost of the extra marshal is negligible compared to policy download and evaluation. Unifying to a single input path requires changing the conftest runner to accept parsed input, which is a larger change. Agreed it's worth doing eventually to eliminate the dual-source contract risk.

internal/evaluator/opa_evaluator.go
joejstuart: Done — extracted `collectRuleNames` (rule iteration + dedup) and `evaluateRule` (exception check + rule query + categorization) out of `queryNamespace`. The orchestrator now just loops over rules and accumulates results. Added `TestCollectRuleNames` and `TestEvaluateRule` to test the helpers independently.

st3penta (dismissed):
LGTM

I left the comments that need a followup unresolved, so you can track them and then close them yourself.

I think you have to regenerate the snapshots to get the acceptance tests green, ping me for a rubber stamp once you do it.

internal/evaluator/opa_evaluator.go
joejstuart: I just logged the exception query. We don't use exception rules anyways.

internal/evaluator/base_evaluator.go
joejstuart: I feel like I can make a story for this and it'll never get done. There's probably multiple functions that could use a refactor. wdyt? I'd like to just resolve this.

st3penta (dismissed):

internal/evaluator/base_evaluator.go
```diff
@@ -0,0 +1,426 @@
...
+				ext := filepath.Ext(d.Name())
+				if ext == ".json" || ext == ".yaml" || ext == ".yml" {
+					dirsWithDataFiles[filepath.Dir(p)] = true
+				}
+			}
+			return nil
+		})
+		if err != nil {
+			continue
+		}
```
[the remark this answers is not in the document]
joejstuart: Fixed

joejstuart:
@st3penta PTAL again. I had to regenerate the docs.

st3penta (approved):
LGTM

PR #3277, opened 2026-05-07
Fix UPDATE_SNAPS support in make acceptance

st3penta:
The go-snaps v0.5.19 upgrade broke UPDATE_SNAPS=true for acceptance
tests. The library's Clean() function requires snapshot entry headers
to start with "[Test", but errCapture.Name() returned bare scenario
names. In v0.5.7 this was harmless because Clean() only truncated
files when obsolete entries were found; in v0.5.19 it always truncates
when UPDATE_SNAPS is set, then rewrites from parsed entries — since
nothing parsed, every snap file was emptied.

- Prefix errCapture.Name() with "TestFeatures/" so Clean() can parse
  the entries, and update all existing snapshot headers to match
- Fix the acceptance Makefile target to propagate snap file deletions
  back from the temp workdir when UPDATE_SNAPS is set

Ref: https://redhat.atlassian.net/browse/EC-1813

### DIFF
--- a/features/__snapshots__/validate_image.snap
+++ b/features/__snapshots__/validate_image.snap
@@ -1,4 +1,5 @@
-[happy day with git config and yaml:stdout - 1]
+
+[TestFeatures/happy day with git config and yaml:stdout - 1]
 {
   "success": true,
   "components": [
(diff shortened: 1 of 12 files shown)


joejstuart (approved):

features/__snapshots__/validate_image.snap
```diff
@@ -3680,1163 +3608,11 @@ Error: success criteria not met
 }
 ---
 
-[sigstore functions:stderr - 1]
```
simonbaird, line 3683: Is it some stale snapshot data that should have been removed a while ago?

simonbaird (approved):
Good stuff.

features/__snapshots__/validate_image.snap
st3penta: yes exactly, i double checked that all the removed stuff is from scenarios that aren't there anymore

PR #3311, opened 2026-05-14
feat(EC-1829): include annotations in OCI referrers response

joejstuart:
## Summary
- Adds `annotations` field (map[string]string) to the `ec.oci.image_referrers` Rego builtin response
- Enables Rego policies to access manifest annotations on referrer descriptors, allowing filtering/sorting of duplicate referrers (e.g., by timestamp annotation)
- Uses the existing `newAnnotationsTerm()` helper, consistent with how annotations are handled in other OCI builtins

## Changes
- `internal/rego/oci/oci.go`: Added annotations to the descriptor type declaration and response building
- `internal/rego/oci/oci_test.go`: Added annotations field presence assertion and dedicated structural test

## Test plan
- [x] All existing `TestOCIImageReferrers` subtests pass
- [x] New `annotations field is an object` subtest validates annotations are returned as an object on every descriptor
- [x] Full `internal/rego/oci` unit test suite passes with no regressions

Jira: EC-1829

### DIFF
--- a/docs/modules/ROOT/pages/ec_oci_image_referrers.adoc
+++ b/docs/modules/ROOT/pages/ec_oci_image_referrers.adoc
@@ -12,4 +12,4 @@ Discover artifacts attached to an image via OCI Referrers API.
 
 == Return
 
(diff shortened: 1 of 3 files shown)


robnester-rh (approved):
LGTM

PR #3342, opened 2026-06-17
fix: handle digest-pinned policy refs in pin-konflux-policy-bundle.sh

joejstuart:
## Summary

- Fix `pin-konflux-policy-bundle.sh` to handle policy references that already contain a pinned `@sha256:` digest, preventing invalid double-digest OCI references (e.g. `@sha256:new@sha256:old`)
- Add test script (`hack/pin-konflux-policy-bundle_test.sh`) with 6 scenarios and 10 assertions covering tag-only, tag+digest, unrelated refs, empty digest, bare hex normalization, and inline JSON input modes

## Problem

`EnterpriseContractPolicy` resources can contain digest-pinned policy references like:
```
oci::quay.io/conforma/release-policy:konflux@sha256:6eb386f...
```

The `step-pin-policy-bundle` step's sed pattern only matched `:konflux` and replaced it with `@sha256:new`, but left the existing `@sha256:old` suffix in place, producing:
```
oci::quay.io/conforma/release-policy@sha256:new@sha256:old
```

This caused `ec validate` to fail with `invalid checksum digest length` — observed in [konflux-ci/release-service#1678](https://github.com/konflux-ci/release-service/pull/1678).

## Fix

The sed pattern now optionally matches and consumes a trailing `@sha256:<hex>` digest:
```diff
-sed -i "s|${ORIGINAL}|${REPLACEMENT}|g" "$WORKING_POLICY"
+sed -i "s|${ORIGINAL}\(@sha256:[a-f0-9]*\)\{0,1\}|${REPLACEMENT}|g" "$WORKING_POLICY"
```

## Test plan

- [x] `bash hack/pin-konflux-policy-bundle_test.sh` — all 10 assertions pass
- [ ] Verify fix resolves the e2e failure in release-service PR #1678

https://redhat.atlassian.net/browse/EC-1909

### DIFF
--- a/hack/pin-konflux-policy-bundle.sh
+++ b/hack/pin-konflux-policy-bundle.sh
@@ -23,6 +23,8 @@
 # - POLICY_BUNDLE_DIGEST: OCI digest to pin to (e.g. "sha256:abc123..." or "abc123...").
 #   If empty, the script is a no-op.
 # - HOME: Home directory (defaults to /tekton/home in Tekton tasks).
+#
+# Tests: bash hack/pin-konflux-policy-bundle_test.sh
 
 set -o errexit
 set -o nounset
@@ -78,6 +80,13 @@ REPLACEMENT="oci::quay.io/conforma/release-policy@${POLICY_BUNDLE_DIGEST}"
 
 if ! grep -q "${ORIGINAL}" "$WORKING_POLICY"; then
   echo "'${ORIGINAL}' not found in policy configuration, nothing to do."
+  exit 0
+fi
+
+# If the reference is already digest-pinned (e.g. :konflux@sha256:...),
+# respect the existing pin rather than overriding it.
+if grep -q "${ORIGINAL}@sha256:" "$WORKING_POLICY"; then
+  echo "'${ORIGINAL}' already has a pinned digest, skipping override."
   exit 0
 fi
 
--- a/hack/pin-konflux-policy-bundle_test.sh
+++ b/hack/pin-konflux-policy-bundle_test.sh
@@ -1,0 +1,231 @@
+#!/usr/bin/env bash
+# Copyright The Conforma Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Tests for pin-konflux-policy-bundle.sh
+#
+# Run: bash hack/pin-konflux-policy-bundle_test.sh
+
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPT="${SCRIPT_DIR}/pin-konflux-policy-bundle.sh"
+
+# On macOS, sed -i behaves differently. Use GNU sed if available.
+if [[ "$(uname)" == "Darwin" ]] && command -v gsed &>/dev/null; then
+  mkdir -p /tmp/pin-test-bin
+  ln -sf "$(command -v gsed)" /tmp/pin-test-bin/sed
+  export PATH="/tmp/pin-test-bin:${PATH}"
+fi
+
+PASS=0
+FAIL=0
+
+assert_contains() {
+  local file="$1" expected="$2" label="$3"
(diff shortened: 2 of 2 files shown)


hack/pin-konflux-policy-bundle.sh
```diff
@@ -81,7 +83,9 @@ if ! grep -q "${ORIGINAL}" "$WORKING_POLICY"; then
   exit 0
 fi
 
-sed -i "s|${ORIGINAL}|${REPLACEMENT}|g" "$WORKING_POLICY"
+# Match the tag and any existing @sha256:... digest suffix so we don't
+# produce a double-digest reference like ...@sha256:new@sha256:old.
+sed -i "s|${ORIGINAL}\(@sha256:[a-f0-9]*\)\{0,1\}|${REPLACEMENT}|g" "$WORKING_POLICY"
```
simonbaird: Hmm I feel like maybe we should leave it alone entirely if there is already a digest. I think it would be weird to replace an already pinned reference with another different pin.

joejstuart: Done — updated to skip replacement entirely when the reference already has a digest. The script now exits early with a message when it finds `:konflux@sha256:` in the policy config.

hack/pin-konflux-policy-bundle.sh
```diff
@@ -81,7 +83,9 @@ if ! grep -q "${ORIGINAL}" "$WORKING_POLICY"; then
   exit 0
 fi
```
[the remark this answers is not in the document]
joejstuart: No longer applicable — the complex sed regex was removed. The current sed only matches the literal `:konflux` string (tag-only refs), since the new guard exits before sed runs when a digest is present.

hack/pin-konflux-policy-bundle_test.sh
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
joejstuart: Added as Test 8 — verifies clean early exit when the file has no `:konflux` tag (only `@sha256:old` remains from a previous run).

hack/pin-konflux-policy-bundle.sh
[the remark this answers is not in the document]
joejstuart: Agreed this is valid but it's pre-existing and not introduced by this PR. The `POLICY_BUNDLE_DIGEST` param is set by Tekton task infrastructure, not user input. Happy to add digest validation as a follow-up.

hack/pin-konflux-policy-bundle.sh
```diff
@@ -81,6 +83,13 @@ if ! grep -q "${ORIGINAL}" "$WORKING_POLICY"; then
...
 
+# If the reference is already digest-pinned (e.g. :konflux@sha256:...),
+# respect the existing pin rather than overriding it.
+if grep -q "${ORIGINAL}@sha256:" "$WORKING_POLICY"; then
+  echo "'${ORIGINAL}' already has a pinned digest, skipping override."
+  exit 0
+fi
```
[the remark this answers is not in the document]
joejstuart: This scenario (mixed pinned and unpinned `release-policy:konflux` refs in the same policy file) doesn't occur in practice — each ECP `sources[].policy` array has a single release-policy entry. The early exit is intentional per the reviewer's feedback: if someone explicitly pinned the reference with a digest, that's a deliberate choice and the task shouldn't override it.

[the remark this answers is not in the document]
simonbaird: TBH it would be possible for an ECP to include multiple policy sources, and hence one could be pinned and one could use the tag only. In that scenario, pinning the tag only ref, but leaving the pre-pinned one alone would make sense. But I think there are zero real world examples of that, so it's an edge case we can ignore for now.

hack/pin-konflux-policy-bundle_test.sh
```diff
@@ -0,0 +1,231 @@
...
+  (( FAIL++ ))
+fi
+rm -rf "$TMPDIR"
+
+# -------------------------------------------------------------------
+echo ""
+echo "Results: ${PASS} passed, ${FAIL} failed"
+if (( FAIL > 0 )); then
+  exit 1
+fi
```
simonbaird, line 231: FYI there is some coverage in the acceptance tests for the behavior, so you could also add a test case there if you want to.

simonbaird (approved):

hack/pin-konflux-policy-bundle.sh
simonbaird, line 78: I was thinking we could just change this to:


```suggestion
ORIGINAL="oci::quay.io/conforma/release-policy:konflux$"
```

then make sure the grep uses the regex end of line marker.

simonbaird: (Not a suggestion or a blocker, just fyi)

simonbaird: Not sure if we'd need to worry about trailing spaces in this implementation.

joejstuart: yeah, I might leave it. We could probably make something work, but just adding `$` breaks some tests.

PR #3386, opened 2026-07-06
Add validate input --server flag for persistent HTTP/REST

simonbaird:
This adds an `ec validate input --server` command that starts up a web service that responds on `/v1/validate/input` with the results the same as if you ran `ec validate input --output json` with the request body as the input file

Ref: https://redhat.atlassian.net/browse/EC-1882

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -227,15 +227,15 @@ LINT_TO_GITHUB_ANNOTATIONS='map(map(.)[])[][] as $$d | $$d.posn | split(":") as 
 .PHONY: lint
 lint: tekton-lint go-mod-lint ## Run linter
 # addlicense doesn't give us a nice explanation so we prefix it with one
-	@go run -modfile tools/go.mod github.com/google/addlicense -c '$(COPY)' -y '' -s -check $(LICENSE_IGNORE) . | sed 's/^/Missing license header in: /g'
+	@git ls-files -z | xargs -0 go run -modfile tools/go.mod github.com/google/addlicense -c '$(COPY)' -y '' -s -check $(LICENSE_IGNORE) | sed 's/^/Missing license header in: /g'
 # piping to sed above looses the exit code, luckily addlicense is fast so we invoke it for the second time to exit 1 in case of issues
-	@go run -modfile tools/go.mod github.com/google/addlicense -c '$(COPY)' -y '' -s -check $(LICENSE_IGNORE) . >/dev/null 2>&1
+	@git ls-files -z | xargs -0 go run -modfile tools/go.mod github.com/google/addlicense -c '$(COPY)' -y '' -s -check $(LICENSE_IGNORE) >/dev/null 2>&1
 	@go run -modfile tools/go.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run $(if $(GITHUB_ACTIONS), --timeout=10m0s)
 	@(cd acceptance && go run -modfile ../tools/go.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run --path-prefix acceptance $(if $(GITHUB_ACTIONS), --timeout=10m0s))
 
 .PHONY: lint-fix
 lint-fix: ## Fix linting issues automagically
-	@go run -modfile tools/go.mod github.com/google/addlicense -c '$(COPY)' -y '' -s $(LICENSE_IGNORE) .
+	@git ls-files -z | xargs -0 go run -modfile tools/go.mod github.com/google/addlicense -c '$(COPY)' -y '' -s $(LICENSE_IGNORE)
 	@go run -modfile tools/go.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run --fix
 	@(cd acceptance && go run -modfile ../tools/go.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run --path-prefix acceptance --fix)
 # We don't apply the fixes from the internal (error handling) linter.
--- a/acceptance/cli/server.go
+++ b/acceptance/cli/server.go
@@ -162,98 +162,65 @@ func getServerState(ctx context.Context) (*serverState, error) {
 	return state, nil
 }
 
-func aGETRequestIsSentToTheServer(ctx context.Context, urlPath string) (context.Context, error) {
+func doRequest(ctx context.Context, method, urlPath, contentType string, body io.Reader) (context.Context, error) {
 	state, err := getServerState(ctx)
 	if err != nil {
 		return ctx, err
 	}
 
 	url := fmt.Sprintf("http://127.0.0.1:%d%s", state.port, urlPath)
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	req, err := http.NewRequestWithContext(ctx, method, url, body)
 	if err != nil {
-		return ctx, fmt.Errorf("GET %s: %w", urlPath, err)
+		return ctx, fmt.Errorf("%s %s: %w", method, urlPath, err)
+	}
+	if contentType != "" {
+		req.Header.Set("Content-Type", contentType)
 	}
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return ctx, fmt.Errorf("GET %s: %w", urlPath, err)
+		return ctx, fmt.Errorf("%s %s: %w", method, urlPath, err)
 	}
 	defer resp.Body.Close()
 
-	body, err := io.ReadAll(resp.Body)
+	respBody, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return ctx, fmt.Errorf("reading response: %w", err)
 	}
 
 	return context.WithValue(ctx, httpResponseKey, &httpResponse{
 		statusCode: resp.StatusCode,
-		body:       string(body),
+		body:       string(respBody),
 	}), nil
+}
+
+func expandBody(ctx context.Context, content *godog.DocString) (string, error) {
+	state, err := getServerState(ctx)
+	if err != nil {
+		return "", err
+	}
+	return os.Expand(content.Content, func(key string) string {
+		return state.vars[key]
+	}), nil
+}
+
+func aGETRequestIsSentToTheServer(ctx context.Context, urlPath string) (context.Context, error) {
+	return doRequest(ctx, http.MethodGet, urlPath, "", nil)
 }
 
 func aPOSTRequestIsSentToTheServerWithBody(ctx context.Context, urlPath string, content *godog.DocString) (context.Context, error) {
-	state, err := getServerState(ctx)
+	body, err := expandBody(ctx, content)
 	if err != nil {
 		return ctx, err
 	}
-
-	body := os.Expand(content.Content, func(key string) string {
-		return state.vars[key]
-	})
-
-	url := fmt.Sprintf("http://127.0.0.1:%d%s", state.port, urlPath)
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, strings.NewReader(body))
-	if err != nil {
-		return ctx, fmt.Errorf("POST %s: %w", urlPath, err)
-	}
-	req.Header.Set("Content-Type", "application/json")
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return ctx, fmt.Errorf("POST %s: %w", urlPath, err)
-	}
-	defer resp.Body.Close()
-
-	respBody, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return ctx, fmt.Errorf("reading response: %w", err)
-	}
-
-	return context.WithValue(ctx, httpResponseKey, &httpResponse{
-		statusCode: resp.StatusCode,
-		body:       string(respBody),
-	}), nil
+	return doRequest(ctx, http.MethodPost, urlPath, "application/json", strings.NewReader(body))
 }
 
 func aPOSTRequestIsSentToTheServerWithContentTypeAndBody(ctx context.Context, urlPath, contentType string, content *godog.DocString) (context.Context, error) {
-	state, err := getServerState(ctx)
+	body, err := expandBody(ctx, content)
 	if err != nil {
 		return ctx, err
 	}
-
-	body := os.Expand(content.Content, func(key string) string {
-		return state.vars[key]
-	})
-
-	url := fmt.Sprintf("http://127.0.0.1:%d%s", state.port, urlPath)
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, strings.NewReader(body))
-	if err != nil {
-		return ctx, fmt.Errorf("POST %s: %w", urlPath, err)
-	}
-	req.Header.Set("Content-Type", contentType)
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return ctx, fmt.Errorf("POST %s: %w", urlPath, err)
-	}
-	defer resp.Body.Close()
-
-	respBody, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return ctx, fmt.Errorf("reading response: %w", err)
-	}
-
-	return context.WithValue(ctx, httpResponseKey, &httpResponse{
-		statusCode: resp.StatusCode,
-		body:       string(respBody),
-	}), nil
+	return doRequest(ctx, http.MethodPost, urlPath, contentType, strings.NewReader(body))
 }
 
 func aPOSTRequestIsSentToTheServerWithFile(ctx context.Context, urlPath, filePath string) (context.Context, error) {
@@ -271,27 +238,7 @@ func aPOSTRequestIsSentToTheServerWithFile(ctx context.Context, urlPath, filePat
 		return ctx, fmt.Errorf("reading file %s: %w", expanded, err)
 	}
 
-	url := fmt.Sprintf("http://127.0.0.1:%d%s", state.port, urlPath)
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, strings.NewReader(string(data)))
-	if err != nil {
-		return ctx, fmt.Errorf("POST %s: %w", urlPath, err)
-	}
-	req.Header.Set("Content-Type", "application/yaml")
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return ctx, fmt.Errorf("POST %s: %w", urlPath, err)
-	}
-	defer resp.Body.Close()
-
-	respBody, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return ctx, fmt.Errorf("reading response: %w", err)
-	}
-
-	return context.WithValue(ctx, httpResponseKey, &httpResponse{
-		statusCode: resp.StatusCode,
-		body:       string(respBody),
-	}), nil
+	return doRequest(ctx, http.MethodPost, urlPath, "application/yaml", strings.NewReader(string(data)))
 }
 
 func getResponse(ctx context.Context) (*httpResponse, error) {
--- a/acceptance/cli/server.go
+++ b/acceptance/cli/server.go
@@ -223,6 +223,39 @@ func aPOSTRequestIsSentToTheServerWithBody(ctx context.Context, urlPath string, 
 	}), nil
 }
 
+func aPOSTRequestIsSentToTheServerWithContentTypeAndBody(ctx context.Context, urlPath, contentType string, content *godog.DocString) (context.Context, error) {
+	state, err := getServerState(ctx)
+	if err != nil {
+		return ctx, err
+	}
+
+	body := os.Expand(content.Content, func(key string) string {
+		return state.vars[key]
+	})
+
+	url := fmt.Sprintf("http://127.0.0.1:%d%s", state.port, urlPath)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, strings.NewReader(body))
+	if err != nil {
+		return ctx, fmt.Errorf("POST %s: %w", urlPath, err)
+	}
+	req.Header.Set("Content-Type", contentType)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return ctx, fmt.Errorf("POST %s: %w", urlPath, err)
+	}
+	defer resp.Body.Close()
+
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return ctx, fmt.Errorf("reading response: %w", err)
+	}
+
+	return context.WithValue(ctx, httpResponseKey, &httpResponse{
+		statusCode: resp.StatusCode,
+		body:       string(respBody),
+	}), nil
+}
+
 func aPOSTRequestIsSentToTheServerWithFile(ctx context.Context, urlPath, filePath string) (context.Context, error) {
 	state, err := getServerState(ctx)
 	if err != nil {
@@ -330,6 +363,7 @@ func AddServerStepsTo(sc *godog.ScenarioContext) {
 	sc.Step(`^ec server is started with "(.+)"$`, ecServerStartedWith)
 	sc.Step(`^a GET request is sent to the server at "(.+)"$`, aGETRequestIsSentToTheServer)
 	sc.Step(`^a POST request is sent to the server at "(.+)" with body$`, aPOSTRequestIsSentToTheServerWithBody)
+	sc.Step(`^a POST request is sent to the server at "(.+)" with content type "(.+)" and body$`, aPOSTRequestIsSentToTheServerWithContentTypeAndBody)
 	sc.Step(`^a POST request is sent to the server at "(.+)" with file "(.+)"$`, aPOSTRequestIsSentToTheServerWithFile)
 	sc.Step(`^the response status should be (\d+)$`, theResponseStatusShouldBe)
 	sc.Step(`^the response should contain "(.+)"$`, theResponseShouldContain)
--- a/acceptance/cli/server.go
+++ b/acceptance/cli/server.go
@@ -1,0 +1,340 @@
+// Copyright The Conforma Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package cli
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"path"
+	"runtime"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/cucumber/godog"
+
+	log "github.com/conforma/cli/acceptance/log"
+	"github.com/conforma/cli/acceptance/snaps"
+)
+
+type serverStateKey int
+
+const (
+	serverKey serverStateKey = iota
+)
+
+type serverState struct {
+	cmd    *exec.Cmd
+	port   int
+	stdout *bytes.Buffer
+	stderr *bytes.Buffer
+	vars   map[string]string
+}
+
+type httpResponse struct {
+	statusCode int
+	body       string
+}
+
+type responseKey int
+
+const (
+	httpResponseKey responseKey = iota
+)
+
+func ecServerStartedWith(ctx context.Context, parameters string) (context.Context, error) {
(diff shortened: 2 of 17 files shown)


cmd/validate/input.go
```diff
@@ -107,6 +143,23 @@ func validateInputCmd(validate InputValidationFunc) *cobra.Command {
...
+
+				srv := server.New(server.Config{
+					Port:               data.serverPort,
+					Policy:             data.policy,
+					Info:               data.info,
+					Workers:            data.workers,
+					ShowSuccesses:      showSuccesses,
+					ShowWarnings:       showWarnings,
+					ShowPolicyDocsLink: showPolicyDocsLink,
+				})
```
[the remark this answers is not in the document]
simonbaird: Resolved in an upcoming revision.

internal/server/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	var allResults []evaluator.Outcome
+	for _, e := range s.evaluators {
+		results, err := e.Evaluate(ctx, evaluator.EvaluationTarget{Inputs: []string{tmpPath}})
+		if err != nil {
+			log.WithField("error", err).Error("Evaluation failed")
+			writeError(w, http.StatusInternalServerError, fmt.Sprintf("evaluation failed: %v", err))
+			return
+		}
+		allResults = append(allResults, results...)
+	}
```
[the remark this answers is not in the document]
simonbaird: Adding a timeout, will push it soon.

internal/server/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	ct := r.Header.Get("Content-Type")
+	switch ct {
+	case "application/json":
+		return utils.IsJson(string(body))
+	case "application/yaml", "application/x-yaml", "text/yaml":
+		return true
+	default:
+		return utils.IsJson(string(body)) || utils.IsYamlMap(string(body))
+	}
+}
```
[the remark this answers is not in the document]
simonbaird: Fixed in upcoming revision.

internal/server/server.go
```diff
@@ -0,0 +1,149 @@
...
+	mux := http.NewServeMux()
+	s.registerRoutes(mux)
+
+	handler := requestLoggingMiddleware(recoveryMiddleware(mux))
+
+	httpServer := &http.Server{
+		Addr:              fmt.Sprintf(":%d", s.cfg.Port),
+		Handler:           handler,
+		ReadHeaderTimeout: 10 * time.Second,
+	}
```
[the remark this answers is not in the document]
simonbaird: Fixed in an upcoming revision.

cmd/validate/input.go
```diff
@@ -107,6 +143,23 @@ func validateInputCmd(validate InputValidationFunc) *cobra.Command {
...
+				return srv.Start(cmd.Context())
+			}
+
```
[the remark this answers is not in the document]
simonbaird: Fix incoming.

internal/server/server.go
```diff
@@ -0,0 +1,149 @@
...
+
+type Config struct {
+	Port               int
+	Policy             policy.Policy
+	Info               bool
+	Workers            int
+	ShowSuccesses      bool
+	ShowWarnings       bool
+	ShowPolicyDocsLink bool
+}
```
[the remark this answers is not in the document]
simonbaird: Workers param is being removed. I think that's good enough for now.

internal/server/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	tmpFile.Close()
+
+	var allResults []evaluator.Outcome
+	for _, e := range s.evaluators {
+		results, err := e.Evaluate(ctx, evaluator.EvaluationTarget{Inputs: []string{tmpPath}})
+		if err != nil {
+			log.WithField("error", err).Error("Evaluation failed")
+			writeError(w, http.StatusInternalServerError, fmt.Sprintf("evaluation failed: %v", err))
+			return
```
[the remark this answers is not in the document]
simonbaird: Fixed in revision that will be pushed soon.

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+			log.WithField("error", err).Error("Evaluation failed")
+			// Avoid leaking internal details (file paths, engine errors) to API clients.
+			msg := "evaluation failed"
+			if evalCtx.Err() == context.DeadlineExceeded {
+				msg = "evaluation timed out"
+			}
+			writeError(w, http.StatusInternalServerError, msg)
+			return
+		}
+		allResults = append(allResults, results...)
```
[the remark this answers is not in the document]
simonbaird: The workdir is effectively readonly once it is created, so we think there's no risk of problems here. Added a comment to explain the reasoning (will be pushed in next revision).

internal/server/server.go
```diff
@@ -0,0 +1,151 @@
...
+	s.registerRoutes(mux)
+
+	handler := requestLoggingMiddleware(recoveryMiddleware(mux))
+
+	httpServer := &http.Server{
+		Addr:              fmt.Sprintf(":%d", s.cfg.Port),
+		Handler:           handler,
+		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      120 * time.Second,
```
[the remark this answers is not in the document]
simonbaird: It's a reasonable suggestion, but let's not worry about it for now. Added a todo comment just to note it. Don't want to file a Jira unless we decide it's important to implement soon.

internal/server/server.go
```diff
@@ -0,0 +1,151 @@
...
+func New(cfg Config) *Server {
+	return &Server{cfg: cfg}
+}
+
+func (s *Server) Start(ctx context.Context) error {
+	// The CLI defaults to Warn level, but a long-running server needs Info
+	// for startup, shutdown, and per-request logging. Don't downgrade if
+	// the user already requested a more verbose level (--debug/--verbose).
+	if log.GetLevel() < log.InfoLevel {
+		log.SetLevel(log.InfoLevel)
```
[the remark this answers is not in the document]
simonbaird: Added a comment.

acceptance/cli/server.go
```diff
@@ -0,0 +1,354 @@
...
+		return ctx, err
+	}
+
+	vars["SERVER_PORT"] = fmt.Sprintf("%d", port)
+
+	args := os.Expand(parameters, func(key string) string {
+		return vars[key]
+	})
+
+	cmd := exec.Command(ec)
```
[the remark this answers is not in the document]
simonbaird: Noted.

simonbaird:
Last change is adding explanatory comments to address the potential concerns.

simonbaird:
I've addressed a few more agentic review suggestions and nitpicks.

simonbaird:
Claude Opus 4.6's response to the remaining nitpick:

> The 200ms timeout isn't a race — it's a minimum run time. The server starts in single-digit milliseconds (no-op policy, port 0), then the timeout fires and triggers graceful shutdown. The timeout would need to
  be too short for the server to start, not too short for a sleep to fire. If CI is so slow that binding a localhost port takes 200ms, we have bigger problems. Bumping it to e.g. 1s "just in case" would just make the test slower for no real benefit.

simonbaird:
The e2e fail log includes this, so maybe a flake:
```
  [TIMEDOUT] A suite timeout occurred
  In [It] at: /tmp/tmp.pg5FZDV85w/e2e-tests/tests/contract/contract.go:213 @ 07/07/26 21:43:16.638
```

simonbaird:
I rebased, fixed some conflicts, squashed the "add a comment to explain a decision" commits.

internal/evaluator/evaluator.go
```diff
@@ -27,6 +27,10 @@ type EvaluationTarget struct {
 	ParsedInput   map[string]any
 }
 
+// Evaluator implementations must be safe for concurrent use. The server (when
```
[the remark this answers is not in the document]
simonbaird: Reasonable idea, but I don't think it's high value enough to implement it here.

Makefile
```diff
@@ -227,15 +227,15 @@ LINT_TO_GITHUB_ANNOTATIONS='map(map(.)[])[][] as $$d | $$d.posn | split(":") as
 .PHONY: lint
 lint: tekton-lint go-mod-lint ## Run linter
 # addlicense doesn't give us a nice explanation so we prefix it with one
-	@go run -modfile tools/go.mod github.com/google/addlicense -c '$(COPY)' -y '' -s -check $(LICENSE_IGNORE) . | sed 's/^/Missing license header in: /g'
+	@git ls-files -z | xargs -0 go run -modfile tools/go.mod github.com/google/addlicense -c '$(COPY)' -y '' -s -check $(LICENSE_IGNORE) | sed 's/^/Missing license header in: /g'
```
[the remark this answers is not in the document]
simonbaird: This is true, but since our CI is quite slow and expensive, I choose to be lazy and just included it in this PR. Besides, I'm comfortable reverting commit by commit as needed, rather than needing to revert an entire PR atomically (should it be required).

cmd/validate/input.go
```diff
@@ -310,6 +361,20 @@ func validateInputCmd(validate InputValidationFunc) *cobra.Command {
 		panic(err)
 	}
 
+	cmd.Flags().BoolVar(&data.serverMode, "server", false, hd.Doc(`
```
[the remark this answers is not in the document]
simonbaird: Good catch, fixed in next revision.

internal/server/server_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+type mockEvaluator struct {
+	results []evaluator.Outcome
+	err     error
+	target  evaluator.EvaluationTarget
+}
+
+func (m *mockEvaluator) Evaluate(_ context.Context, target evaluator.EvaluationTarget) ([]evaluator.Outcome, error) {
+	m.target = target
+	return m.results, m.err
```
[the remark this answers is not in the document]
simonbaird: Fixed in next revision.

simonbaird:
I'm ignoring the two CodeRabbit nitpicks about the license check change. Made changes for a couple of fullsend's fresh nitpicks.

joejstuart (dismissed):
LGTM

joejstuart:
Tested paths 

- POST /v1/validate/input — passing JSON input
- POST /v1/validate/input — passing YAML input
- POST /v1/validate/input — input with violations
- POST /v1/validate/input — empty body
- POST /v1/validate/input — invalid (not JSON or YAML)
- GET /v1/validate/input — wrong HTTP method
- GET /live — liveness probe
- GET /ready — readiness probe

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+		writeError(w, http.StatusInternalServerError, "failed to write temp file")
+		return
+	}
+	tmpFile.Close()
+
+	evalCtx, evalCancel := context.WithTimeout(ctx, evaluationTimeout)
+	defer evalCancel()
+
+	var allResults []evaluator.Outcome
+	for _, e := range s.evaluators {
```
st3penta, line 113: If `s.evaluators` is nil or empty, this loop is a no-op and the handler returns HTTP 200 with `success: true` and zero violations. A misconfigured policy spec with zero sources would silently approve every input indefinitely.

Worth adding a guard here (or in `Start()`) that returns an error when there are no evaluators, rather than running a vacuous evaluation.

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+
+	tmpFile, err := os.CreateTemp("", "ec-server-input-*"+inputExtension(r, body))
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, "failed to create temp file")
+		return
+	}
+	tmpPath := tmpFile.Name()
+	defer os.Remove(tmpPath)
+
+	if _, err := tmpFile.Write(body); err != nil {
```
st3penta: `io.ReadAll`, `os.CreateTemp`, `tmpFile.Write`, and `json.Marshal(report)` errors are returned as HTTP error responses but the actual Go error value is discarded without logging. Other error paths in this same function do log (e.g. line 88, 131), so this looks like an oversight. Without the error in logs, operators can't diagnose root cause.

Something like `log.WithField("error", err).Error("...")` before each `writeError` call would be consistent with the rest of the handler.

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+	_ = json.NewEncoder(w).Encode(errorResponse{Error: message, Status: status})
+}
+
+func recoveryMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
+			if rec := recover(); rec != nil {
+				log.WithField("panic", rec).Error("Recovered from panic in handler")
```
st3penta: This logs the panic value but not the stack trace. Without `debug.Stack()`, operators can't tell where in the code the panic originated.

```suggestion
			log.WithFields(log.Fields{"panic": rec, "stack": string(debug.Stack())}).Error("Recovered from panic in handler")
```

(Needs `"runtime/debug"` import.)

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+		return ".yaml"
+	default:
+		if utils.IsJson(string(body)) {
+			return ".json"
+		}
+		return ".yaml"
+	}
+}
+
+func isValidInput(r *http.Request, body []byte) bool {
```
st3penta, line 193: `isValidInput` and `inputExtension` both use exact string matching on Content-Type. HTTP clients routinely send parameters like `application/json; charset=utf-8`, which fails the exact match and falls through to the `default` branch (more permissive auto-detection).

`mime.ParseMediaType` or splitting on `;` before matching would handle this correctly.

internal/server/server.go
```diff
@@ -0,0 +1,157 @@
...
+
+	select {
+	case err := <-errCh:
+		s.destroyEvaluators()
+		return err
+	case <-ctx.Done():
+	}
+
+	log.Info("Shutting down server...")
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
st3penta: The shutdown context gives 10s for in-flight requests to drain, but `evaluationTimeout` in handler.go is 90s. If a SIGTERM arrives during a long evaluation, the shutdown deadline fires first, then `destroyEvaluators()` runs while handler goroutines may still be mid-evaluation, racing against destroyed resources.

Either the shutdown timeout should be >= evaluationTimeout, or the handler should check for context cancellation from shutdown and bail early.

internal/server/server_test.go
```diff
@@ -0,0 +1,356 @@
...
+			_ = httpServer.Shutdown(shutdownCtx)
+		}()
+
+		errCh <- httpServer.ListenAndServe()
+	}()
+
+	time.Sleep(100 * time.Millisecond)
+
+	resp, err := http.Get(fmt.Sprintf("http://127.0.0.1:%d/live", port))
+	if err == nil {
```
st3penta: This `if err == nil` guard means that when the server fails to start (port conflict, scheduling delay, etc.), the test passes without asserting anything. Should be `require.NoError(t, err)` so a startup failure is caught instead of silently skipped.

internal/server/server_test.go
```diff
@@ -0,0 +1,356 @@
...
+	s.evaluators = []evaluator.Evaluator{&mockEvaluator{}}
+	s.ready.Store(true)
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	errCh := make(chan error, 1)
+	go func() {
+		mux := http.NewServeMux()
+		s.registerRoutes(mux)
+		httpServer := &http.Server{
```
st3penta: This test manually constructs an `http.Server` and calls `ListenAndServe` instead of calling `Server.Start()`. That means policy loading, graceful shutdown, evaluator cleanup, and log configuration from `Start()` are all untested. Worth testing through the public API so those code paths get coverage.

internal/server/server_test.go
```diff
@@ -0,0 +1,356 @@
...
+			<-ctx.Done()
+			shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 2*time.Second)
+			defer shutdownCancel()
+			_ = httpServer.Shutdown(shutdownCtx)
+		}()
+
+		errCh <- httpServer.ListenAndServe()
+	}()
+
+	time.Sleep(100 * time.Millisecond)
```
st3penta: Fixed 100ms sleep for server startup is timing-dependent and can flake under CI load. The acceptance tests already have a `waitForReady` helper with a poll loop. `require.Eventually` would also work here.

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	data, err := json.Marshal(report)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, "failed to marshal report")
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
```
st3penta: All completed evaluations return HTTP 200 regardless of the `success` value. While parsing the JSON is the canonical consumer path, a response header like `X-Conforma-Result: pass|fail` would let status-based consumers extract the signal without parsing the body.

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+const maxRequestBodySize = 10 << 20 // 10 MB
+
+var evaluationTimeout = 90 * time.Second
+
+type errorResponse struct {
+	Error  string `json:"error"`
+	Status int    `json:"status"`
+}
+
+func (s *Server) handleValidateInput(w http.ResponseWriter, r *http.Request) {
```
st3penta, line 49: Minor: the ~85 lines here interleave HTTP request parsing, temp file I/O, policy evaluation, and report assembly at different abstraction levels. Extracting the evaluate-and-build-report path into its own method would make this easier to read and test in isolation.

internal/server/middleware.go
```diff
@@ -0,0 +1,52 @@
...
+	sc.status = code
+	sc.ResponseWriter.WriteHeader(code)
+}
+
+func requestLoggingMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		sc := &statusCapture{ResponseWriter: w, status: http.StatusOK}
+
+		next.ServeHTTP(sc, r)
```
st3penta, line 63: `log.WithFields` with request context (method, path, remote addr) is called after `next.ServeHTTP` returns, so it only populates the final access log entry. Log calls inside the handler (e.g. "Evaluation failed") use the bare global logger without request fields, making it hard to correlate errors to specific requests.

Creating the request-scoped logger before calling `next` and passing it via `r.Context()` would let the handler include request context in its own log entries.

internal/server/server.go
```diff
@@ -0,0 +1,157 @@
...
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/conforma/cli/internal/evaluation_target/input"
+	"github.com/conforma/cli/internal/evaluator"
+	"github.com/conforma/cli/internal/policy"
+)
+
+type Config struct {
```
st3penta, line 35: EC-1885 names `--filter-type` as a flag that should apply as a server-wide default. The `Config` struct doesn't have a filter-type field and the handler doesn't process it. Probably intentional scoping for now, but worth tracking for a follow-up.

cmd/validate/input.go
```diff
@@ -138,6 +172,23 @@ func validateInputCmd(validate InputValidationFunc) *cobra.Command {
 			return
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if data.serverMode {
```
st3penta, line 179: When `--server` is set, flags like `--output`, `--strict`, `--workers`, `--filter-type`, `--no-color`, and `--color` are silently accepted but have no effect. Users won't realize the flags are being ignored. Worth either warning on explicitly-set incompatible flags or documenting which ones are no-ops in server mode.

cmd/validate/input.go
```diff
@@ -95,6 +112,19 @@ func validateInputCmd(validate InputValidationFunc) *cobra.Command {
...
+
+			  ec validate input --server --policy my-policy.yaml
+
+			Start the server on a custom port:
+
+			  ec validate input --server --server-port 9090 --policy my-policy.yaml
+
+			Send a request to the evaluation endpoint:
+
+			  curl -X POST -H 'Content-Type: application/json' --data-binary @input.json http://localhost:8080/v1/validate/input
```
st3penta: The example above sets `--server-port 9090`, but these curl examples hardcode port 8080. Sequential reading produces a broken workflow. Probably should use 9090 here or add a separate curl section under the default-port example.

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"github.com/conforma/cli/internal/evaluator"
+	"github.com/conforma/cli/internal/input"
+	"github.com/conforma/cli/internal/output"
+	"github.com/conforma/cli/internal/utils"
+)
+
+const maxRequestBodySize = 10 << 20 // 10 MB
+
+var evaluationTimeout = 90 * time.Second
```
st3penta: The 10MB body limit and 90s evaluation timeout are hardcoded but not mentioned anywhere in user-facing docs or CLI help text. Operators need these values to configure client timeouts correctly.

docs/modules/ROOT/pages/ec_validate_input.adoc
```diff
@@ -13,6 +13,17 @@ defined in the EnterpriseContractPolicy.
...
 If both are provided, the values are combined.
 
+When --server is provided, a persistent HTTP server is started instead of running a one-shot
+evaluation. Policies are loaded once at startup. The server exposes the following endpoints:
+
+  POST /v1/validate/input  - Evaluate input (JSON or YAML body)
+  GET  /live               - Liveness probe (always 200)
+  GET  /ready              - Readiness probe (200 when policies are loaded)
+
+The evaluation endpoint returns the same JSON structure as --output json. HTTP 200 is returned
```
st3penta, line 23: The docs describe the success path (HTTP 200 with JSON) but don't mention error responses. The server returns 400, 413, and 500 with `{"error":"...","status":N}` but consumers have no documented contract for those cases. Worth adding a short section covering the error response shape so integrators can handle failures.

internal/server/handler.go
```diff
@@ -0,0 +1,189 @@
...
+		writeError(w, http.StatusRequestEntityTooLarge, "request body exceeds 10MB limit")
+		return
+	}
+
+	if !isValidInput(r, body) {
+		writeError(w, http.StatusBadRequest, "request body is not valid JSON or YAML")
+		return
+	}
+
+	tmpFile, err := os.CreateTemp("", "ec-server-input-*"+inputExtension(r, body))
```
st3penta, line 73: Every test uses a single mock evaluator. The multi-evaluator aggregation path (the realistic production scenario with multiple policy sources) is never exercised.

internal/server/middleware.go
```diff
@@ -0,0 +1,52 @@
...
+package server
+
+import (
+	"net/http"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type statusCapture struct {
```
st3penta, line 39: The `statusCapture` wrapper's `WriteHeader` override and the implicit-200 path have zero test coverage. A double-write bug or broken delegation would silently cause every logged status to be wrong.

features/validate_input_server.feature
```diff
@@ -0,0 +1,67 @@
+@server
+Feature: validate input server mode
+  The ec command line should be able to run as a persistent HTTP server
+  for policy evaluation
+
+  Background:
+    Given stub git daemon running
+
+  Scenario: server mode evaluation with passing policy
```
st3penta, line 9: All four acceptance scenarios use JSON bodies. YAML input and `/ready` are tested at the unit level only, not end-to-end.

acceptance/cli/server.go
```diff
@@ -0,0 +1,354 @@
...
+		return ctx, fmt.Errorf("reading response: %w", err)
+	}
+
+	return context.WithValue(ctx, httpResponseKey, &httpResponse{
+		statusCode: resp.StatusCode,
+		body:       string(body),
+	}), nil
+}
+
+func aPOSTRequestIsSentToTheServerWithBody(ctx context.Context, urlPath string, content *godog.DocString) (context.Context, error) {
```
st3penta: These three HTTP helper functions (`aGETRequestIsSentToTheServer`, `aPOSTRequestIsSentToTheServerWithBody`, `aPOSTRequestIsSentToTheServerWithFile`) share roughly 80% identical structure. Could be consolidated into a common `doRequest` helper to reduce duplication.

internal/server/middleware.go
st3penta: Imho is also very useful to generate unique request UUIDs for each request, and add them as a field in the logs, so we can correlate log traces more easily for troubleshooting purposes

st3penta (commented):
Solid implementation overall, i'm satisfied with it.
I left a few improvement suggestions, and found some test coverage gaps.

internal/server/server.go
st3penta: @simonbaird i see you're following up about this, so i guess it can be resolved, up to you

st3penta (dismissed):

simonbaird:
I thought there were some more suggestions here, but actually they're just notes for about things we don't need right now. I think we're ready to merge. The failing conforma check is a known issue, so re-run should get this green.

internal/server/server.go
simonbaird: Yeah, I think I'll amend the description in EC-1885 and note that we currently do not pass that through to the evaluator for `validate input --server`.

simonbaird: Actually I decided to remove the flag.

simonbaird:
Added one more commit "Remove unused filter-type flag from validate input" and gave it one more rebase. I'm intending to merge if it goes green.

internal/server/handler.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/conforma/cli/internal/evaluator"
+	"github.com/conforma/cli/internal/input"
+	"github.com/conforma/cli/internal/output"
+	"github.com/conforma/cli/internal/utils"
+)
+
+const MaxRequestBodySize = 80 << 20 // 80 MB
+
+const EvaluationTimeout = 90 * time.Second
+
```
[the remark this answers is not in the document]
simonbaird: Let's leave it for a future PR.

internal/server/server.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (s *Server) registerRoutes(mux *http.ServeMux) {
+	mux.HandleFunc("GET /live", s.handleLive)
+	mux.HandleFunc("GET /ready", s.handleReady)
+	mux.HandleFunc("POST /v1/validate/input", s.handleValidateInput)
+}
+
+func (s *Server) handleLive(w http.ResponseWriter, _ *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
simonbaird: Response from my local agent:

> The channel is buffered (size 1), so the goroutine never leaks — it writes and exits. And line 108 already filters out ErrServerClosed, which is the only error Serve returns after Shutdown is called. So in the ctx.Done() path,
  nothing will be sent to errCh in practice. The goroutine finishes, the channel gets GC'd, done.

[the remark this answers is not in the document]
simonbaird: 🤷

internal/server/handler.go
```diff
@@ -0,0 +1,220 @@
...
+		logger.WithField("error", err).Error("Failed to marshal report")
+		return nil, false, fmt.Errorf("failed to marshal report")
+	}
+
+	return data, inp.Success, nil
+}
+
+func mediaType(r *http.Request) string {
+	ct := r.Header.Get("Content-Type")
+	if ct == "" {
```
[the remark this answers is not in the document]
simonbaird: I did note that, and it is a little inconsistent, but I'm going to leave it as-is for now.

internal/evaluator/conftest_evaluator.go
```diff
@@ -430,7 +430,10 @@ func (c conftestEvaluator) Evaluate(ctx context.Context, target EvaluationTarget
 	nonAnnotatedRules := nonAnnotatedRules{}
 	// Track data source directories for prepareDataDirs
 	dataSourceDirs := []string{}
-	// Download all sources
+	// Resolve all policy sources. Despite calling GetPolicy on every Evaluate,
+	// actual downloads are cached via sync.OnceValues in getPolicyThroughCache,
```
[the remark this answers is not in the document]
simonbaird: Fixed in upcoming revision.

internal/server/handler.go
```diff
@@ -0,0 +1,220 @@
...
+	if _, err := tmpFile.Write(body); err != nil {
+		tmpFile.Close()
+		logger.WithField("error", err).Error("Failed to write temp file")
+		writeError(w, http.StatusInternalServerError, "failed to write temp file")
+		return
+	}
+	tmpFile.Close()
+
+	data, success, err := s.evaluateAndBuildReport(ctx, tmpPath)
+	if err != nil {
```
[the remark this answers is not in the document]
simonbaird: Let's save this for a future PR.

simonbaird:
Okay, I found the fullsend comments than were more significant than nitpicks. I pushed on more race condition fix.

joejstuart (approved):

PR #3418, opened 2026-07-16
fix(EC-1912): validate in-toto statement type in bundle attestation path

BohdanMar:
`ProvenanceFromBundlePayload` accepted any `_type` value in in-toto statements, unlike the SLSA-specific parsers which reject unknown types. This adds the same validation to the bundle path — accept `v0.1` and `v1`, reject anything else.

Defense-in-depth only — the `attestation_type` policy rule already catches this at evaluation time.

Jira: [EC-1912](https://redhat.atlassian.net/browse/EC-1912) | Ref: [EC-1838](https://redhat.atlassian.net/browse/EC-1838)


st3penta (approved):
lgtm

PR #3423, opened 2026-07-18
chore(deps): Pin Go builder image by digest

### LINKED ISSUE #3358: Pin Go builder image in Dockerfile by digest
## Summary
Track a follow-up hardening change to pin the Go builder image in `Dockerfile` by digest instead of using only the mutable `docker.io/library/golang:1.26.3` tag.

## Rationale
The current build stage image reference is tag-only, which makes rebuilds less reproducible and weakens supply-chain hardening. The repository already uses digest-pinned base images in related container definitions (for example, `Dockerfile.dist`), so this would align the main `Dockerfile` with that pattern.

## Affected area
- `Dockerfile`

## Requested change
Update the build stage `FROM` instruction to pin the Go image by digest, e.g.:
- from: `docker.io/library/golang:1.26.3`
- to: `docker.io/library/golang:1.26.3@sha256:<resolved-digest>`

Keep the change localized to the builder `FROM` line unless additional supporting updates are required.

## Acceptance criteria
- The Go builder image in `Dockerfile` is pinned by digest.
- The referenced digest matches the intended `golang:1.26.3` image variant used by this build.
- The change is made in a follow-up commit or PR, not as part of #3357.

## Backlinks
- Follow-up requested by @simonbaird
- PR: https://github.com/conforma/cli/pull/3357
- Review comment: https://github.com/conforma/cli/pull/3357#discussion_r3477893391

iRayans:
Pins the Go builder image by digest, matching Dockerfile.dist and the final image in this file.
Digest taken from the manifest list, so all platforms in `BUILD_LIST` still resolve. Other FROM in this file is already pinned.
Fixes #3358

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@
 
 ## Build
 
-FROM docker.io/library/golang:1.26.3 AS build
+FROM docker.io/library/golang:1.26.3@sha256:2d6c80227255c3112a4d08e67ba98e58efd3846daf15d9d7d4c389565d881b1a AS build
 
 ARG TARGETOS
 ARG TARGETARCH


st3penta:
Checked with:
```
$ skopeo inspect docker://docker.io/library/golang:1.26.3 | jq -r .Digest
sha256:2d6c80227255c3112a4d08e67ba98e58efd3846daf15d9d7d4c389565d881b1a

$ make build-image
...

$ make verify-image
...
```

Confirmed working. Thanks @iRayans!

st3penta (approved):

PR #3425, opened 2026-07-20
fix(EC-1994): correct off-by-one in validate image worker loop

cuipinghuo:
## Summary
- Fix off-by-one in worker loop at `cmd/validate/image.go:404`: `<=` changed to `<` so the loop spawns exactly `numWorkers` goroutines instead of `numWorkers+1`
- Bump task worker defaults by 1 to preserve the effective concurrency users had before the fix (since the old `<=` loop spawned N+1 workers for a configured value of N):
  - `verify-conforma-konflux-ta`: 4 → 5
  - `verify-enterprise-contract`: 1 → 2
- Update READMEs and generated docs to match new defaults; fix stale documentation that incorrectly claimed WORKERS was unused and defaulted to 35

## Test plan
- [x] `go build ./cmd/validate/` passes
- [x] `go test -tags=unit ./cmd/validate/ -run TestValidateImage` passes
- [x] `make generate` produces no additional diff

Resolves: https://redhat.atlassian.net/browse/EC-1994

### DIFF
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -400,7 +400,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			results := make(chan validate_utils.Result, numComponents)
 			// Initialize each worker. They will wait patiently until a job is sent to the jobs
 			// channel, or the jobs channel is closed.
-			for i := 0; i <= numWorkers; i++ {
+			for i := 0; i < numWorkers; i++ {
 				go worker(i, jobs, results)
 			}
 			// Initialize all the jobs. Each worker will pick a job from the channel when the worker
--- a/cmd/validate/image.go
+++ b/cmd/validate/image.go
@@ -401,7 +401,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			results := make(chan validate_utils.Result, numComponents)
 			// Initialize each worker. They will wait patiently until a job is sent to the jobs
 			// channel, or the jobs channel is closed.
-			for i := 0; i <= numWorkers; i++ {
+			for i := 0; i < numWorkers; i++ {
 				go worker(i, jobs, results)
 			}
 			// Initialize all the jobs. Each worker will pick a job from the channel when the worker
--- a/tasks/verify-conforma-konflux-ta/0.1/README.md
+++ b/tasks/verify-conforma-konflux-ta/0.1/README.md
@@ -31,7 +31,7 @@ kubectl apply -f https://raw.githubusercontent.com/conforma/cli/main/tasks/verif
 * **HOMEDIR**: Value for the HOME environment variable. (default: "/tekton/home")
 * **EFFECTIVE_TIME**: Run policy checks with the provided time. (default: "now")
 * **EXTRA_RULE_DATA**: Merge additional Rego variables into the policy data. Use syntax "key=value,key2=value2..." (default: "")
-* **WORKERS**: Number of parallel workers to use for policy evaluation. This parameter is currently not used. All policy evaluations are run with 35 workers. (default: "35")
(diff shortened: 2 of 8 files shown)


jsmid1 (dismissed):

joejstuart:
The change looks good, I'm just unsure of the impact to the users that modify the `workers` outside of the task (ie: in konflux-release-data). All configs that modify the count are set to `35`, which means we're only taking away `1` worker, so it should be negligible. They could see a little slower runtimes though. In theory it should slow a 10 minute Conforma run by 17 seconds.
The alternative is we live with this bug.

tasks/verify-conforma-konflux-ta/0.1/README.md
```diff
@@ -31,7 +31,7 @@ kubectl apply -f https://raw.githubusercontent.com/conforma/cli/main/tasks/verif
 * **HOMEDIR**: Value for the HOME environment variable. (default: "/tekton/home")
 * **EFFECTIVE_TIME**: Run policy checks with the provided time. (default: "now")
 * **EXTRA_RULE_DATA**: Merge additional Rego variables into the policy data. Use syntax "key=value,key2=value2..." (default: "")
-* **WORKERS**: Number of parallel workers to use for policy evaluation. This parameter is currently not used. All policy evaluations are run with 35 workers. (default: "35")
```
simonbaird, line 34: This was some left-over incorrect documentation, right??

cuipinghuo: Yes, it was stale/incorrect documentation. It was introduced in 5302c521 (`fix(ec-1404)`) — that commit rewrote this README line to say "This parameter is currently not used. All policy evaluations are run with 35 workers." The "35" likely came from `conforma.workerCount: 35` which is set in ~107 `ReleasePlanAdmission` CRDs in `konflux-release-data`. However, that `workerCount` field is never wired to the task's `WORKERS` param — the release pipelines in `release-service-catalog` don't pass a `WORKERS` param at all, so the task YAML default always controlled `--workers`. The README was wrong from the moment it was written.

cuipinghuo: Correction to my previous comment: my local clone of `release-service-catalog` was stale. The upstream `konflux-ci/release-service-catalog` **does** pass `WORKERS` in the managed pipelines (under `pipelines/managed/`), e.g. https://github.com/konflux-ci/release-service-catalog/blob/main/pipelines/managed/rh-advisories/rh-advisories.yaml — the value comes from `$(tasks.collect-task-params.results.extractedValues[0])`, which extracts `conforma.workerCount: 35` from `konflux-release-data`.

So for release pipelines, the task default doesn't matter — it gets overridden to 35 (which with the old `<=` bug meant 36 actual goroutines). The task default only applies when the task is invoked without an explicit `WORKERS` param (e.g. integration pipelines via `verify-conforma-konflux-ta`).

The README text "This parameter is currently not used" was still misleading though — the parameter **is** used, it's just that the release pipelines override it.

jsmid1 (dismissed):

BohdanMar (dismissed):

cmd/validate/image.go
```diff
@@ -401,7 +401,7 @@ func validateImageCmd(validate imageValidationFunc) *cobra.Command {
 			results := make(chan validate_utils.Result, numComponents)
 			// Initialize each worker. They will wait patiently until a job is sent to the jobs
 			// channel, or the jobs channel is closed.
-			for i := 0; i <= numWorkers; i++ {
+			for i := 0; i < numWorkers; i++ {
```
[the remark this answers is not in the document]
simonbaird: I'm not sure that's worth it.

simonbaird (approved):

simonbaird:
Rebase should fix the EC violation now that #3496 is merged.

cmd/validate/image.go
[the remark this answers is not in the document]
cuipinghuo: won't fix.

robnester-rh (approved):
LGTM

PR #3431, opened 2026-07-22
Fix pinned ref cache miss to avoid policy refetch

simonbaird:
URL pinning mutates p.Url after the first GetPolicy call (e.g. github.com/org/repo -> git::github.com/org/repo?ref=abc123), causing subsequent cache lookups to miss and re-download into a second directory. OPA then sees duplicate packages and errors out. Fix by also storing the cached download function under the pinned URL.

I noticed this problem when testing the new http service with a git url in the policy.yaml.

Ref: https://redhat.atlassian.net/browse/EC-2017

### DIFF
--- a/internal/policy/source/source.go
+++ b/internal/policy/source/source.go
@@ -188,9 +188,17 @@ func (p *PolicyUrl) GetPolicy(ctx context.Context, workDir string, showMsg bool)
 	}
 
 	var pinErr error
+	originalUrl := p.Url
 	p.pinOnce.Do(func() {
 		p.Url, pinErr = metadata.GetPinnedURL(p.Url)
 		log.Debug("Pinned URL: ", p.Url)
+		// Register the cached download under the pinned URL too, so
+		// subsequent lookups (which use the now-mutated p.Url) still hit.
+		if pinErr == nil && p.Url != originalUrl {
+			if cached, ok := downloadCache.Load(originalUrl); ok {
+				downloadCache.LoadOrStore(p.Url, cached)
+			}
+		}
 	})
 	if pinErr != nil {
 		return "", pinErr
--- a/internal/policy/source/source.go
+++ b/internal/policy/source/source.go
@@ -74,6 +74,7 @@ type PolicyUrl struct {
 	Url     string
 	Kind    PolicyType
 	pinOnce sync.Once
+	urlMu   sync.RWMutex
 }
 
 // downloadCache is a concurrent map used to cache downloaded files.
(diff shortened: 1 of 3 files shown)


internal/policy/source/source.go
```diff
@@ -188,9 +188,17 @@ func (p *PolicyUrl) GetPolicy(ctx context.Context, workDir string, showMsg bool)
...
 	p.pinOnce.Do(func() {
 		p.Url, pinErr = metadata.GetPinnedURL(p.Url)
 		log.Debug("Pinned URL: ", p.Url)
+		// Register the cached download under the pinned URL too, so
+		// subsequent lookups (which use the now-mutated p.Url) still hit.
+		if pinErr == nil && p.Url != originalUrl {
+			if cached, ok := downloadCache.Load(originalUrl); ok {
+				downloadCache.LoadOrStore(p.Url, cached)
+			}
+		}
```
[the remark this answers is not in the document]
simonbaird: Should be fixed now.

internal/policy/source/source_test.go
```diff
@@ -634,6 +634,49 @@ func TestDownloadCacheWorkdirMismatch(t *testing.T) {
 	assert.Equal(t, destination1, destination2)
 }
 
+// TestGetPolicyPinnedURLCacheConsistency verifies that URL pinning doesn't
+// cause duplicate policy directories. After the first GetPolicy call, the URL
+// is pinned (e.g. github.com/org/repo -> git::github.com/org/repo?ref=abc123),
+// which changes the cache key. Without the fix, the second call would miss the
+// cache and re-download into a new directory, causing OPA duplicate package errors.
```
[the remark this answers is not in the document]
simonbaird: Should be fixed now.

internal/policy/source/source.go
```diff
@@ -74,6 +74,7 @@ type PolicyUrl struct {
 	Url     string
 	Kind    PolicyType
 	pinOnce sync.Once
+	urlMu   sync.RWMutex
```
st3penta, line 77: Not for this PR, but `PolicyUrl` now has three sync primitives: `pinOnce`, `urlMu`, `downloadCache`. Maybe we should consider a consolidation of the sync surface?

st3penta (approved):

internal/policy/source/source.go
simonbaird: Yeah, sounds like a good idea.

simonbaird: Created https://redhat.atlassian.net/browse/EC-2028 .

PR #3433, opened 2026-07-23
[EC-1776] docs: Document Cicada content stream update in release process

jsmid1:
Add a section to the release process docs explaining how to update Pyxis content_stream_tags via Cicada when creating or retiring a release branch.

Ref: https://issues.redhat.com/browse/EC-1776


simonbaird (approved):
Nice!

PR #3476, opened 2026-08-07
[DNM] Add dummy task, pipeline, and ITS for EC-2011 POC

simonbaird:
## Summary
- Adds a dummy-check Tekton task that produces configurable pass/fail/warn results with a test-result attestation step (via the `attest-test-result` step action)
- Adds a `reqd-task-poc-ec2011` pipeline that parses a Snapshot and runs the dummy-check task
- Adds `hack/create-dummy-its.sh` to create the IntegrationTestScenario in the cluster

## Context
POC for [EC-2011](https://redhat.atlassian.net/browse/EC-2011) — dog-fooding a required task in an ITS instead of the build pipeline.

## Test plan
- [ ] ITS created in `rhtap-contract-tenant` namespace
- [ ] Pipeline triggers on PR and runs successfully
- [ ] Test-result attestation is created and attached to the image
- [ ] Conforma can discover and evaluate the attestation


st3penta (commented):
nice!
i won't approve since it's DNM, just to be safe, but looks good as a POC

PR #3491, opened 2026-08-11
Pin conforma tekton task to digest

joejstuart:
Pin this to a digest so we can control when changes are released. This is to sync the changes with the release-service.


robnester-rh (approved):
LGTM

simonbaird (changes requested):
I think that's the bad sha.

PR #3493, opened 2026-08-13
chore(deps): Update ubi-minimal base image

st3penta:


simonbaird (approved):
FYI I have a nicer automation wrapper for this in https://github.com/conforma/cli/pull/3430 which I should have suggested that you try.

PR #2176, opened 2024-11-19
Revert "Merge pull request #2150 from zregvart/issue/EC-963"

robnester-rh:
Revert "Use given source URLs as keys in download cache"

This reverts commit 087f4659b10ada36b092d3f599e024d5c79ac955.

Ref: EC-1023
Signed-off-by: robnester-rh <rnester@redhat.com>

PR #3505, opened 2026-08-18
Restore commented out acceptance test scenario

simonbaird:
I believe it was failing at some point, so it was commented out. https://redhat.atlassian.net/browse/EC-1023 is meant to describe the reason for the failure.

It's not clear to me what the detailed story is, but this comment and the revert PR it is part of are definitely clues: https://github.com/conforma/cli/pull/2176#issuecomment-2486792741

Seems to work fine now, so let's bring it back. See also https://github.com/conforma/cli/commit/44618120b295398353f99b565c8c2ef24a9231eb to help understand the purpose of the test.

### DIFF
--- a/features/validate_image.feature
+++ b/features/validate_image.feature
@@ -1461,26 +1461,25 @@ Feature: evaluate enterprise contract
     Then the exit status should be 0
     Then the output should match the snapshot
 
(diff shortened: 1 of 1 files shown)


st3penta (approved):
LGTM. Thanks, i'll revert the snapshot cleanup in https://github.com/conforma/cli/pull/3489

PR #3530, opened 2026-09-02
Find roxctl reports also using benchmark offliner

simonbaird:
We're moving from clair-scan to roxctl-scan. This change should ensure that the benchmark offliner utility finds the roxctl-scan reports the same way it finds the clair-scan reports.


jsmid1 (approved):
lgtm

cuipinghuo (approved):

PR #3537, opened 2026-09-03
fix: Remove reference to deleted quick-build-args.conf

dheerajodha:
#### What:

Since PR #2478 removed the build args files, the release branch script should set both build args to empty string instead of referencing the no longer existing quick-build-args.conf.

#### Why:

So, we won't have to manually remove it anymore.

#### Tickets:

Ref: https://redhat.atlassian.net/browse/EC-2154


BohdanMar (approved):
Lgtm

PR #3541, opened 2026-09-03
upgrade go to version 1.26.7

Acepresso:
Also updated 3 packages in package-lock.json following npm complain.


simonbaird (changes requested):
Should be abandoned, do you agree?

PR #3557, opened 2026-09-10
EC-2028: consolidate sync primitives in PolicyUrl (source.go)

jsmid1:
Replace the four separate sync primitives that had accumulated in internal/policy/source/source.go with two small, purpose-built types:

- urlPin bundles PolicyUrl's pinOnce (sync.Once) + urlMu (sync.RWMutex) into one type with read/write/doOnce methods, since both fields existed only to coordinate lazy, one-time URL pinning.

- policyDownloadCache bundles the package-level downloadCache and symlinkMutexes sync.Map variables into a single struct that owns both the download memoization and the per-destination symlink coordination for that same caching feature.

This is a pure refactor: locking granularity and semantics are unchanged (in particular, the existing pin-once-and-don't-retry behavior on pinning errors is preserved), only the surface is consolidated and given clearer names/docs. Two test call sites that poked the internal sync.Map directly were updated to go through the new nested field.

No public API changes: PolicyUrl{Url, Kind} construction, Get/PolicyUrl(), Subdir(), Type(), and ClearDownloadCache() are unaffected.

Resolves: [EC-2028](https://redhat.atlassian.net/browse/EC-2028)

### DIFF
--- a/internal/policy/source/source.go
+++ b/internal/policy/source/source.go
@@ -71,22 +71,81 @@ type PolicySource interface {
 
 type PolicyUrl struct {
 	// A string containing a go-getter style source url compatible with conftest pull
-	Url     string
-	Kind    PolicyType
-	pinOnce sync.Once
-	urlMu   sync.RWMutex
+	Url  string
+	Kind PolicyType
+	pin  urlPin
 }
 
-// downloadCache is a concurrent map used to cache downloaded files.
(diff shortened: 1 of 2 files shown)


simonbaird (approved):
Lgtm.

BohdanMar (approved):
Lgtm.

PR #3558, opened 2026-09-10
chore(deps): Update testcontainers-go and moby/go-archive (main)

simonbaird:
## What

Remediates **CVE-2026-17106** (moby/go-archive "CopyEscape" — arbitrary file write via link following in tar extraction), fixed in `github.com/moby/go-archive` **v0.3.0+**.

## Why the extra work

go-archive was pinned to **v0.1.0** via a `replace` directive because `docker/docker` v28.5.2's `pkg/archive` uses the deprecated `archive.Compression` root alias that go-archive removed in v0.2.0. The CVE fix (v0.3.0+) also lacks that alias, so it can't compile against docker v28.5.2.

Bumping docker isn't an option: there is **no v29 Go module** for `github.com/docker/docker` (it tops out at v28.5.2; Docker v29 moved to `github.com/moby/moby/v2`, which currently only has betas).

The real lever is **testcontainers-go**, the only thing that pulls in go-archive (via benchmark + acceptance tooling — go-archive is **not** in the shipped `ec` binary). **testcontainers-go v0.44.0** dropped `docker/docker/pkg/archive` in favour of `moby/moby/{api,client}` and depends on go-archive directly, so the `replace` can be removed and go-archive bumped to v0.3.1.

## Changes

- **root + `acceptance/` go.mod**: drop the go-archive `replace`, bump `testcontainers-go` v0.34 → **v0.44.0** (+ `modules/registry`), `go-archive` → **v0.3.1**
- **benchmark**: import `moby/moby/api/types/container` (config-modifier signatures moved there in testcontainers v0.44)
- **acceptance**: `MappedPort` now returns `moby/.../network.Port`; use `Num()` instead of the removed `Int()`
- **acceptance**: testcontainers v0.44 no longer accepts a host-binding IP in `ExposedPorts`, so the IPv4-only port binding (HACBS-1354) moved into a `HostConfigModifier` that sets `PortBindings`

## Testing

- `go build ./...` + `go vet ./...` clean in both the root and `acceptance/` modules
- Full acceptance suite passes locally

> Note: running the acceptance suite locally needs a reasonable `fs.inotify.max_user_instances` (the KIND-based scenarios plus the zot registry's fsnotify hot-reloader exhaust the default 128 on some setups, causing the registry container to exit 1 on startup). Raising it to 1024 resolves it.


robnester-rh (approved):
LGTM

joejstuart (approved):

